PR #6, opened 2016-08-16
Requires illuminate/routing, extends Illuminate\Routing\Controller on

### LINKED ISSUE #5: Call to undefined method Laravel\Passport\Http\Controllers\AuthorizationController
Following through the initial setup on a clean 5.3 install and getting this error when using `php artisian route:list` after adding in Passport::routes() to the AuthServiceProvider.

```
[Symfony\Component\Debug\Exception\FatalThrowableError]
  Call to undefined method Laravel\Passport\Http\Controllers\AuthorizationController::getMiddleware()
```

underparnv:
each controller in the package. Fixes #5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "illuminate/encryption": "~5.3",
         "illuminate/http": "~5.3",
         "illuminate/support": "~5.3",
+        "illuminate/routing": "~5.3",
         "league/oauth2-server": "~5.0",
         "symfony/process": "~3.1",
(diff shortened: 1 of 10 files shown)


jbrooksuk:
Laravel 5.3 controllers don't need to extend the `Controller` class.

taylorotwell:
Not necessary.

taylorotwell:
The other bug you mentioned has already been fixed.

PR #15, opened 2016-08-18
Also tests for PHP 5.6.4

lucasmichot:
Also tests for PHP 5.6.4
Because 5.6 is for the latest 5.6.X version


GrahamCampbell:
This won't work.

lucasmichot:
Seems Travis does not have 5.6.4...

GrahamCampbell:
Yeh, 5.5.9 and 5.3.3 are the only special versions it has.

PR #22, opened 2016-08-19
Fetch token cookie name from config

rigor789:
Fetch the token cookie name from the config instead of hardcoding it.

Updated tests since the `TokenGuard` constructor has been changed.

Added test to confirm the cookie name can be changed.


taylorotwell:
There are better ways to do this which I can explore.

rigor789:
Alright, thanks for considering the proposed change! :)

PR #30, opened 2016-08-20
Fix coding style

nikitakiselev:


taylorotwell:
Was fine as-is I think.

PR #31, opened 2016-08-20
Add keys to .gitignore

mdavis1982:
This PR adds the oauth keys to the `.gitignore` file if they don't already exist in it. This helps to prevent accidentally committing the keys to VCS.


morloderex:
@mdavis1982 The idea here is fine but the implimentation doesn't scale. 
Forexample this prevents the package from beeing used by people without the grep command avaliable to them.

mdavis1982:
@morloderex I'm not sure there's a supported system that doesn't have `grep` installed by default. It's certainly available on Linux and OS X, and the command will fail on Windows anyway because `OpenSSL` won't be installed by default.

maximebeaudoin:
https://github.com/laravel/passport/pull/33 sounds a better solution IMO.

taylorotwell:
Closing because of issues raised in thread.

PR #33, opened 2016-08-20
Store keys in storage/app/keys

lucasmichot:
Store keys in `storage/app/keys`
This way, they are by default ignored by Git


tomschlick:
Whats the best way to sync keys across multiple app servers when/if this happens?

taylorotwell:
I have it on my to-do list to just make this configurable.

PR #34, opened 2016-08-21
Enable client credentials grant type

themsaid:
This grant type is useful to allow clients to access data that's not owned by a certain user, for example: searching tweets, listing supported countries, etc...

The PR implements a middleware that can be used to check for client credentials.

### DIFF
--- a/src/Http/Middleware/CheckClientCredentials.php
+++ b/src/Http/Middleware/CheckClientCredentials.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Laravel\Passport\Http\Middleware;
+
+use Closure;
+use League\OAuth2\Server\ResourceServer;
+use Illuminate\Auth\AuthenticationException;
+use League\OAuth2\Server\Exception\OAuthServerException;
+use Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory;
+
+class CheckClientCredentials
+{
+    /**
+     * The Resource Server instance.
+     *
+     * @var ResourceServer
+     */
+    private $server;
+
+    /**
+     * Create a new middleware instance.
+     *
+     * @param  ResourceServer  $server
+     * @return void
+     */
+    public function __construct(ResourceServer $server)
+    {
+        $this->server = $server;
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  \Closure $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $psr = (new DiactorosFactory)->createRequest($request);
+
+        try{
+            $this->server->validateAuthenticatedRequest($psr);
+        } catch (OAuthServerException $e) {
+            throw new AuthenticationException;
+        }
+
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CheckClientCredentials.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this->server = $server;
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  \Closure $next
+     * @return mixed
```
lucasmichot: Missing `@throws \Illuminate\Auth\AuthenticationException`

themsaid: Updated, thanks!

taylorotwell:
Why define a whole new middleware? The TokenGuard will already call the validateAuthenticatedRequest method on the resource server.

themsaid:
For this type of grants there's no User, the TokenGuard expects to be able to successfully retrieve a user.

ganey:
I'm also interested in this, I was using lucadegasperi/oauth2-server-laravel with client_credential tokens before upgrading to 5.3

src/Http/Middleware/CheckClientCredentials.php
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @throws \Illuminate\Auth\AuthenticationException
+     */
+    public function handle($request, Closure $next)
```
ganey: If you wanted to check scopes too, something along the following lines works:
This then works in the same way as the CheckForAnyScope middleware.

```
public function handle($request, Closure $next, ...$scopes)
  {
    $psr = (new DiactorosFactory)->createRequest($request);

    try {
      /**
       * @var $serverRequest ServerRequest
       */
      $serverRequest = $this->server->validateAuthenticatedRequest($psr);
    } catch (OAuthServerException $e) {
      throw new AuthenticationException;
    }

    foreach ($scopes as $scope) {
      if (!in_array($scope,$serverRequest->getAttribute('oauth_scopes'))) {
        throw new AuthenticationException;
      }
    }

    return $next($request);
  }
```

enx1998:
Hi everyone, 
I don't know if this is the right place di place this question, but the argument is the client_credentials grant type.
If I understand well, the client_credential grant was integrated in the code, but I don't know how can I use it to protect my routes.
All I want to do is to make a route available to client that have client_credential token.

Thanks ENx

src/Http/Middleware/CheckClientCredentials.php
endyjasmi: This works except when you try to request access token with '*' as scope. Seems like the logic for handling all scope in client credential grant is left out.

ankurk91:
An access_token that was generated by `password` grant can used by bypass this middleware.
Does specs allow this?

driesvints:
@ankurk91 yeah, you can use the same client for multiple grants but it's not recommended. It's best to only use one grant per client.

ankurk91:
Thanks for the answer.

PR #39, opened 2016-08-22
Fix the docblock

jbrooksuk:


src/Guards/TokenGuard.php
```diff
@@ -194,7 +194,7 @@ protected function decodeJwtTokenCookie($request)
      * Determine if the CSRF / header are valid and match.
      *
      * @param  array  $token
-     * @param  string  $header
+     * @param  Request  $request
```
GrahamCampbell: fqcn plz :trollface:

jbrooksuk: I did it this way as the file doesn't use FQCN's yet. I do have a branch where I'm going through each docblock and importing it fully. For now, it follows the rest of them 👍

PR #49, opened 2016-08-23
OAuth Client ID with string

### LINKED ISSUE #14: Client_ID
It feels a bit strange to get a client ID as the primary key of oauth_clients @taylorotwell 
Would it not make sense to provide a uuid field for that ?

nueko:
Fix #14


GrahamCampbell:
It's too late to change this now that 1.0.0 is out.

nueko:
although i don't understand the 'too late' words mean. That's okay, I made it for my sake :smile:

GrahamCampbell:
Well, it could be considered for 2.x I guess.

lucasmichot:
Hey @nueko , thanks for bootstrapping this !
For the client ID, I think I would stick to a random string I guess, unless @taylorotwell got another approach

As long as it's unique AND indexed

nueko:
You are welcome.
i think, if random string will never generate duplicate, it will be ok.

PR #57, opened 2016-08-27
Update comment to include $response

jasrys:
First ever pull request. Hope I'm doing this right. :)


morloderex:
Could you typehind what class it is.

Issue #52, opened 2016-08-25
Ajax authentication issue


sagarrabadiya:
CreateFreshApiToken middleware should issue `laravel_token` cookie if post request is coming through ajax, so that one can allow user to login with ajax for ex. vue-resource.

for now because of this i can send authentication data using vue resource to server, but because of issuing token only for get request it's not returning laravel_token cookie so user can not access api correctly if login is done using ajax.

for example
`return $request->isMethod('GET') && $request->user();`

can be
`return ($request->isMethod('GET') || $request->ajax() ) && $request->user();`

Townfolio:
I don't believe your problem lies in how you're posting your data to the server to login. I think it might lay in what you're doing after that login screen. In order for a response to receive a fresh token, it has to meet a few criteria, the first one being what you've shown above.

```
/**
 * Determine if the request should receive a fresh token.
 *
 * @param  \Illuminate\Http\Request  $request
 * @return bool
 */
protected function requestShouldReceiveFreshToken($request)
{
    return $request->isMethod('GET') && $request->user();
}
```

The second one (and I believe this might be tripping you up possibly) is this.

```
/**
 * Determine if the response should receive a fresh token.
 *
 * @param  \Illuminate\Http\Response  $request
 * @return bool
 */
protected function responseShouldReceiveFreshToken($response)
{
    return $response instanceof Response &&
                ! $this->alreadyContainsToken($response);
}
```

That `instanceof Response` means that the response has to be `\Illuminate\Http\Response` and the only way it wouldn't (in my mind) in your app is if you're doing a SPA-like front end where after you login you're doing some front end routing instead of redirecting to a new page from the backend. If that is the case, then you should consider making the login screen separate from the front end routing. 

If that isn't the case, please provide more context so we can figure out a solution for you.

sagarrabadiya:
yeah, i am redirecting user after login using vue route but why is it necessary to keep login age seperate from my spa? its not that good experience as it do full age load

Townfolio:
It's not necessary, but in order to make your app work with Passport and the `CreateFreshApiToken` middleware, you would need to. Passport is expecting the GET response to be an instance of `\Illuminate\Http\Response` in order to attach the cookie, but since you're not logged in at the time of it issuing the only `\Illuminate\Http\Response`, no cookie can ever be issued.

[...]

PR #59, opened 2016-08-29
fix for #52

sagarrabadiya:
If the login to the application is performed with ajax and single page application, then it should generate api token cookie, so single page application can consider a successful login.

based on discussion at #52


taylorotwell:
Eh, this isn't really needed. As long as your page loaded with a GET request you already have a token. If you want to create a new token just make a GET request to any route. The cookie will automatically be sent on all other requests.

sagarrabadiya:
do you mean after post ajax request, i need to call any one get request with ajax to set the token as cookie? if so, I already tried it and it was not working.

sagarrabadiya:
without full page load cookie isn't stored correctly, after ajax post login

PR #72, opened 2016-09-07
Use phpseclib for generating keys

samgreenwood:
Sometimes the 'openssl' cli utility may not me available, especially on windows this may require some ceremony.

Updated this command to leverage phpseclib instead.

### DIFF
--- a/src/Console/KeysCommand.php
+++ b/src/Console/KeysCommand.php
@@ -2,8 +2,8 @@
 
 namespace Laravel\Passport\Console;
 
+use phpseclib\Crypt\RSA;
 use Illuminate\Console\Command;
-use Symfony\Component\Process\Process;
 
 class KeysCommand extends Command
 {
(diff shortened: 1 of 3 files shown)


tests/KeysCommandTest.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        Mockery::close();
+
+        @unlink(storage_path('oauth-private.key'));
+        @unlink(storage_path('oauth-public.key'));
+    }
+
+    public function testPrivateAndPublicKeysAreGenerated()
+    {
+        $command = Mockery::mock(\Laravel\Passport\Console\KeysCommand::class)
```
GrahamCampbell: No leading slash needed please.

tests/KeysCommandTest.php
```diff
@@ -0,0 +1,33 @@
...
+            ->makePartial()
+            ->shouldReceive('info')
+            ->with('Encryption keys generated successfully.')
+            ->getMock();
+
+        $rsa = new \phpseclib\Crypt\RSA();
```
GrahamCampbell: same

src/Console/KeysCommand.php
```diff
@@ -24,16 +24,15 @@ class KeysCommand extends Command
     /**
      * Execute the console command.
      *
+     * @param RSA $rsa
```
GrahamCampbell: Please format this correctly.

GrahamCampbell:
Have you tested this output exactly matches the other command?

samgreenwood:
CS fixes done.

From what I can tell the keys generated are the same.

nueko:
I think it should be on require-dev.

taylorotwell:
It actually can't be require-dev because require-dev is only installed for the root package (laravel/framework)

PR #79, opened 2016-09-12
Add XRSF token validation as alternative

RSchwan:


taylorotwell:
Just sticking with what we have.

RSchwan:
Laravel generates a [`XSRF-TOKEN token` cookie](https://laravel.com/docs/5.3/csrf#csrf-x-xsrf-token). Why not give the user the opportunity to use it?

RSchwan:
@taylorotwell I'm trying to integrate passport into spark and this small addition would make it way easier. Without the XSRF token validation it's going to be hard to implement a working session expired modal.
And also as mentioned in the docs, for some frameworks like angular the XSRF token is handled automatically. The XSRF token is basically nothing else then an other representation of the CSRF token so I think this would be a nice addition / alternative.

williammck:
@taylorotwell Is there a specific reason this was closed other than just "keeping it the way it is"?

For single page applications this would be a godsend, and while extending the TokenGuard to add this functionality *works*, wouldn't this be beneficial for others who aren't using Spark but are building a single page app? As @RSchwan mentioned above, the functionality for handling XSRF tokens is already built into Laravel, seems odd that it's not here.

PR #80, opened 2016-09-12
Fixed #53

### LINKED ISSUE #53: Weird UI Issue along with Console Warnings.
Hello, 

I tried installing passport and following the tutorial and the first time it all worked fine but later on, I broke something and decided to start fresh. I deleted the database and got a fresh copy of Laravel. 

Now when I tried to follow the exact same tutorial on the Docs and the video Taylor made i get some issues. I don't know what's causing them and I don't know how to fix them.

Here are the issues.

screen shot 2016-08-26 at 00 01 58

When i click the delete buttons i get this message

screen shot 2016-08-26 at 00 16 07

I have used a new database and i tried in incognito and cleared all my browsers cookies and cache. 

I have asked on IRC but no one can help me.

fwartner:
bildschirmfoto 2016-09-12 um 17 10 00

### DIFF
--- a/resources/assets/js/components/AuthorizedClients.vue
+++ b/resources/assets/js/components/AuthorizedClients.vue
@@ -76,8 +76,8 @@
              */
             getTokens() {
                 this.$http.get('/oauth/tokens')
-                        .then(response => {
-                            this.tokens = response.data;
+                        .then((response) => {
+                            this.tokens = response.json();
                         });
             },
 
@@ -86,7 +86,7 @@
              */
             revoke(token) {
                 this.$http.delete('/oauth/tokens/' + token.id)
-                        .then(response => {
+                        .then((response) => {
                             this.getTokens();
                         });
             }
--- a/resources/assets/js/components/Clients.vue
+++ b/resources/assets/js/components/Clients.vue
@@ -256,8 +256,8 @@
              */
             getClients() {
                 this.$http.get('/oauth/clients')
-                        .then(response => {
-                            this.clients = response.data;
+                        .then((response) => {
+                            this.clients = response.json();
                         });
             },
 
@@ -306,7 +306,7 @@
                 form.errors = [];
 
                 this.$http[method](uri, form)
-                    .then(response => {
+                    .then((response) => {
                         this.getClients();
 
                         form.name = '';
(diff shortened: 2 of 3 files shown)


resources/assets/js/components/AuthorizedClients.vue
```diff
@@ -76,8 +76,8 @@
              */
             getTokens() {
                 this.$http.get('/oauth/tokens')
-                        .then(response => {
-                            this.tokens = response.data;
+                        .then((response) => {
+                            this.tokens = response.json();
```
jbrooksuk: You don't need to do this if there is only one parameter.

fwartner: I don´t need to do what exactly?

taylorotwell:
@themsaid isn't using `.data` still fine? @fwartner can you describe the actual issue?

fwartner:
The issue (#53) was that when passport runs behind https, the ui gave us some weird stuff. Also the console.

I had the problem once in 5.3 with the ```response.data```.. After i changed it to ```response.json()```it worked.

fwartner:
I think it has something to do with the vue-resource package..

taylorotwell:
Yeah it's a vue-resource bug that they need to fix on their end. As far as I know there is a PR open on their side.

resources/assets/js/components/AuthorizedClients.vue
jbrooksuk: To wrap the callback in brackets.

fwartner:
@taylorotwell can´t find the pr you mentioned. Would you like to reference it?

fwartner:
[Response Types](https://github.com/vuejs/vue-resource/blob/master/docs/http.md#response) HA! i was right! 😆

themsaid:
Yeah I totally believe it's an issue in vue-resource, will need to dig deeper into that though.

fwartner:
@themsaid check out the response-types of vue-resource. There´s no ```data``` object referenced..

taylorotwell:
https://github.com/vuejs/vue-resource/issues/317

themsaid:
@fwartner there's a getter for data that returns `this.body`

PR #86, opened 2016-09-15
Dev master

yoatphotoup:


taylorotwell:
Has no description of what this fixes. Re-submit this when it has a proper description of the problem and how it fixes it.

yoatphotoup:
Hey Taylor, sorry it was negligence on my part, it wasn't meant for merging into laravel:master. Anyways, I would like to take this opportunity to suggest a good feature for Passport. Currently, Passport doesn't allow you to assign scopes to clients, it only allows you to assign scopes to tokens. I think it's good to have a feature that allow us to  assign different scopes to different clients, so when a client requests for an access token with specified scopes, passport should check first whether the requesting client has those scopes assigned to it, and will reject requests if no scopes assigned even though the scopes exist in the application.

PR #88, opened 2016-09-16
Set unstable version

anteriovieira:
Perhaps, this is the ideal moment to create the branch 1.0


taylorotwell:
I've fixed this and created proper branches.

anteriovieira:
@taylorotwell , the same for the other packages?

PR #89, opened 2016-09-17
Introduce PassportServiceProvider::$migrate property to allow fallback to publishing migration.

crynobone:
By default all packages' Service Provider should be able to load custom migration paths, however the developer of each application
should be able to override this by extending the original service provider. As example:

```php
<?php

namespace App\Providers;

class PassportServiceProvider extends \Laravel\Passport\PassportServiceProvider
{
    protected $migrate = false;
}
```

Signed-off-by: crynobone <crynobone@gmail.com>


taylorotwell:
Tweaked this and did a bit different...

Use `Laravel\Passport\Passport::ignoreMigrations()` in your `AppServiceProvider` (or any other provider's) `register` method.

Issue #77, opened 2016-09-11
Fire Passport Events


mikemclin:
What about firing Passport events throughout the process?  There are several reasons why someone might be interested in these details. One might be for logging when someone gets a token (see: https://github.com/thephpleague/oauth2-server/issues/628).

For example: in `Laravel\Passport\Bridge\RefreshTokenRepository::persistNewRefreshToken()`, I added `event('passport.refresh', $refreshTokenEntity->getAccessToken()->getUserIdentifier());` to the bottom of the method, because I needed to know when a refresh token is created, and for whom.

It would be nice if Passport had these types of events natively.

etelford:
I've submitted [PR #90](https://github.com/laravel/passport/pull/90) that would address events when creating Access Tokens.

PR #90, opened 2016-09-17
Use Token model/repository when working with AccessTokenRepository

etelford:
This replaces the usage of using the Database Connection in favor of TokenRepository which then calls `Token` directly.

A big benefit of this is so that Model events can be observed when creating Access Tokens.

I believe this would cover at least part of [Issue #77](https://github.com/laravel/passport/issues/77).

### DIFF
--- a/src/Bridge/AccessTokenRepository.php
+++ b/src/Bridge/AccessTokenRepository.php
@@ -3,7 +3,7 @@
 namespace Laravel\Passport\Bridge;
 
 use DateTime;
-use Illuminate\Database\Connection;
+use Laravel\Passport\TokenRepository;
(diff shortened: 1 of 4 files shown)


mlanin:
@taylorotwell when it will be merged in release? This PR could help a LOT in using Passport with mongodb.

taylorotwell:
It's already been released.

mlanin:
@taylorotwell I mean do you plan to merge it in 1.0 version? Because now it's only available in dev-master build.

livingos:
Please don't. Master branch breaks scope functionality - see #148 

You need an Auth Code Repository too with scopes cast to/from array and then to change FormatScopesForStorage not to encode the scopes. Or just remove cast from AccessToken eloquent model.

PR #104, opened 2016-09-25
Replace connection instance with repositories and add support to change the database connection name

kevindierkx:
This replaces the usage of Database Connection instances for the auth codes and refresh tokens in favor of repositories, similar to: laravel/passport:#90

And this adds the ability to change the connection name used during queries using `Passport::setConnectionName('name-here')`.
_This somewhat resolves my issue mentioned here: laravel/framework#15597_

### DIFF
--- a/src/AuthCodeRepository.php
+++ b/src/AuthCodeRepository.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Laravel\Passport;
+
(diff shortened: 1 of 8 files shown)


kevindierkx:
Changed my mind, thought about something much simpler. Will issue a new PR regarding this topic.

PR #115, opened 2016-09-28
Throw AuthorizationException if user doesn't have scope

Shkeats:
Should this not throw AuthorizationException when the user doesn't have the required scopes as the user is authenticated but doesn't have the scope to access the route?


Shkeats:
guessing this should have been directed at master? More a point for discussion.

taylorotwell:
Yes that is probably correct. Can you update the tests?

Shkeats:
@taylorotwell Test revised.

PR #120, opened 2016-10-05
Show description on scopes list

nueko:


jbrooksuk:
We should show the ID and description in my honest opinion.

taylorotwell:
Ideally I think you should link to documentation about your supported scopes. Full descriptions could be way too long to put in this list.

Issue #84, opened 2016-09-14
revokeOtherAccessTokens does not revoke/delete Refresh Tokens


naabster:
I am using the password grant. When I request a new access token, all existing access tokens are revoked / deleted.

But the refresh tokens for the revoked access tokens are not revoked / deleted. So you end up with lots of not-revoked refresh tokens pointing to non existing or revoked access tokens in your Database (depending on `Passport::pruneRevokedTokens();` ).

As an aside: see #83 - I think an option to allow multiple access tokens for one client_id would be really nice (say iPad, iPhone and Android logged in at the same time)!

craigpaul:
It would seem to me in a perfect world, people wouldn't request new access tokens, they would use their existing refresh token. Following that flow would mean the token becomes `revoked` or deleted like you mentioned based on `Passport::$pruneRevokedTokens`. In that same breath though, of course we don't live in a perfect world, people aren't all smart.

The simplest solution, in my mind, would be to set up a scheduled job to prune the refresh tokens based on either `revoked` equaling true or the `expires_at` timestamp being passed. Whether that should be included with Passport or not, I'm not sure.

The other avenue would be probably to submit a PR to change `Laravel\Passport\TokenRepository` and update the `revokeOtherAccessTokens()` method to handle both types of tokens, that might get a bit messy though.

jlozano254:
Maybe a new static field can be added for example `Passport::$allowMultipleAccessTokens`.

And prevent to call method `revokeOtherAccessTokens()` if it's true...

http://stackoverflow.com/questions/39467566/laravel-passport-multiple-connexions-via-password-client/39802823#39802823

themsaid:
Using a cron job should do the trick.

pierrocknroll:
Please see https://github.com/pierrocknroll/lavarel-passport-purge-command

PR #121, opened 2016-10-06
Allow the many Grant Type Password Access Token's

krcgk:
Perhaps safety problems may occur. Do you think such a suitable option?

ref #84


taylorotwell:
What is the point of revoking them in this case? Can't you setup a cron job to remove expired tokens if you want?

taylorotwell:
Closing pending inactivity.

PR #139, opened 2016-10-14
Allow to configure ApiTokenCookie name

SebastianS90:
This can be useful if you have two Laravel apps on the same domain and do not want to restrict cookies to a specific path.


taylorotwell:
I think it's not that great of an approach to be injecting the configuration everywhere just to do this. I would put a static method on the Passport class like `Passport::cookie('foo')` and if you call it like `Passport::cookie()` just have it return the value. Then you can call that from anywhere. Would you mind opening up a fresh PR with that approach? Thanks!

PR #167, opened 2016-11-02
add UUID support

heisian:
In this pull request we still use the original `id` as the primary key in the `oauth_clients` table.  However, we've added a `char(36)` column called `uuid` right before the `secret` column.

The only difference here is that instead of using the value of `oauth_clients.id` to pass in as the `client_id` for authentication, we instead pass in the value found in the `oauth_clients.uuid` column.

Thus, we can make a request like so:
```
curl -X POST \
    www.my-app.com/oauth/token \
    -d grant_type=password \
    -d client_id=e961c6dc-3761-4d6f-bd7b-681c3e264deb \
    -d client_secret=uWmHeqsjhMPgNXt7rYTnfDhxdGdOKNKK5gRvK9PC \
    -d username=user@email.com \
    -d password=myreallysecurepassword
```
... and it will pass client authentication.

**CAVEAT:**
I added another dependency (which ideally we'd avoid altogether): `webpatser/laravel-uuid` that is used to generate the version 4 UUID when a new client is generated.  If anyone has something simpler, please chime in.

**How to introduce this without making breaking changes?**
We will need to add in a configuration like `Passport::useClientUuids()` that will enable or disable these features.  I'm not sure how to do this without making what I've modified more complicated (yet), so if anyone's got any ideas let's hear'em.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "league/oauth2-server": "~5.0",
         "symfony/psr-http-message-bridge": "^0.3.0",
         "zendframework/zend-diactoros": "~1.0",
-        "phpseclib/phpseclib": "^2.0"
+        "phpseclib/phpseclib": "^2.0",
+        "webpatser/laravel-uuid": "2.*"
     },
     "require-dev": {
         "mockery/mockery": "~0.9",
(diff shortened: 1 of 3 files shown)


taylorotwell:
Breaking change and seems fairly unnecessary at this point.

PR #168, opened 2016-11-04
Add option for using UUID client IDs (non-breaking)

heisian:
Introduce client UUIDs as a non-breaking change #14 

## On existing installations
- Add `Passport::useClientUUIDs()` to `app/Providers/AuthServiceProvider.php` to enable
- run `php artisan passport:uuid` to add a `uuid` column and generate UUIDs for existing clients.

## New installations
- Add `Passport::useClientUUIDs()` to `app/Providers/AuthServiceProvider.php` to enable


Modelizer:
What if for those who already started using `passport` and have 100s of users?

How to migrate those users once this PR is been merged?

taylorotwell:
I'm just not sure I see a big need for this at the moment.

Modelizer:
It was a big step to change id to UUID. @heisian I recommend it can be discussed on stack @larachat #internals channel before submitting a PR.

phoenixgao:
@heisian really need this, thanks for saving my day

heisian:
Sure, just be careful with this fork.  Although unit tests all pass for it, I've not manually tested certain things like the authorization redirect. I mainly use OAuth for a stateless API, a la Github accesse tokens.

MicroDroid:
I can't believe Taylor rejected a pull-request that adds an **optional** feature than many people want, just because he thinks it is not necessary.

victorcastro:
@taylorotwell you have to consider. Please

phoenixgao:
@MicroDroid Lol, how surprising! Well I could tell it's not the first time and won't be the last

mosleim:
uuid is not working for new version?

veelasky:
@mosleim it's not even merged.

heisian:
You all may want to check out @ssmulders package https://github.com/ssmulders/hashed-passport

It essentially is middleware that will hash and unhash the client ID without changing the way it is stored in the database. It is a non-invasive method of handling what we want here and actually I think a better solution than what I wrote.

Use caution as always, however, I haven't tested the package myself. Cheers all!

mohamedsharaf:
i wish if uuid is supported

Modelizer:
@heisian don't you think so hashing and unhashing every time will put load on server?

grEvenX:
@heisian Even @ssmulders's package doesn't cover all cases well and as long as this feature is not built-in, it will be hard to keep everything working as expected for something that seems pretty trivial to support (it can be discussed if it even should be the default as most OAuth references advice against having this as a guessable sequential number).

One case this package doesn't support is when using HTTP BASIC auth for sending client-id/client-secret (used in machine-to-machine/application context to retrieve a non-personal access token for a client) because the package relies on intercepting the request as a middleware and looking for `client_id` parameter.

PR #173, opened 2016-11-08
Modifying UserRepository so it has another option if the User's class password isn't a standard hashed password.

huntsfromshadow:
An issue with the Password Grant is it assumes
that you are using a user model with a hashed password.

This may not be true depending on a number of use cases.

This pull request offers an option similar to what was done with findForPassport earlier in getUserEntityByUserCredentials.

If the code sees a method 'validateForPassportPasswordGrant' in the user class it calls that and lets
the class validate the password sent into the password grant.

First time submitting to a laravel project so if something is wrong in terms of formatting, naming, or something else please let me know so I can fix it.

### DIFF
--- a/src/Bridge/UserRepository.php
+++ b/src/Bridge/UserRepository.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport\Bridge;
 
 use RuntimeException;
+use Illuminate\Support\Facades\App;
 use Illuminate\Contracts\Hashing\Hasher;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 use League\OAuth2\Server\Repositories\UserRepositoryInterface;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you share a bit more about your use case for this?

huntsfromshadow:
Sure.

We are developing an in house API that connects to a legacy system.  Due to how the architecture of the system the API won't be directly talking to the legacy DB but will be talking through a code layer used by both a legacy webapp and the new api layer (built in laravel). All laravel can do is ask the layer beneath it 'Please verify the following email/password and let me know if it is valid'.  

The laravel layer doesn't have specific knowledge on 1) where the fields are located in the db and 2) what hashing/encryption is used on the password in the lower layer.

Issue #166, opened 2016-11-02
Passport::pruneRevokedTokens() broken


bhosie:
@taylorotwell may i suggest this commit be reverted and the idea revisited in a different way? https://github.com/laravel/passport/commit/b95791d42f42d405ffacdf21007f80ef7798b164

 Since `$this->revokeOtherAccessTokens()` no longer runs, `Passport::pruneRevokedTokens();` is now broken. 

To reproduce:

- [ ] in AuthServiceProvider.php

```php
 Passport::routes();
 Passport::pruneRevokedTokens();
```

- [ ] now, make a call to `/oauth/token`
- [ ] verify that tokens are not getting pruned from the db
- [ ] manually change a token to `revoked=1` in the db. Call `/oauth/token` again and confirm that  the revoked token did not get pruned

PR #176, opened 2016-11-10
re-enable revoking access tokens

bhosie:
Having this line commented out creates 2 problems: the datastore can potentially fill up with many live tokens, and it breaks pruning revoked tokens.

To reproduce (before including this patch):

- [ ] in AuthServiceProvider.php

```php
 Passport::routes();
 Passport::pruneRevokedTokens();
```

- [ ] now, make a call to `/oauth/token`
- [ ] verify that tokens are not getting pruned from the db
- [ ] manually change a token to `revoked=1` in the db. Call `/oauth/token` again and confirm that  the revoked token did not get pruned

closing https://github.com/laravel/passport/issues/166 in lieu of this PR

### DIFF
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -70,7 +70,7 @@ class AccessTokenController
         $payload = json_decode($response->getBody()->__toString(), true);
 
         if (isset($payload['access_token'])) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Instead of just uncommenting this again, maybe we should just address the root problem so that everyone is happy? This will just break applications that depend on other access tokens not being revoked when a new one is issued?

Issue #47, opened 2016-08-23
Laravel Passport Personal Token failing Validation (exp=0)


gbdematos:
I'm trying to create a Personal Access Token to use on Postman.

I followed the same steps as the video from "What's new in Laravel 5.3" where Taylor does that, but I keep getting "Unauthorized" on Postman...

By copying and pasting the generated token on jwt.io, I'm getting "exp" value = 0. Trying to get to the root of the error, I found out that on the "validation" process, it fails exactly on the "exp" validation.

Does anyone have any idea what is happening?

My topic on Laracasts:
https://laracasts.com/discuss/channels/laravel/laravel-passport-personal-token-failing-validation-exp0

Another user with the same problem:
https://laracasts.com/discuss/channels/laravel/laravel-passport-and-postman-gets-all-the-time-unauthorised

Jonastouw:
I ran into the same issue, and finally figured out how to make it work. The expiry date is set to now + 100 years in Passport.php, line 167.

`return static::$tokensExpireAt ? Carbon::now()->diff(static::$tokensExpireAt) : new DateInterval('P100Y');`

If you set it to, i.e., P1Y, it is working. 
Something like:

``return static::$tokensExpireAt ? Carbon::now()->diff(static::$tokensExpireAt) : new DateInterval('P1Y');``

I guess the 100 years DateInterval yields an int (seconds) somewhere along the way that the exp value is not expecting, setting it to zero and invalidating it.

gbdematos:
@Jonastouw I'm trying to do this you described, but no success... Even after altering P100Y to P1Y, the generated tokens are being set to 2116 on the database (and exp to 0). 

Is your issue also with Personal Tokens?

I'm thinking how on earth are people not having this issue? I found no reports except mine and another user's. Is it something I did that caused this?

Jonastouw:
@gbdematos I just updated my post to include the personal tokens. Have you tried setting those as well?

gbdematos:
@Jonastouw it worked now, modifying PassportServiceProvider.php line 84!! Can't believe it, after two days of headache! Thank you so much!

But I'm still curious, is this normal behavior or something I did or my environment maybe? :/

Should I let this Issue open for verification?

nueko:
Hmm, is this the [Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem) fx :1234: ?

roulendz:
I did as needed, and OAuth client started to work!
But still
[...]

PR #179, opened 2016-11-11
fix token expiry issue #47

seahken:
Changed token expiry from P100Y to P1Y


taylorotwell:
They are mean't to be long lived by default and you can customize it from there.

PR #181, opened 2016-11-11
static setter for revoking other tokens

bhosie:
see https://github.com/laravel/passport/pull/176#issuecomment-259957727

how about something like this?

### DIFF
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -69,8 +69,8 @@ class AccessTokenController
 
         $payload = json_decode($response->getBody()->__toString(), true);
 
-        if (isset($payload['access_token'])) {
-            // $this->revokeOtherAccessTokens($payload);
+        if (isset($payload['access_token']) && Passport::$revokeOtherTokens) {
+             $this->revokeOtherAccessTokens($payload);
         }
 
         return $response;
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -17,6 +17,13 @@ class Passport
     public static $pruneRevokedTokens = false;
 
     /**
+     * Indicates if Passport should revoke existing tokens.
+     *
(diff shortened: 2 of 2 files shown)


taylorotwell:
So, what is the purpose of revoking all other access tokens when another access token is issued for the client? I think the current setup's intention is that you will setup your own cron job to take care of managing your data store token pruning of expired token.

bhosie:
That's a great question. The purpose of revoking all other access tokens is to allow a client to invalidate all of their existing tokens by simply requesting a new one. This is especially useful in a `client_credentials` scenario. I made the assumption that this was intentional default behavior because it was in place up until v1.0.4 before is was quietly removed (thereby breaking dependent applications).

It makes it easy for a client to regenerate a token and know that any previous tokens would be invalidated/revoked. As it stands now, the client could continually make requests to the `/oauth/token` route and generate many tokens that effectively allow the same access without understanding that this is what is actually happening. 

It's possible i completely misunderstood the purpose of commenting the line out and disabling the original default behavior, and maybe this 'feature' was actually an unintentional bug. Regardless, it's very useful, and I'm proposing we allow it to be enabled without it being the default. Why introduce the extra overhead of a separate cron when a setter could easily support either behavior?

Secondly, we need to get back allowing pruning of the revoked tokens via `Passport::pruneRevokedTokens()` when a new token is issued. This documented feature is currently broken.

` If you would like Passport to automatically delete your revoked tokens, you should call the pruneRevokedTokens method from the boot method of your AuthServiceProvider: ...  revoked tokens will be deleted when a user requests a new access token ...` https://laravel.com/docs/5.3/passport#pruning-revoked-tokens

Back to the first issue (to support the value of the second issue) - it seems more valuable to be able to prune revoked tokens when a new one is issued if the previous tokens were actually revoked during the re-issue.

PR #183, opened 2016-11-14
Fix csrf validation

mleczek:
Allow request `_token` input and `X-XSRF-TOKEN` header. The same as in the [VerifyCsrfToken::tokensMatch](https://github.com/laravel/framework/blob/5.3/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L108) method.

This allow us instead of using (the only valid method now):
```JavaScript
Vue.http.interceptors.push((request, next) => {
    request.headers.set('X-CSRF-TOKEN', Laravel.csrfToken);

    next();
});
```

To use below method (my favourite one):
```JavaScript
Vue.http.interceptors.push((request, next) => {
    request.headers.set('X-XSRF-TOKEN', getCookie("XSRF-TOKEN")); // where getCookie returns encrypted X-XSRF-TOKEN cookie

    next();
});
```
... or of course the `_token` input.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -200,8 +200,14 @@ class TokenGuard
      */
     protected function validCsrf($token, $request)
     {
+    	$reqToken = $request->input('_token') ?: $request->header('X-CSRF-TOKEN');
+
+        if (! $reqToken && $header = $request->header('X-XSRF-TOKEN')) {
+            $reqToken = $this->encrypter->decrypt($header);
+        }
+    	
         return isset($token['csrf']) && hash_equals(
-            $token['csrf'], (string) $request->header('X-CSRF-TOKEN')
+            $token['csrf'], (string) $reqToken
         );
     }
 }


taylorotwell:
I don't see any bug or problem with current implementation.

mleczek:
```PHP
// routes/web.php
Route::post('/user', function(Request $request) {
    return $request->user();
})->middleware('auth');

// routes/api.php
Route::get('/user', function (Request $request) {
    return $request->user();
})->middleware('auth:api');
```

```JavaScript
// resources/assets/js/bootstrap.js
Vue.http.interceptors.push((request, next) => {
    request.headers.set('X-CSRF-TOKEN', Laravel.csrfToken);

    next();
});
```

```JavaScript
// resources/assets/js/components/Example.vue
this.$http.post('/user') // OK, CSRF valid
this.$http.get('/api/user') // OK, CSRF valid
```

Now when I switched to:
```JavaScript
// resources/assets/js/bootstrap.js
Vue.http.interceptors.push((request, next) => {
    request.headers.set('X-XSRF-TOKEN', getCookie('XSRF-TOKEN'));

    next();
});
```

I got following results:
```JavaScript
// resources/assets/js/components/Example.vue
this.$http.post('/user') // OK, CSRF valid
this.$http.get('/api/user') // 401, because X-CSRF-TOKEN header doesn't exist
```

I found it a bug or at least the inconsistency (it's hard to say if the word "fix" used in the PR title was appropriate). I know this is a separate library, but serve as part of Laravel and [for me] it should be compatible with [this part of documentation](https://laravel.com/docs/5.3/csrf). Although [there isn't](https://laravel.com/docs/5.3/passport#consuming-your-api-with-javascript) anything about it.

Issue #162, opened 2016-10-30
Personal Access Token expiration can't be set and it fails to check the Y2K38 bug.


fgoni:
Using the Passport facade to set the Expiration time for Authorization Code and Password Grant you can avoid the dreaded Y2K38 bug but Personal Access Tokens are hardcoded to last another hundred years from now, which puts it well beyond 2038.

This is a problem because on most Windows environments the tokens will never match and never authenticate the user.

The only way to bypass this is to modify the vendor files and do it every time Passport is upgraded.

nugrahawahyu:
Yeah I always get negative value of expires_in ( -1477799197  ) and the only way to fix it is to modify the vendor files.

zsimple:
Yes, I just wasted one day to find this problem, It happend only happend on 32bit php.

For now, maybe we need modify passport/src/Passport.php @190, `new DateInterval('P100Y')` --> `new DateInterval('P1Y');`.

100 Years is really too long :-D

badknight:
> Yeah I always get negative value of expires_in ( -1477799197 ) and the only way to fix it is to modify the Vendor files.

@nugrahawahyu you can simply change the [token-lifetime](https://laravel.com/docs/5.3/passport#token-lifetimes) - it solved my problems

```php

    /**
     * Register any authentication / authorization services.
     *
     * @return void
     */
    public function boot()
    {
        $this->registerPolicies();
        
        Passport::routes();
        Passport::tokensExpireIn(Carbon::now()->addDays(30));
    }
```

fgoni:
TokensExpireIn 'only' works for Authorization Code and Password Grant tokens, not for personal   access tokens.

themsaid:
a PR was submitted to solve the issue: https://github.com/laravel/passport/pull/185

diegcode:
To change the expiration date of a PersonalAccessToken could create a class that extends from the original class of passport. for example: To add 15 minutes

<?php

namespace App;

use Carbon\Carbon;
use Laravel\Passport\PersonalAccessTokenFactory as PersonalAccessTokenFactoryPassport;
use Laravel\Passport\PersonalAccessTokenResult;

class PersonalAccessTokenFactory extends PersonalAccessTokenFactoryPassport
{
    public function make($userId, $name, array $scopes = [])
    {
        $response = $this->dispatchRequestToAuthorizationServer(
            $this->createRequest($this->clients->personalAccessClient(), $userId, $scopes)
        );

[...]

PR #185, opened 2016-11-14
Expire in 1 year by default

themsaid:
In reference to https://github.com/laravel/passport/issues/162 and https://github.com/laravel/passport/issues/47

100 years seems to be too long for some environments, maybe 1 year is enough?

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -204,7 +204,7 @@ class Passport
         if (is_null($date)) {
             return static::$tokensExpireAt
                             ? Carbon::now()->diff(static::$tokensExpireAt)
-                            : new DateInterval('P100Y');
+                            : new DateInterval('P1Y');
         } else {
             static::$tokensExpireAt = $date;
         }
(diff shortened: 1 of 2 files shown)


Modelizer:
Will it be a better idea to set year in `.env` file? I think this will give more flexibility for users who wants to alter.

hinaloe:
Why the personal token limit static 1year?

diazemiliano:
Hi guys, I've set the `expires_at` date to `one year` before now. But it seems Passport isn't checking expiration dates on `personal access tokens`. That's a default behavior? Thanks.

MightyPork:
I found this workaround for the unfortunate fixed 1 year expiry.

After creating a token:

```
$tokenResult = $user->createToken($name);
$tokenResult->token->expires_at = Carbon::now()->addDays($expiryDays);
$tokenResult->token->save();
```

not sure who thought 1 year is enough, it's really inconvenient to have to regenerate your tokens every year.. this way, if I understand it right, you can set any arbitrary expiry period.

Issue #165, opened 2016-11-01
Password Grant and independant SPA issue


Joshgallagher:
Hello all,

I have been messing around with Passport and specifically the Password Grant. The reason for this is I plan to build and API with Laravel and Passport that will consume an SPA. The SPA will not be part of Laravel - not built within Blade views.

The main issue I am running in to is that I am required to share both the Client ID and Secret. However, according to the OAuth spec the Client Secret should never be exposed to the end user as it will make the flow redundant I believe.

I believe this is an issue because Passport has not been provided with a way to combat this security issue. I have read that a proxy should be used to inject the Client Secret into the POST/GET request and then sent to the API, however, I am not sure if this is correct.

It would be a great help if anyone had any ideas on how to deal with this issue.

Also, if you would like me to elaborate further please ask!

craigpaul:
Hey @Joshgallagher, sounds to me like you just have the wrong idea about what oAuth grant to be using. For a SPA (aka anything sending requests from the client side) ou should be using the implicit grant, not the password grant.

Granted it's not in the documentation, but it is in passport, not sure the exact release, but I can find it later if no one else finds it. You can read about the implicit grant here though.

https://oauth2.thephpleague.com/authorization-server/implicit-grant/

Joshgallagher:
Hey @craigpaul 

I do know about the Implicit Grant but did not know it was available in Passport. Do you have any idea when the docs will be updated? Also, to clear this up - from what I have previously read you should use the Implicit Grant when communicating to a web app/SPA and the Password Grant should be used in mobile apps? I have seen and read that you can use the Password Grant and a proxy for SPAs, as it allows you to use the refresh token due to it being encrypted in a cookie?

craigpaul:
Hey, turns out I was wrong, it never got reopened, my mistake. You are right, you could use the Password Grant like that, but I'm not sure how you would do that. It's much easier to just use the Implicit Grant. The PR that was originally proposed was #63 so I guess we just need to get Taylor to take another look at that and get it merged. I believe his requirements to merge it have been met so it should be good to go.

Again, sorry my mistake about thinking that was already merged.

Joshgallagher:
Thats fine, I struggled to find it. 

I think the Implicit Grant should be implemented ASAP! I've seen many people implementing the Password Grant incorrectly in live projects which is not good. Also, if Taylor does not want to merge the Implicit Grant PR, then maybe he should look into providing some middleware that can proxy the Password Grant.

[...]

PR #191, opened 2016-11-17
Adds middleware to attach client credentials

craigpaul:
In reference to #165. I imagine this will work by adding this middleware before your `auth:api` call in your main application's routes. Basically you can make a POST request with the following parameters.

```js
{
  grant_type: 'password',
  client_id: 1,
  scope: '',
  username: 'email@example.com',
  password: 'secret'
}
```

The middleware will then attempt to find the client in it's database, otherwise a `ModelNotFoundException` will be thrown. If it finds the client, it will attach the client secret to the request. This will allow Passport's `TokenGuard` to continue with handling authentication assuming the username and password were correct.

### DIFF
--- a/src/Http/Middleware/AttachesPasswordGrantCredentials.php
+++ b/src/Http/Middleware/AttachesPasswordGrantCredentials.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Laravel\Passport\Http\Middleware;
+
+use Closure;
+use Laravel\Passport\Client;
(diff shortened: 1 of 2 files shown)


Joshgallagher:
Does it return the access and refresh token in a HTTPS cookie?

craigpaul:
No, going to rethink this tomorrow with a fresh mind.

Joshgallagher:
I will also give it a go. I haven't had much time due to uni, but I will definitely help out with this!

PR #193, opened 2016-11-17
Adds support for password grant cookies

craigpaul:
In reference to #165. This will allow decoupled SPA's to use the `laravel_token` cookie as a holder for a password grant access token. It accomplishes this by having the following configuration set.

AuthServiceProvider.php
```php
public function boot()
{
    Passport::respondWithCookie();

    Passport::routes();

    ...
}
```

Kernel.php
```
protected $routeMiddleware = [
    ...
    'handle-injections' => \Laravel\Passport\Http\Middleware\HandlesGrantParameterInjections::class,
    'add-cookie' => \Laravel\Passport\Http\Middleware\AddsPasswordGrantCookie::class,
]
```

When this is active, the attach middleware will check if they are using the password grant and then attempt to find their client by the passed in `client_id`. If one is not found an exception will be thrown, otherwise the `client_secret` will be added to the request parameter bag.

The call will then go through Passport's existing logic and end up at the add middleware. That will attach an encrypted cookie containing the access token from the existing response and reuse the `Passport::cookie()` to make it "compatible" with the `TokenGuard`.

The cookie isn't inherently secure though as it still relies on the `session.secure` config option.

Once we have our cookie and we are ready to make a request, we will attempt to decode the cookie by first using the public key method (otherwise it falls back to the base implementation necessary for `CreateFreshApiToken` to work. Assuming the user has send a valid token, we then just check if that token contains the `aud` claim, if so, it skips over the `validCsrf` check and proceeds to retrieve a user and finish the request.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -4,7 +4,6 @@ namespace Laravel\Passport\Guards;
 
 use Exception;
 use Firebase\JWT\JWT;
-use Laravel\Passport\Token;
 use Illuminate\Http\Request;
 use Laravel\Passport\Passport;
 use Illuminate\Container\Container;
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -161,12 +161,15 @@ class TokenGuard
             return;
         }
 
-        // We will compare the CSRF token in the decoded API token against the CSRF header
-        // sent with the request. If the two don't match then this request is sent from
-        // a valid source and we won't authenticate the request for further handling.
-        if (! $this->validCsrf($token, $request) ||
-            time() >= $token['expiry']) {
-            return;
+        if ($this->hasNoAudience($token)) {
+            // We will compare the CSRF token in the decoded API token against the CSRF header
+            // sent with the request. If the two don't match then this request is sent from
+            // a valid source and we won't authenticate the request for further handling.
+            if (!$this->validCsrf($token, $request) ||
+                time() >= $token['expiry']
+            ) {
+                return;
+            }
         }
 
         // If this user exists, we will return this user and attach a "transient" token to
@@ -185,10 +188,28 @@ class TokenGuard
      */
     protected function decodeJwtTokenCookie($request)
     {
+        if (Passport::$respondWithSecureCookie) {
+            return (array) JWT::decode(
+                $this->encrypter->decrypt($request->cookie(Passport::cookie())),
+                'file://'.Passport::keyPath('oauth-public.key'), ['RS256']
+            );
+        }
+
         return (array) JWT::decode(
             $this->encrypter->decrypt($request->cookie(Passport::cookie())),
             $this->encrypter->getKey(), ['HS256']
         );
+    }
+
+    /**
+     * Determine if the token does not have an audience attached to it.
+     *
+     * @param  array  $token
+     * @return bool
+     */
+    protected function hasNoAudience(array $token)
+    {
+        return !isset($token['aud']);
     }
 
     /**
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -187,7 +187,7 @@ class TokenGuard
      */
     protected function decodeJwtTokenCookie($request)
     {
-        if (Passport::$respondWithSecureCookie) {
+        if (Passport::$respondWithCookie) {
             return (array) JWT::decode(
                 $this->encrypter->decrypt($request->cookie(Passport::cookie()))['access_token'],
                 'file://'.Passport::keyPath('oauth-public.key'), ['RS256']
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -166,8 +166,7 @@ class TokenGuard
             // sent with the request. If the two don't match then this request is sent from
             // a valid source and we won't authenticate the request for further handling.
             if (!$this->validCsrf($token, $request) ||
-                time() >= $token['expiry']
-            ) {
+                time() >= $token['expiry']) {
                 return;
             }
         }
@@ -190,7 +189,7 @@ class TokenGuard
     {
         if (Passport::$respondWithSecureCookie) {
             return (array) JWT::decode(
-                $this->encrypter->decrypt($request->cookie(Passport::cookie())),
+                $this->encrypter->decrypt($request->cookie(Passport::cookie()))['access_token'],
                 'file://'.Passport::keyPath('oauth-public.key'), ['RS256']
             );
         }
--- a/src/Http/Middleware/AddsPasswordGrantCookie.php
+++ b/src/Http/Middleware/AddsPasswordGrantCookie.php
@@ -5,6 +5,7 @@ namespace Laravel\Passport\Http\Middleware;
 use Closure;
 use Carbon\Carbon;
 use Laravel\Passport\Passport;
+use Illuminate\Encryption\Encrypter;
 use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Response;
 use Illuminate\Contracts\Config\Repository as Config;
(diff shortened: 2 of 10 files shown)


Joshgallagher:
I have noticed that you are only putting the access token in the cookie. You should be storing the refresh token in there also. Then, if the access token expires the refresh token will be available to the authorisation server.

matthewlilley:
A couple of questions on this...

What about 1st party clients that have no concept of a cookie store?

And 1st party clients on a different domains?

There's a good article on this at auth0. https://auth0.com/blog/cookies-vs-tokens-definitive-guide/

craigpaul:
@Joshgallagher Fair point, I will see what I can figure out regarding that.

@matthewlilley Are you referring to mobile apps? Even with this enabled it still sends back the JSON response containing the access token and refresh token. Although you do get the cookie in that situation too, so I suppose one of two things could be done.

1. You manually override the `oauth/token`  route and wrap it with the `AttachesPasswordGrantCredentials` middleware.

2. We change this to use two different "flags" instead of just `Passport::respondWithSecureCookie()` to control which middleware get attached to that route.

matthewlilley:
Correct me if I'm wrong, this is my understanding of it 

.A client should be able to receive one JWT token, and that be stored in localstorage if they choose. This is a common pattern with decoupled SPAs, native apps etc... Because of the issues with sending cookies cross domain, and in the case where the client has no concept of a cookie store. I think the cookie method here is valid, but it doesn't solve the use case I mention.

craigpaul:
@matthewlilley I completely agree with you, the cookie method wouldn't work in your use case. If I remember correctly, all you mentioned was that you were wanting some way of having a "proxy" that would inject *both* the client id and secret. 

Now, this doesn't inject the client id, it expects it to be passed to it, but it will inject the client secret assuming the id you pass is valid and is for a password grant. Then you would receive a JSON response and you can continue going on just like normal with your access token and refresh token. Unless I'm missing something, the only thing wrong (in your use case in regards to this PR) is that it will always try to attach the cookie to the response as well as inject the client secret, instead of just injecting the client secret. Is that correct? :)

Joshgallagher:
The whole point of this middleware is to stop people storing a JWT or any type of tokens in local storage. It is bad practice, imagine if someone steals them... local storage is not safe. The cookie approach has its fallbacks, but not as many as local storage. Implementing CSRF will negate one of the main problems with cookies.

@matthewlilley could you elaborate on your scenario? What clients are you using that do not have any concept of persistent cookies stores?

Modelizer:
@Joshgallagher Can you explain how `localStorage` are not safe? because when we build SPA if we want to store for a long run we need to use local storage. Regarding HTML 5 `state`, it is kind of tempory storage which needs to be sync with local storage.

Joshgallagher:
Its a bad idea to store your access and refresh token in plain text. This is because local storage isn't safe and attackers can get the tokena from it using a variety of attacks. That's why I suggest storing the credentials in a HTTPS cookie, this way it can be used for a long time also.

Joshgallagher:
@craigpaul Does the cookie have to be sent on each request with your middleware? Also, do you think theres a high possibility of this being merged?

craigpaul:
@Joshgallagher The cookie gets sent to you when you send a request to `/oauth/token`. Then you have to send the cookie on each request to your api just like if you were using the access token in the `Authorization` header.

As for if it will get merged, I have no idea whether they want Passport to handle this or not, so I can't say one way or the other.

Joshgallagher:
@craigpaul I think we will have to cross our fingers on this one. I do not see why @taylorotwell would not merge this PR, all it does is cover people who want to separate logic into an SPA and API. There is already a special middleware integrated into the Passport framework - CreateFreshAPIToken, which allows SPA functionality to be built into blade views. So, should there not be a middleware provided for this situation?

craigpaul:
@Joshgallagher they aren't part of the `auth:api` middleware, but the `TokenGuard` was modified to accept this cookie, so using `auth:api` you can accept this cookie just like normal.

taylorotwell:
Can you use the new implicit grant for this?

craigpaul:
@taylorotwell You could totally use the implicit grant, I think the reason @Joshgallagher brought it up is because of the views of how [insecure localStorage is for storing access tokens and refresh tokens](https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage). 

Now personally, I just use the out of the box middleware since I build stuff normally with blade and vue, but if someone was using Laravel as the API and a completely decoupled (from Laravel) SPA they would have to, right now, use the Implicit Grant and have to store in localStorage.

This implementation might need more thought though as I am unsure how you could implement CSRF with a Password Grant to secure that cookie completely.

Joshgallagher:
@taylorotwell This could be used with the Implicit Grant type, however, that grant is only usually used for 3rd parties. This middleware would allow developers to use the Password Grant Type for their 1st party apps.

The problem is that the Implicit grant Type is unsafe, therefore, a short expire time must be set. This means that the usability of an app with this Grant is poor. However, by implementing this middleware it allows a developer to use the Password Grant Type in their decoupled SPA as the access and refresh token is stored in an encrypted cookie. Also, I have seen in many projects the Client Secret has been stored in plain text... as you know this is bad and breaks the whole point of the grant. Again, this middleware will allow an SPA to use the access and refresh token, which in turn will remove the hindrance of the Implicit Grant Type.

I do believe this should be strongly considered. This was, also suggested by Alex Bilbie!

Modelizer:
@craigpaul even I'm not sure how it will be implemented CSRF with passport grant? If we implemented CSRF  then we need to start a session I guess. Then there will be no point of creating a cookie. Correct me if I'm wrong.

Also, I think can we brainstorm and come up with a more simple solution? It seems we need to compromise with flexibility in terms of security or vice-versa.

I would suggest creating a personal token and assigning a name which will be user IP. When a SPA request to API then API cross checks it with an IP. This way if a hacker gets the access token still he will not be able to process further due to access_token is bound to an IP.

Note: 
1. The reason I emphasize on the personal token is I don't count SPA as a third party app.
2. I don't have a proper idea that at what time interval does ISP give a new IP to a system. I believe it would be done when the user restarts the system.

themsaid:
@Modelizer I feel we're overcomplicating Passport, I understand that security is an important issue but imho we should stick to the basic grant types and let people build on top of it in their apps because there are a lot of concerns that depends on the nature of every app.

craigpaul:
I agree @themsaid, you beat me to it lol. This is too much complexity for what Passport should be. This logic could be easily implemented on top of Passport, in your own or in a separate package. I think this should be closed unless anyone has anything other to say about it.

@Modelizer in regards to the CSRF token comment I made, I was reading through that article by alex and the comments left in #165 regarding a CSRF token. You are right to my knowledge, you would need a session, so that's out the window.

Joshgallagher:
@Modelizer I believe an IP can be spoofed? If so, then this won't work.

As for the CRSF problem, the double submit cookie pattern could be used which is stateless.

@themsaid and @craigpaul. I see your point about over complicating Passport. However, this is just like the createFreshApiToken middleware. The whole problem here is that Passport can't be used for decoupled SPA's, and from what I have seen that is the new thing - backed API and front end SPA. Plus, this can be quite daunting for some people to implement in their projects, which means people will either implement OAuth wrong and compromise their whole app or they just won't use Passport. I think this should be implemented because it's stopping a large majority of people from creating decoupled SPA's.

Modelizer:
I totally agree with @themsaid as well as @Joshgallagher 

I do see there will be a gap between decoupled SPA's and Passport. Also, I think it's getting complicated. But still, there is no way out educating people when they build SPA's.

I'll go with @Joshgallagher because I believe SPA's is the future! and we should provide at least basic support in Passport.

Now totally depends on @taylorotwell whether he would like to see it implemented or not? 🤔 

As far as double submit cookie, @craigpaul is it implement in this PR?

craigpaul:
@Modelizer No it is not implemented, but it could be. I still am against my own PR though, I think this would be better implemented as a separate package on top of Passport so we don't muck up the codebase.

Joshgallagher:
@craigpaul I do not think it would muck up Passports codebase, only enhance it. Like I have pointed out, Taylor created a special middleware for SPA's inside of blade views. However, this middleware is more relevant as it invites more front end devs who want to work with separate code bases in their projects. Also, like I have said people are not implementing it correctly and who will create a package and maintain it?

taylorotwell:
I don't want to invent new authentication methods if possible.

renanwilliam:
this PR would be nice for Angular apps, it's a recurring issue

Issue #16, opened 2016-08-18
Bypass auth screen if we already have a token


jbrooksuk:
Unless the token has expired, the consumer has explicitly requested re-auth or the scopes have changed, we shouldn't need to continue showing the auth screen.

marcoscoelho:
Same question..

geodeveloper:
@taylorotwell can you take a look at this issue ? This seems problematic when an app it's depending only on oAuth server to authenticate users. And it's annoying to ask each time the user to authorize (even if previously has approved the request).

nerfe:
Does anybody have a workaround on this yet? What i am planning is 

-Store Authorization Code of the user (in my own REST api)
-Create an endpoint to use Authorization Code of user to generate new token and continue login
-If authorization code is expired then show the authorization page

I am using Laravel Passport with my own REST api services.

PR #199, opened 2016-11-21
Bypass auth screen if we already have a token

themsaid:
In reference to https://github.com/laravel/passport/issues/16

This PR skips the approval screen if the user has a valid token for the same client that has the same scope as the scope requested, Passport will auto-approve the request and return a new token.

### DIFF
--- a/src/TokenRepository.php
+++ b/src/TokenRepository.php
@@ -2,6 +2,9 @@
 
 namespace Laravel\Passport;
 
+use Carbon\Carbon;
+use Illuminate\Database\Eloquent\Model;
+
 class TokenRepository
 {
     /**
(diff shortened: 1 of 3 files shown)


src/TokenRepository.php
```diff
@@ -27,6 +30,22 @@ public function save($token)
...
+     * @param  Model  $userId
+     * @param  Client  $client
+     * @return Token|null
+     */
+    public function getValidToken($user, $client)
+    {
+        return $client->tokens()
+                      ->whereUserId($user->id)
+                      ->whereRevoked(0)
+                      ->where('expires_at', '>', Carbon::now()->toDateTimeString())
```
taylorotwell: I'm not sure you really need to make a string, you can just pass in the Carbon instance.

themsaid: Updated.

jbrooksuk:
Yay! Thanks for this @themsaid!

themsaid:
@jbrooksuk welcome :)

raftalks:
@themsaid is this no longer supported on latest version?

Issue #198, opened 2016-11-20
scope/s auth checks throw dedicated error


di5abled:
Hello,

when a user can't access a given endpoint due to the scope/s middleware I want to inform him which scope/s failed. Probably the best way would be to make an own exception extending the current AuthenticationException. Then I could check for this in the exception handler and build an response containing all needed infos.

Is this something we could/should handle via passport or should I just create my own middleware?

craigpaul:
Seems like a good idea to me, could be an InvalidScopesException that can be passed an array or string of the scope(s). You could make a PR and see what you come up with, seems like something Passport could benefit from.

Keep in mind that the exception handler only triggers report, not render so you wouldn't be able to catch it in the main applications handler using this custom exception. Passport would have to have a custom generic response that would list out the scopes.

anhducbkhn:
scope failed for me.
Also, when I input an incorrect access_token, it does  not response a json message, it respon a html page.
The issue is in CheckScopes:

class CheckScopes
{
    /**
     * Handle the incoming request.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @param  array  $scopes
     * @return \Illuminate\Http\Response
     */
    public function handle($request, $next, ...$scopes)
    {
        if (! $request->user() || ! $request->user()->token()) {
            throw new AuthenticationException;
        }

        foreach ($scopes as $scope) {
            if (! $request->user()->tokenCan($scope)) {
                throw new AuthorizationException;
            }
        }

        return $next($request);
    }
}

$request->user() will be null if access_token is incorrect.

themsaid:
@di5abled are you willing to make a PR?

craigpaul:
@di5abled @themsaid I had some time so I went ahead and made a PR, but feel free to make another if it's not what you were thinking 👍

craigpaul:
@di5abled As the PR was merged, I hope that helps with your use case. Please don't forget to close this issue :) thanks!

themsaid:
@craigpaul  👍♥️

PR #202, opened 2016-11-24
Adds invalid scopes exception

craigpaul:
In reference to #198. This adds the `InvalidScopesException` to allow the user to send a custom response showing which scope(s) have failed authorization. You can catch this exception in your application's exception handler as shown below.

```php
use \Laravel\Passport\Exceptions\InvalidScopesException;

...

/**
 * Render an exception into an HTTP response.
 *
 * @param  \Illuminate\Http\Request $request
 * @param  \Exception $exception
 *
 * @return \Illuminate\Http\Response
 */
public function render($request, Exception $exception)
{
    if ($exception instanceof InvalidScopesException) {
        return response()->json(['scopes' => $exception->scopes()]);
    }

    return parent::render($request, $exception);
}
```

### DIFF
--- a/src/Exceptions/InvalidScopesException.php
+++ b/src/Exceptions/InvalidScopesException.php
@@ -2,9 +2,9 @@
 
 namespace Laravel\Passport\Exceptions;
 
-use Exception;
+use Illuminate\Auth\Access\AuthorizationException;
 
-class InvalidScopesException extends Exception
+class InvalidScopesException extends AuthorizationException
 {
     /**
      * The scope(s) that failed authorization.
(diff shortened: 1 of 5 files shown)


taylorotwell:
You might want to extend AuthorizationException or something. With this change the exception will no longer be rendered with a  403 response but instead with a 500 I believe.

craigpaul:
@taylorotwell Ya i was thinking about that the other day but I figured I would wait to see what you had to say. You definitely are correct, it would yield a 500 instead of a 403, which of course would be more appropriate, either way, its updated now and yields a 403 upon response upon failure.

PR #214, opened 2016-12-21
Fixed Auth::user() exception

johnnyfreeman:
There is an issue where calling `Auth::user()` will throw an `Exception` if the access token was generated by `ClientCredentialsGrant`. What happens is the TokenGuard tries to `$this->provider->retrieveById` with an id of `""`.

Admittedly, I don't know the library super well. So, even though this fixed the issue for me, it could use an experienced set of eyes to make sure I'm not causing any breakage elsewhere.


taylorotwell:
I've fixed this. Modified the code a bit.

PR #216, opened 2016-12-22
Customisable Authentication Model

gauravmak:
users provider was set statically in the code. Changed to use one specified for api configuration

### DIFF
--- a/src/Bridge/UserRepository.php
+++ b/src/Bridge/UserRepository.php
@@ -32,8 +32,10 @@ class UserRepository implements UserRepositoryInterface
      */
     public function getUserEntityByUserCredentials($username, $password, $grantType, ClientEntityInterface $clientEntity)
     {
-        if (is_null($model = config('auth.providers.users.model'))) {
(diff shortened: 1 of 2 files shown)


yokesharun:
@gauravmak 
Now i Can able to change the auth model for passport using this commit  [b582109](https://github.com/laravel/passport/commit/b582109666583b514c79f43eb84489ed0051b040) but how can i get the oauth token for those diff models?

gauravmak:
I don't think Passport allows the selection of guard currently.

joshmbg:
This doesn't help - you've just changed it from reading one hardcoded config section to a different hardcoded config section - it's still hardcoded!

How is this an improvement?

andrewmclagan:
Agreed this merged PR does little to nothing to address the underlying issue of Multi-Auth.

ErwanTouba:
First of all, thanks for this nice Oauth2 implementation.
I too would love to see a 'multi-auth' feature in laravel-passport.

PR #218, opened 2016-12-23
Added token getter for guard

mbm-rafal:
It's prepared for any other token provider


taylorotwell:
Will just leave this as-is since it is not broken.

mbm-rafal:
I'm using laravel as service providing API without bootstraping it or rendering. I'm using this service with oAuth which provides me JWT tokens, so I can send it back to service from frontend. 
My channel is request not a cookie.

PR #220, opened 2016-12-28
Allow X-XSRF-TOKEN

joecampo:
Allow the XSRF token cookie to validate the CSRF token as well. I noticed that the switch to axios in L5.4 no longer requires you to pass along the X-CSRF-TOKEN header like vue-resource did. This will allow passport to work out of the box w/ the CreateFreshApiToken middleware w/ L5.4

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -200,8 +200,16 @@ class TokenGuard
      */
     protected function validCsrf($token, $request)
     {
-        return isset($token['csrf']) && hash_equals(
-            $token['csrf'], (string) $request->header('X-CSRF-TOKEN')
-        );
+        $csrfToken = $request->header('X-CSRF-TOKEN');
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
axios does pass along the header like vue-resource, it just does it automatically like Angular does.

taylorotwell:
This opens up a security hole for cross-site request forgery since the X-XSRF-Token cookie is will automatically be sent to the backend even for requests made from other websites. That is the point in forcing the X-CSRF-Token is because cross-domain cookie policies prevent malicious sites from reading the CSRF token and inserting it.

vxzhong:
Sorry. But I still cannot figure out the reason. The XSRF-Token cookie also cannot be read and inserting by the malicious sites because of the [Same-Origin Policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy). And only XSRF-Token will be sent to the backend automatically, no the X-XSRF-Token.

Issue #217, opened 2016-12-23
auth/token route, ThrottlesRequests middleware, decay and max attempts params


erdemkeren:
In `Laravel\Passport\RouteRegistrar` class, inside `forAccessTokens`method, the registered middleware `throttle` is not configurable.

see:

```php
$this->router->post('/oauth/token', [
    'uses' => 'AccessTokenController@issueToken',
    'middleware' => 'throttle'
]);
```

I think we need some way to edit the `maxAttempts` and `decayMinutes` without extending the actual middleware class.

craigpaul:
@erdemkeren The quickest way for you to solve this is to overwrite the route and still point to the right controller. In your `RouteServiceProvider`'s `map` method, you could do something like the following.

```php
public function map()
{
    $this->mapApiRoutes();

    $this->mapWebRoutes();

    Route::post('/oauth/token', [
        'uses' => '\Laravel\Passport\Http\Controllers\AccessTokenController@issueToken',
        'middleware' => 'throttle:60,5',
    ]);
}
```

I will gladly submit a PR for this after the holidays though if no one else does and it isn't deemed unnecessary by @taylorotwell or @themsaid. Hopefully this solves your problem in the meantime though :)

craigpaul:
@erdemkeren Taylor decided he'd like to keep the defaults, so go ahead and stick with the solution I gave above. You can also close this issue when you get a chance. Thanks!

erdemkeren:
@craigpaul Thanks for your interest. Already guessed that and done that way.

akem3g:
Question, is there a way to catch the error inside `RouteServiceProvider`'s  `map` method?

PR #223, opened 2016-12-28
Adds configurable options for throttle middleware

craigpaul:
References #217

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -82,6 +82,20 @@ class Passport
     public static $runsMigrations = true;
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
I really feel like the defaults are fine here.

mbm-rafal:
I feel like this could be fine regarding to using laravel as a part of NIST (https://www.nist.gov/) and FedRAMP (https://www.fedramp.gov/) certifications where this should be defined as low as possible.

craigpaul:
Sounds good to me, I left another solution in the issue anyway :) thanks!

PR #238, opened 2017-01-13
Allow for easy configuration of oauth route prefix

inxilpro:
The `/oauth/` prefix is currently hardcoded into each of the route registrar's methods. This PR moves the prefix into the default configuration for the entire passport route group. Now you can easily do:

```php
Passport::routes(null, ['prefix' => 'oauth2/v1']);
```

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -107,6 +107,7 @@ class Passport
 
         $options = array_merge($options, [
             'namespace' => '\Laravel\Passport\Http\Controllers',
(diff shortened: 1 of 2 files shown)


inxilpro:
Oops, this isn't going to work until this get's swapped around:

```php
$options = array_merge($options, [
  'namespace' => '\Laravel\Passport\Http\Controllers',
  'prefix' => 'oauth',
]);
```

It really seems like `$options` should run second here.

PR #239, opened 2017-01-15
Update phpunit.xml.dist

vinkla:
Added extra configuration and the package name.


taylorotwell:
This is causing tests to fail.

vinkla:
Oh, it seems we have a couple of risky tests without assertions.

PR #258, opened 2017-01-30
Removed use of Facades where applicable

4levels:
This allows the Passport project to be used in Lumen without Facades


taylorotwell:
Please send to 2.0 branch.

PR #259, opened 2017-01-30
Replaced Facade calls where needed

4levels:
Replaced Facade calls needed to get this to work in Lumen without Facades
Please verify if other Facade calls can / should be removed as I only checked / tested the calls used by Lumen.

### DIFF
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -226,7 +226,7 @@ class PassportServiceProvider extends ServiceProvider
      */
     protected function registerGuard()
     {
-        Auth::extend('passport', function ($app, $name, array $config) {
+        $this->app['auth']->extend('passport', function ($app, $name, array $config) {
             return tap($this->makeGuard($config), function ($guard) {
                 $this->app->refresh('request', $guard, 'setRequest');
             });
@@ -244,7 +244,7 @@ class PassportServiceProvider extends ServiceProvider
         return new RequestGuard(function ($request) use ($config) {
             return (new TokenGuard(
                 $this->app->make(ResourceServer::class),
-                Auth::createUserProvider($config['provider']),
+                $this->app['auth']->createUserProvider($config['provider']),
                 new TokenRepository,
                 $this->app->make(ClientRepository::class),
                 $this->app->make('encrypter')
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't really like modifying component code to satisfy Lumen like that. Facades should just be enabled when using Passport with Lumen.

4levels:
Hi Taylor,

I asked on IRC and all seem to agree that Facades are not necessary (and even unwanted).  Like I said, I don't want Facades: I like to require and use only the things I need. My API relies heavily on Queues and Jobs where I don't need a single Facade..
The whole Facade requirement implied by these lines in the PassportServiceProvider is really not nice and it seemed that these lines are remnants from the very first days when Passport was created.

Please accept this PR, or I will have to resort on forks and what more. Your argument "I don't really like" doesn't cut it for me!

Thanks

4levels:
Really, your statement "Facades should just be enabled" makes no sense, I don't get it why you act like this

4levels:
@SebastianS90, @RyanTheAllmighty @themsaid - just adding you guys since you worked on the PassportServiceProvider as well..

skyrpex:
I agree that facades should only be used in user land (never in library code).

SebastianS90:
@skyrpex ack!

@4levels you should also remove the other facades and eliminate their `use` statements.
Just grep for ["use Illuminate\Support\Facades"](https://github.com/laravel/passport/search?utf8=%E2%9C%93&q=%22use+Illuminate%5CSupport%5CFacades%22&type=Code)

4levels:
Hi guys,

thanks for your support!
@SebastianS90 I'm not in a position to remove the other Facade calls as I currenly don't use Laravel - only Lumen.  This means that I cannot easily verify if my changes don't break anything.

Feel free to continue my efforts here and remove the remaining Facade calls for Cookie and Request..

Kind regards!

4levels:
Hi @taylorotwell @SebastianS90,

I'm sure changing the service provider to remove all Facade calls is a 5 minute job for you.  As I explained before, I can not easily test if removing the oher facades (namely Cookie and Request) won't break anything in a working Laravel project.

I'll create a new PR anytime soon if no-one steps up to the task.

Kind regards,

Erik

PR #263, opened 2017-02-01
Determine if the client is a personal token client

JuanDMeGon:
There could be useful to know if a particular client is for personal tokens.

```
if ($user->token()->client->personal()) {
    //do something ONLY for personal user actions
}
```

:)


JuanDMeGon:
My bad, this can be done easily using direct access to the attribute.

It is not required:
if ($user->token()->client->personal_access_client) {
    //do something ONLY for personal user actions
}

Issue #273, opened 2017-02-10
[Missing] Bypass auth screen if we already have a token


raftalks:
On version 1.x there was this handy PR (by @themsaid ) that was merged.

That PR skips approval screen if the user has a valid token for the same client that has the same scope as the scope requested, Passport does auto-approve the request and return a new token.

https://github.com/laravel/passport/pull/199

This is no longer implemented on version 2.x as likely this might have been missed due to early merge.

You can see on 2.0 https://github.com/laravel/passport/blob/2.0/src/Http/Controllers/AuthorizationController.php#L51
there are no checks done.

PR #274, opened 2017-02-10
Add missing feature to bypass auth screen if we already have a token #273

raftalks:
PR for the issue https://github.com/laravel/passport/issues/273

I have copied the missing changes from version 1.x to 2.x


raftalks:
will get back to this after fixing the tests

PR #282, opened 2017-02-12
Adding a command to list the current clients

JuanDMeGon:
The developer can easily check the list of the current clients.

Can be filtered by password type or personal type.

Use `php artisan passport:clients` to list ALL the clients

Use `php artisan passport:clients --password` to list ONLY the password clients

Use `php artisan passport:clients --personal` to list ONLY the personal

### DIFF
--- a/src/Console/ClientsCommand.php
+++ b/src/Console/ClientsCommand.php
@@ -1,0 +1,107 @@
+<?php
+
+namespace Laravel\Passport\Console;
+
+use DateTime;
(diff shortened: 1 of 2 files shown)


taylorotwell:
These can be easily viewed in a database client.

JuanDMeGon:
@taylorotwell It is true but is easier in this way.

The queue:failed jobs can be viewed from a database client as well but the command makes it easy for the developer.

Does not the queue:failed command do a similar stuff?

PR #284, opened 2017-02-13
The user bridge should not assume the column name for the password column

xel1045:
We should use the getAuthPassword() method defined in the AuthenticatableContract instead.


taylorotwell:
Wrong branch. Send to 2.0.

PR #292, opened 2017-02-23
Option to use with phpseclib ^3.0

vkosachev:
If you are using Tymon JWT and want to support both ways of autentication there is no way to install both on Laravel 5.3
deps:
tymon/jwt-auth: 0.5.*


taylorotwell:
phpseclib has no stable 3 release.

PR #298, opened 2017-02-27
Added support for multi auth

eh-steve:
Previously, using multiple user providers with Passport would work in a completely insecure way (due to user_id collisions in the access tokens).

These changes allow for password grants to specify a provider (defaulting to "users"), instead of hard-coding "users" within Passport. Other grants haven't been updated to do the same yet (need to update the new tokens with the provider name after the AuthorizationServer issues them).

TokenGuard authentication now checks that the token belongs to the actual user (allowing for multiple providers), rather than just matching the user_id and resulting in potential unintentional access when using multiple user providers/Passport guards.

Since multiple user providers are not part of the OAuth2 spec, and since all the token issuing and response preparation is done by league/oauth2-server, the adjustments here are quite clunky (reinjecting an amended UserRepository, parsing AuthorizationServer grant responses and updating the corresponding token) to avoid having to change or affect the oauth2-server.

### DIFF
--- a/database/migrations/2017_02_01_000006_add_provider_to_access_tokens_table.php
+++ b/database/migrations/2017_02_01_000006_add_provider_to_access_tokens_table.php
@@ -1,0 +1,35 @@
+<?php
+
+use \Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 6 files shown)


taylorotwell:
No plans to add this for the time being.

PR #299, opened 2017-02-28
Use config file for passport

taitai42:
Hi, i came accross a situation where i needed to use different guards for laravel passport .

it happens that the guard in passport is hardcoded, so i thought it would be nice to use a config for this.

furthermore, this is the first step to put more stuff in config file for this awesome package, like table names for exemple ? 

tell me what you think about it.


taylorotwell:
No plans to add this right now.

PR #301, opened 2017-03-02
Accept * scope as valid when verifying client_credentials

chris-moreton:
When using the middleware 'client_credentials:some_scope', requests are rejected if the
token was requested using '*' scope during a client credentials grant. They are accepted
if the token was explicitly requested using 'some_scope', in this case.

If I understand correctly that the '*' scope, if granted, would allow all requests for
routes protected by any scope, then this pull request will allow this to happen.


taylorotwell:
@themsaid do you know of any problem with this?

themsaid:
No I don't see an issue with that

Issue #305, opened 2017-03-11
Support models extending 


deividaspetraitis:
Hi,

I'm working with project that is using other model rather default `Illuminate\Database\Eloquent\Model` and current solution to extend `hardcoded` model is very limiting. There is no way to use other models for example for MongoDB.

I created this issue in order to reference my pull request. Basic idea is to have configuration file where you can define your own models.

themsaid:
Closing for lack of activity, hope you got the help you needed :)

PR #306, opened 2017-03-11
Support models extending 

deividaspetraitis:
Please see #305


taylorotwell:
Passport doesn't support MongoDB. It only supports Eloquent.

deividaspetraitis:
Since lucadegasperi/oauth2-server-laravel is deprecated from 5.4 we are forced move to passport. However passport is limited with Eloquent. Having such config for our models would solve this issue. There's a lot requests from community for such feature.

PR #307, opened 2017-03-12
Token Issuing Should Use Api Middleware Group

JuanDMeGon:
Currently, the route for token issuing uses only the throttle middleware, but it should use all the middlewares defined in the "api" middleware group, giving more control over this to the developer. This will allow to the programmer to include additional middleware to this route quickly.

For example, CORS middleware, of course, the throttling, and other custom middleware.

As it is working currently, the developer must register this route again in the `routes/api.php` to use all the "api" middleware.


JuanDMeGon:
My bad (again). I just missed that by default the api routes use the auth:api stuff, making this impossible to use.

Closing.

PR #315, opened 2017-03-18
Secure isRefreshTokenRevoked

xluccianox:
I've noticed that the isRefreshTokenRevoked method from RefreshTokenRepository class returns true when the refresh token doesn't exists in the database:
`return $this->database->table('oauth_refresh_tokens')->where('id', $tokenId)->where('revoked', true)->exists();`
Although the supplied request must have a valid refresh_token (pass the decryption stage), a valid cliend_id and client_secret in order to get a new valid access_token and refresh_token... I think it's a security flaw.
I've changed the logic using the same that was implemented in isAccessTokenRevoked from TokenRepository class in here:
https://github.com/laravel/passport/blob/22a37c62cb9f8e32c384ea526f2dd3a6ed7e3d1e/src/TokenRepository.php#L76

### DIFF
--- a/src/Bridge/RefreshTokenRepository.php
+++ b/src/Bridge/RefreshTokenRepository.php
@@ -69,11 +69,29 @@ class RefreshTokenRepository implements RefreshTokenRepositoryInterface
     }
 
     /**
-     * {@inheritdoc}
+     * Check if the refresh token has been revoked.
+     *
+     * @param string $tokenId
+     *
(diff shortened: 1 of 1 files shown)


taylorotwell:
There is no description of what the security problem is.

xluccianox:
You can provide a non existing refresh_token and get a new valid access_token and refresh_token.
I've tested it.

PR #318, opened 2017-03-21
Update TokenRepository.php

Luceos:
Seems the phpdoc was not matching the actual code. Not sure whether type hinting `Model` here makes sense, just pointing it out.


taylorotwell:
Would rather not change method signatures on an existing release.

Luceos:
Didn't even notice that. My bad.

PR #325, opened 2017-03-29
Add user_id awareness to client credentials grant

paulchiu:
# Issue

Currently:

- Column `user_id` exists in the `oauth_access_tokens` table
- When generating a client credentials grant token using endpoint `/oauth/token` a user id is never used even if set; this is due to `userIdentifier` parameter being hard-coded to `null` in `\League\OAuth2\Server\Grant\ClientCredentialsGrant::respondToAccessTokenRequest`, line 38.

# Solution

- Added the ability to get `Passport\Bridge\User` from `Passport\Bridge\Client`
- Added ability for `Passport\Bridge\ClientRepository` to recongize when we are dealing with `Passport\Client` entities and set the bridge user identifier
- Created `Passport\Bridge\ClientCredentialsGrant` class, which checks for and utilises `Passport\Bridge\User` instances attached to a given `Passport\Bridge\Client` when calling `issueAccessToken`

### DIFF
--- a/src/Bridge/Client.php
+++ b/src/Bridge/Client.php
@@ -6,9 +6,9 @@ use League\OAuth2\Server\Entities\Traits\ClientTrait;
 use League\OAuth2\Server\Entities\Traits\EntityTrait;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 
-class Client implements ClientEntityInterface
+class Client implements ClientEntityInterface, UserAwareInterface
 {
-    use ClientTrait, EntityTrait;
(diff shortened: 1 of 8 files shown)


taylorotwell:
I think the whole point of client_credentials grants is that its *not* tied to a user.
screen shot 2017-03-30 at 9 44 39 am

paulchiu:
That's logical. However, the current implementation of `Laravel\Passport\Guards\TokenGuard` lines 87 to 120 is heavily reliant on the user. Would you prefer a PR to fix that or a bug report?

selection_005

Issue #326, opened 2017-03-29
Refactor request so passport can be more easily used with other ORMs


Evertt:
I'm a huge fan of the [Laravel Doctrine ORM](https://github.com/laravel-doctrine/orm), because I believe that the data-mapper pattern makes it much easier for bigger projects to stay well-organized.

I recently opened [an issue](https://github.com/laravel-doctrine/orm/issues/189) on their repository asking if they could make a bridge to use Laravel Passport with their ORM. After much discussion they concluded that this would be way too big to take on. They said that Laravel Passport is so tightly coupled with Eloquent that one can't make a simple bridge to make it work with Laravel Doctrine. That you'd have to completely rewrite / copy Laravel Passport to make it work with Laravel Doctrine.

So I have a request for Taylor. Would you be willing to refactor passport so it's easier to make it work with a different ORM? I would really love that and as you can see in [this issue](https://github.com/laravel-doctrine/orm/issues/189), I'm not the only one.

patrickbrouwers:
These changes: #327 should be enough to make it integratable.

taylorotwell:
It doesn't make it easier to stay well-organized.

patrickbrouwers:
I personally don't see any problem with using Eloquent next to another ORM for the OAuth stuff. If you're using Laravel it's already loaded anyway. I've tested it next to Laravel-Doctrine and it works just fine for 95%. With some small changes like proposed in #327 it's 100% compatible with Laravel-Doctrine without having to refactor the entire Passport core to be interchangable with other ORM drivers.

Evertt:
@taylorotwell did you ever happen to write or come upon a blog post that explains why a data-mapper pattern doesn't keep things better organized than an active-record pattern? I'd love to read such an article.

PR #327, opened 2017-03-30
Allow Passport to integrate with other user providers

patrickbrouwers:
Currently it's partially possible to user Passport with other user providers (e.g. Doctrine) as long as they embrace Eloquent being used for storing/retrieving the access tokens.

However to make it fully possible a couple of changes are necessary. 

Other user providers can now with the minimum effort of supplying a `getKey()` on their user model, integrate with Passport. 

No breaking changes should have been made. 

Solves: #326

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -41,6 +41,20 @@ class ClientRepository
     }
 
     /**
+     * Get the client instances for the given user ID.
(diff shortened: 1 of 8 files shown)


patrickbrouwers:
Thanks for merging @taylorotwell :)

PR #331, opened 2017-03-31
missing scope exception instead of authentication exception

halaei:
A missing scope should lead to an authorization error instead of an authentication one.


taylorotwell:
Wasn't this recently changed to an Authentication exception?

halaei:
I have no idea about the history of the code, but the correct exception is already used in `CheckScopes` and `CheckScope` middleware.

PR #348, opened 2017-04-14
Fix implicit flow athorization denied redirect

Mechazawa:
Laravel Passport will redirect incorrectly when the authorization request has been denied by a user while using the implicit flow.


taylorotwell:
Can you provide more information - a fuller explanation of the problem, what happens, and how this fixes it?

Mechazawa:
When the `response_type` equals `token` the server should put the parameters in the fragment (behind the `#`). Instead the server will currently put them in the query instead when the authorization is denied. This should also be done for other error responses.

See: https://www.oauth.com/oauth2-servers/authorization/the-authorization-response/

PR #372, opened 2017-05-15
not-found clients are revoked

halaei:
The current code does not work if someone deletes the record from DB. It is a security risk I guess.

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -174,8 +174,8 @@ class ClientRepository
      */
     public function revoked($id)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is very hard to read with the negative boolean checks.

taylorotwell:
I still have no idea what this is doing. Needs to be re-written.

halaei:
This also works, but is slightly less efficient:
```php
    public function revoked($id)
    {
        return ! $this->findActive($id);
    }
```
Will it be fine?

PR #377, opened 2017-05-17
Update TokenRepository.php

vietnguyen09:
Fix Access token has been revoked case if you use mongodb


taylorotwell:
Sounds like a problem with your MongoDB library.

Issue #387, opened 2017-05-30
[bug] Fatal error if refresh_token of a deleted access token


halaei:
If the token is deleted from the db, `RefreshTokenGrant` still assume it is valid. Then it calls `TokenRepository::revokeAccessToken()` which in turn calls `update()` on null, hence fatal error:

```php
    public function revokeAccessToken($id)
    {
        return $this->find($id)->update(['revoked' => true]);
    }
```

PR #388, opened 2017-05-30
fix revoke access token when token is deleted

halaei:
https://github.com/laravel/passport/issues/387

### DIFF
--- a/src/TokenRepository.php
+++ b/src/TokenRepository.php
@@ -86,7 +86,7 @@ class TokenRepository
      */
     public function revokeAccessToken($id)
     {
-        return $this->find($id)->update(['revoked' => true]);
+        return Token::where('id', $id)->update(['revoked' => true]);
     }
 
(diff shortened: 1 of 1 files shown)


halaei:
This fix forces passport to return a 200 response with new tokens, instead of 500. However, I think the correct status code should be 401 and no token should be generated. This also holds when the tokens are revoked.

halaei:
The fix is correct! If token is deleted but refresh token is still valid, the new token should be generated.

taylorotwell:
I don't understand how this isn't the exact same code as before?

halaei:
If find() return null, a fatal error occurs.

taylorotwell:
Thanks

arubacao:
Thanks @halaei . This used to produce a lot of errors in my application

pr388

@taylorotwell will you tag this or should I pull in `dev-master`?

PR #393, opened 2017-06-01
Adding "force" option so the keys are not overwritten

alnutile:
This just guards a developer from running the command and overwriting these files which would then cause issues with decryption


alnutile:
did not think my little message would make it through "You already have keys. Please use the force to overwrite them"

PR #415, opened 2017-07-01
Updated league/oauth2-server to latest version to fix security patches

alexbilbie:
Today I released `league/oauth2-server` which addresses a number of potential security vulnerabilities discovered by a security audit.

I've updated Passport to pull in this version and I've modified the library for the one breaking change to the constructor of `AuthorizationServer` to use the `APP_KEY` environment variable instead of requiring a public key. As a result the encryption is significantly stronger.

I've not tested this in a working Passport app implementation but I don't know of any other changes to the library that will affect Passport.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "illuminate/encryption": "~5.4",
         "illuminate/http": "~5.4",
         "illuminate/support": "~5.4",
-        "league/oauth2-server": "~5.0",
+        "league/oauth2-server": "^6.0",
         "symfony/psr-http-message-bridge": "~1.0",
         "zendframework/zend-diactoros": "~1.0",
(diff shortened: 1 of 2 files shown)


taylorotwell:
Brought this into master.

Kristories:
@alexbilbie @taylorotwell `php artisan route:list`?

Kolovious:
version 1.0 should be updated also, passport for 5.3 use  "league/oauth2-server": "~5.0", and pulls the 5.1.4 with this update.

AlexanderZon:
I have this when i use `php artisan route:list`
`Key path "file://base64:EMCjeF2m2CZkHbyxMW+KTUaII9dGQwXm6qV184pLCLY=" does
  not exist or is not readable`

PR #419, opened 2017-07-03
Set encryption key on the Authorization server.

adibhanna:
After updating composer packages (`composer update`) and pulling the latest Laravel 5.4.28. 
I started getting this message `You must set the encryption key going forward to improve the security of this library - see this page for more information https://oauth2.thephpleague.com/v5-security-improvements/` when I run my tests. 
I checked this page https://oauth2.thephpleague.com/v5-security-improvements/ and checked Passport code, and I couldn't find where you're setting the Encryption key. So I added it.


taylorotwell:
That change does not exist on the 2.0 branch of Passport.

PR #425, opened 2017-07-11
Set encryption key on the Authorization server + Models extending

deividaspetraitis:
Update for encryption key, see: https://oauth2.thephpleague.com/v5-security-improvements/
Also still have a hope to have native model extending in passport that makes it more flexible in case using custom models for different DB support.


taylorotwell:
I don't know what this is doing and it has no tests. Also, Passport 3.0 has the latest League.

PR #431, opened 2017-07-12
Add a message if user has not authorized any application

m1guelpf:
The other two components show a message if empty, this one disappears. This pull request makes the component render and display an empty message,like the others do.


taylorotwell:
I think it's fine that this is empty.

Issue #453, opened 2017-08-03
Key file permissions are not correct


cringer:
I am a Windows user and use Git Bash for my development terminal.
I am using PHP version 7.1.7
Default Browser is Chrome Canary or Chrome.

After installing Laravel Passport, when I run `php artisan route:list` I get the following error.

```
[ErrorException]
Key file "file://C:\Users\jschaffer\Code\passport\storage\oauth-private.key" permissions are no
t correct, should be 600 or 660 instead of 666
```

I understand what it is saying, but this is a windows machine and there is no concept of these types of permissions.  I tried changing the NTFS permissions on the file but it doesn't change this outputted error.

![image](https://user-images.githubusercontent.com/835448/28935890-c786cfc8-783a-11e7-891c-e3dfe6fa492b.png)

![image](https://user-images.githubusercontent.com/835448/28935949-f946dfe4-783a-11e7-8159-84bb4b65dfbc.png)

![image](https://user-images.githubusercontent.com/835448/28935991-172d551a-783b-11e7-990e-b3092e707a30.png)

valeryan:
This is because of a change to the oauth package used by passport. https://github.com/thephpleague/oauth2-server/pull/776. 

Passport will have to be updated to be able to provide the new third flag to turn off this check on dev environments. 

If I figure out where that new flag needs to be I will do a PR.

cringer:
Yeah I was just reading over that now.  Didn't realize the check was in league/oauth-2-server and not laravel passport.

valeryan:
@cringer if you want to get your system back up and running modify a PassportServiceProvider.php in vendor/laravel/passport/src/ around line 192 to look like this:

    /**
     * Make the authorization service instance.
     *
     * @return \League\OAuth2\Server\AuthorizationServer
     */
    public function makeAuthorizationServer()
    {
        $private = new CryptKey('file://'.Passport::keyPath('oauth-private.key'), null, false);
        return new AuthorizationServer(
            $this->app->make(Bridge\ClientRepository::class),
            $this->app->make(Bridge\AccessTokenRepository::class),
            $this->app->make(Bridge\ScopeRepository::class),
            $private,
            app('encrypter')->getKey()
        );
    }

    /**
     * Register the resource server.
     *
     * @return void
     */
    protected function registerResourceServer()
    {
        $public = new CryptKey('file://'.Passport::keyPath('oauth-public.key'), null, false);
        $this->app->singleton(ResourceServer::class, function () use ($public) {
[...]

PR #454, opened 2017-08-04
Send CryptKey instance to oauth2 with new permission check disabled.

valeryan:
This will fix https://github.com/laravel/passport/issues/453 that is caused by the changes to thephpleague/oauth2-server from here https://github.com/thephpleague/oauth2-server/pull/776. 

This disables the new permission check that was added in version 6.0.2 of oauth2-server.  I feel like the permission check should be controlled through configuration and not just be statically disabled but passport does not currently have any need for configuration and it would be a pain to add configuration for one setting. So I think some decisions still need to be made on how to handle this change. However, this is a quick solution that restores passport to working order.

### DIFF
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Facades\Cookie;
 use Illuminate\Support\Facades\Request;
 use Illuminate\Support\ServiceProvider;
 use Laravel\Passport\Guards\TokenGuard;
+use League\OAuth2\Server\CryptKey;
 use League\OAuth2\Server\ResourceServer;
 use League\OAuth2\Server\AuthorizationServer;
 use League\OAuth2\Server\Grant\AuthCodeGrant;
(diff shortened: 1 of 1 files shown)


prageeth:
I was just about to create a patch too 👍  Hopefully the PR gets merged and released soon.

cringer:
The permission checks introduced completely break windows development as their files will always return 0666.  Even though a PHP E_USER_WARNING is thrown, laravel's error handler is converting the user_notice from oauth2 to an exception.

Hope this is pulled in soon.  Having to modify the PassportServiceProvider from my vendor folder just to get past this so I can resume development on my app.  :tophat:

adiachenko:
I'm in favor of this. The new permission check breaks our Docker for Windows setup because chmod doesn't work on bind-mounted volumes.

alexbilbie:
We'll get this merged and a point release out ASAP

taylorotwell:
@alexbilbie so I should merge this?

valeryan:
@taylorotwell @alexbilbie not that my opinion matters, but I think the better solution might be to figure out how to let the oauth user_notice not be an exception so that we don't have to disable the security check. This is my knee jerk fix to get thing working again for windows folks and it might be an okay interim fix while you guys hash out what to really do to fix this permanently.

alexbilbie:
The main issues have been the following:

* Windows server incompatibilities with `chmod` at both a OS and PHP level
* Deployment difficulty with modern tooling like Kubernetes Secrets/Docker Secrets
* The OAuth library trying to automatically fix the incorrect permissions issue which many think is outside the scope of the library.

The OAuth library now no longer tries to automatically fix the issue - only warn the developer that something isn't correct.

If Laravel converts `E_USER_NOTICE` to exceptions then that is an issue for Passport/Laravel to solve not the parent library.

PR #459, opened 2017-08-08
optimize Passport scopeIds(), delete unnecessary call values()

afcn007:
`return static::scopes()->pluck('id')->values()->all();`
This pluck('id') call already return a collection with items [id1, id2, id3], no need to call the values() to resort the key again on the collection.


taylorotwell:
Will keep this just in case.

PR #463, opened 2017-08-10
Support Custom DB Connections

Hussam3bd:
Now you can use custom database connection for Passport by adding your
connection name in config/auth.php (guards.api.connection)


taylorotwell:
No plans to add this.

PR #470, opened 2017-08-14
Use Illuminate Response so that middlewares can always work

rscafi:
If that specific exception is thrown, a middleware applied to `web` that interacts with the response would fail, since the return type would be `Zend\Diactoros\Response` instead of a `Illuminate\Http\Response`. This is a simplistic example of a middleware that would fail:

```php
class SecurityHeader
{
    {
        $response = $next($request);

    }
}
```

I don't know if the approach taken is the best, but I think that it's important to use a standard response for it.


rscafi:
I saw other places where it happens too. I will try to tackle them all at once and then I'll open a new PR.

PR #471, opened 2017-08-15
Use Illuminate Response as a standard so that middlewares can always work

rscafi:
Basically, all responses that were directly returned by `League\OAuth2` classes would fail when used with Laravel middlewares  (`Zend\Diactoros\Response` is incompatible with `Illuminate\Http\Response`). Since passport routes are registered under `web` middlewares, any  middleware applied to `web` that interacts with the response would fail.

This is a simplistic example of a middleware that would fail:

```php
class SecurityHeader
{
    {
        $response = $next($request);

    }
}
```

To avoid monkey patches / `if`s on a bunch of middlewares, my proposal is to use `Illuminate\Http\Response` as a standard.

I don't know if the approach taken is the best, but I think that it's important to use a standard response for it. If needed, we can work on something else. (I can also change base to `master` branch if it's a breaking change)

### DIFF
--- a/src/Http/Middleware/CreateFreshApiToken.php
+++ b/src/Http/Middleware/CreateFreshApiToken.php
@@ -3,7 +3,6 @@
 namespace Laravel\Passport\Http\Middleware;
 
 use Closure;
-use Illuminate\Http\Response;
 use Laravel\Passport\Passport;
 use Laravel\Passport\ApiTokenCookieFactory;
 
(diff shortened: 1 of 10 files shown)


src/Http/Middleware/CreateFreshApiToken.php
```diff
@@ -89,8 +88,7 @@ protected function requestShouldReceiveFreshToken($request)
      */
     protected function responseShouldReceiveFreshToken($response)
     {
-        return $response instanceof Response &&
-                    ! $this->alreadyContainsToken($response);
+        return $this->alreadyContainsToken($response);
```
rscafi: this method is kinda useless now, we can drop it if wanted.

taylorotwell:
Is this a breaking change?

taylorotwell:
Sounds like this would probably have to be considered a breaking change.

PR #474, opened 2017-08-16
Use Illuminate Response as a standard so that middlewares can always work

rscafi:
**Since it's a breaking change, I'm opening it on `master` branch. It would be good to rebase `master` with `3.0` first.**

---

Basically, all responses that were directly returned by `League\OAuth2` classes would fail when used with Laravel middlewares  (`Zend\Diactoros\Response` is incompatible with `Illuminate\Http\Response`). Since passport routes are registered under `web` middlewares, any  middleware applied to `web` that interacts with the response would fail.

This is a simplistic example of a middleware that would fail:

```php
class SecurityHeader
{
    {
        $response = $next($request);

    }
}
```

To avoid monkey patches / `if`s on a bunch of middlewares, my proposal is to use `Illuminate\Http\Response` as a standard.

I don't know if the approach taken is the best, but I think that it's important to use a standard response for it. If needed, we can work on something else.

### DIFF
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -59,7 +59,9 @@ class AccessTokenController
     public function issueToken(ServerRequestInterface $request)
     {
         return $this->withErrorHandling(function () use ($request) {
-            return $this->server->respondToAccessTokenRequest($request, new Psr7Response);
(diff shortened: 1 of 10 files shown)


taylorotwell:
@themsaid can you review this?

themsaid:
ok

Issue #444, opened 2017-07-22
Specific database access for clients


aliasdoc:
I have an application which is a REST gateway for remote PostgreSQL database and I use Passport for authentication.

This app is composed by a local MySQL database for oauth, users, roles, permissions, ... and a remote PostgreSQL with data to serve.

I need to affect a tenant for each clients (personal and password clients) like this:

```
Client A: can fetch data from schema X
Client B: can fetch data from schema Y
Client C: can fetch data from schema X and Y
```

What is the best way to do this.

Thank you.

amazzoccone:
@aliasdoc could you find any solution for this?

aliasdoc:
Hi @amazzoccone, yes I use two connections for database access and specify in my models constructor what connection to use. For tenants, I just make a middleware to set search path in PostgreSQL, it's trivial but work.

driesvints:
Try to use the solution provided here: https://github.com/laravel/passport/issues/630#issuecomment-409436411

PR #487, opened 2017-08-29
Add oauth_client_id attribute to request

amazzoccone:
Actually if someone needs information of the request access token decrypted must extend `CheckClientCredentials`middleware only to add one line. I think it's a common case.
Maybe the other attributes (oauth_access_token_id, oauth_user_id, oauth_scopes) of the access_token could be nice to have it too.
References: 
 - `Laravel\Passport\Http\Middleware\CheckClientCredentials`
 - `League\OAuth2\Server\AuthorizationValidators\BearerTokenValidator`

Thanks for let users contribute and sorry for any english grammar mistake.

### DIFF
--- a/src/Http/Middleware/CheckClientCredentials.php
+++ b/src/Http/Middleware/CheckClientCredentials.php
@@ -44,6 +44,7 @@ class CheckClientCredentials
 
         try {
             $psr = $this->server->validateAuthenticatedRequest($psr);
+            $request->attributes->add(['oauth_client_id' => $psr->getAttribute('oauth_client_id')]);
         } catch (OAuthServerException $e) {
             throw new AuthenticationException;
         }


taylorotwell:
Needs better explanation. @themsaid ?

amazzoccone:
_CheckClientCredentials_ middleware parse the Bearer access token to verify it, so it would be cool to take advantage of that and saves the content of the jwt (oauth_access_token_id, oauth_client_id, oauth_user_id, oauth_scopes) in the current request . With this implement it, if someone wants information about the token, it would not have to extends _CheckClientCredentials_ only to add this attributes.

_Example:_ 
```You are working in a multi-tenant application, so you modify the default Passport migrations and add a new column (`site_id` for example). After _CheckClientCredentials_ was fired, a middleware load tenant based on client ID (you could know it by the content of the `access_token`request parameter). But is imposible to access client ID, because when the previously middleware (_CheckClientCredentials_) parse token to verify, it doesn't saves the attributes in the request, so the only option is to extends the middleware to add the attributes to the request. I think this should do it by default.
```
There was an issue asking for something very similar: https://github.com/laravel/passport/issues/444

amazzoccone:
Okey. I have been deepening in Passport package and trying to testing the authenticacion in my new endpoints. 
I think the problem is that _Client Credentials Grant_ authentication flow don't use a Guard, so after authenticates it doesn't attach the model in the current request. 
Besides that, you cannot use `actingAs` method in your tests because is not a Guard, so you have to make a request to `oauth/token` to get an `access_token` of the OAuth Client.

I make some changes adding a new provider `oauth_clients` and use `auth:api` instead of `client` middleware, and loads successfully the client. Maybe this approach makes more sense than the first one.
Thanks for taking your time!

taylorotwell:
Will wait for @themsaid to give his thoughts on this.

themsaid:
@amazzoccone  the point of the client credential grant is that there's no user attached to the token, it's used for machine-to-machine communication, for that you don't really check the content of token but use scopes to determine if the current token can do this or that, for example `access-products-{tenant_id}`, you can check for this scope to determine which tenant data you need to show.

amazzoccone:
@themsaid thanks for your comment! 
I understand that the Client Credentials grant type is used when the client is requesting access to protected resources under its control (i.e. there is no third party or a user). 
There are two problems using your solution with scopes:
1- The scopes are shown in other authorization grant types for third party (they are not group by client). 
2- There is a security problem because Client could access a scope of other tenant, just estimating with other `tenant-id`.

Ex.: The app (Client) is the resource owner, so it should request an access token and then make calls to a protected API. The API must know the Client ID to load the asociate tenant and return his data asociated.

cweiske:
@themsaid: When following your suggestion, I'd have to iterate over 30.000 scopes "access-products-{tenant_id}" to find out which tenant ID the client is associated with. This is not really feasible.

gerardnll:
What if different machines call to another to run a determined request and you need to know which machine made it?

Imagine you have an instance of your app for 3 different customers. Each instance needs to request giftcodes to one central giftcode provider. Each customer instance has a oauth_client in the giftcodes application. So everytime a customer requests a giftcode you can know who did the request and save it to DB for accounting purposes.

derekrprice:
So, @themsaid, your position is that every client credential grant represents a root user and the root user may restrict access to a subset of their privileges via scopes?  My use case is very similar to @gerardnll's.  It is a multi-tenant system and client credential grants are not associated to a specific user but rather to the organization that they belong to.  Creating a "virtual user" for organizations just so that I can restrict API permissions doesn't feel right.

PR #493, opened 2017-09-03
Make it possible to throw custom exceptions

mhamlet:
It is not possible to throw a custom exception inside of custom Grant Types. The Laravel Passport shows exception message instead of rendering it.

This will fix it.


taylorotwell:
Tests fail.

PR #500, opened 2017-09-07
Change the checkbox label to description instead id

yakidahan:
Change the scopes checkbox label to description instead id in create access token modal.


taylorotwell:
You can export this and change it if you want.

PR #504, opened 2017-09-08
Get proper errors field from JSON response

krzysztofengineer:
As of Laravel 5.5 errors are wrapped within "errors" field in json response. This causes wrong values being shown in passport components.


themsaid:
Thanks :)

PR #509, opened 2017-09-13
[SOLVED] null default from user_id

victorcastro:
I added two codes for resolve default "user_id" from user field UUID


taylorotwell:
Formatting is wrong and no clear description of what this fixes / allows, etc.

PR #523, opened 2017-09-21
Fixed exception when returning a download response using CreateFreshApiToken middleware

### LINKED ISSUE #489: [4.0] Call to undefined method "withCookie"
Hello,

Laravel: 5.5
Passport: 4.0

Problem:
```
(1/1) FatalThrowableErrorCall to undefined method Symfony\Component\HttpFoundation\BinaryFileResponse::withCookie()
--
in CreateFreshApiToken.php (line 51)
```

I switch from 3.0 to 4.0 and now I have the problem.

I use in my ImageController directly ```BinaryFileResponse``` (resp. Intervention/image)  but this does not have the method "withCookie".

This change is the problem: https://github.com/laravel/passport/commit/df5ddb28318a6ffac2cb0c96fc30e045afa2f9b9#diff-21f5bd46f0aa959bdc94060671ea5b18L92

I think we should add ```$response instanceof Response``` again.

doctorallen:
Fixes #489 where returning a BinaryFileResponse would cause an exception to be thrown.

### DIFF
--- a/src/Http/Middleware/CreateFreshApiToken.php
+++ b/src/Http/Middleware/CreateFreshApiToken.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport\Http\Middleware;
 
 use Closure;
+use Illuminate\Http\Response;
 use Laravel\Passport\Passport;
 use Laravel\Passport\ApiTokenCookieFactory;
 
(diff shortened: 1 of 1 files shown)


iraklisg:
@doctorallen @taylorotwell the `Call to undefined method "withCookie" ` error referenced in #489 is still present in `laravel/passport@4.0.2`, despite the merged commit :cry: . Am I missing something ?

PR #386, opened 2017-05-23
Call to registerGuard method from boot method instead of register method (in service provider)

yNaxon:
Installed Passport on a clean Laravel installation, but when I tried to execute `php artisan migrate`, the following error occurred:

>   [ReflectionException]
>   Class auth does not exist

I discovered that the error occurs because of the call to `registerGuard()` method, as you can see in the log:
```
[2017-05-23 14:38:44] local.ERROR: ReflectionException: Class auth does not exist in C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Container\Container.php:721
Stack trace:
#0 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Container\Container.php(721): ReflectionClass->__construct('auth')
#1 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Container\Container.php(600): Illuminate\Container\Container->build('auth')
#2 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Container\Container.php(567): Illuminate\Container\Container->resolve('auth')
#3 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Foundation\Application.php(708): Illuminate\Container\Container->make('auth')
#4 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Container\Container.php(1153): Illuminate\Foundation\Application->make('auth')
#5 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Support\Facades\Facade.php(159): Illuminate\Container\Container->offsetGet('auth')
#6 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Support\Facades\Facade.php(128): Illuminate\Support\Facades\Facade::resolveFacadeInstance('auth')
#7 C:\Apache24\htdocs\learn\passport\vendor\laravel\framework\src\Illuminate\Support\Facades\Facade.php(215): Illuminate\Support\Facades\Facade::getFacadeRoot()
#8 C:\Apache24\htdocs\learn\passport\vendor\laravel\passport\src\PassportServiceProvider.php(233): Illuminate\Support\Facades\Facade::__callStatic('extend', Array)
#9 C:\Apache24\htdocs\learn\passport\vendor\laravel\passport\src\PassportServiceProvider.php(83): Laravel\Passport\PassportServiceProvider->registerGuard()
[...]

PR #524, opened 2017-09-22
Using the 'Container@resolving' method instead of the 'Auth' facade in the "registerGuard" method

vladimir-kzh:


taylorotwell:
No description was given so unclear why this should be merged.

vladimir-kzh:
I'm sorry, I thought it was obvious.

The [documentation](https://laravel.com/docs/5.5/providers#writing-service-providers) states: 

> The Register Method
As mentioned previously, within the register method, you should only bind things into the service container. You should never attempt to register any event listeners, routes, or any other piece of functionality within the register method. Otherwise, you may accidentally use a service that is provided by a service provider which has not loaded yet.

The problem is that when we get an `AuthManager` through the facade, it may not be fully loaded.

For example, if I want to write my own implementation for a user provider like this:
```
class AppServiceProvider extends ServiceProvider
{
    public function register()
    {
        $this->app->resolving(AuthManager::class, function (AuthManager $auth) {
            $auth->provider('eloquent', function ($app, $config) {
                return new MyEloquentUserProvider($app['hash'], $config['model']);
            });
        });
    }
}
```
Having registered it in `\App\Providers\AppServiceProvider`, my extension will never load, because `AuthManager` will be obtained from `\Laravel\Passport\PassportServiceProvider`, **before** registering `\App\Providers\AppServiceProvider`.

Of course, to solve this problem, I can turn off autodiscovery for `laravel/passport`, and manually connect `\Laravel\Passport\PassportServiceProvider` **after** `\App\Providers\AppServiceProvider`, but it's better to do without it.

I'm assuming that in issue #386 there was a similar problem

PR #538, opened 2017-10-15
use app.url config instead of localhost hardcode

faisaladlin:
"php artisan passport:install" command inserts 2 rows into the oauth_clients table.
but the redirect column was hardcoded with 'localhost'.
this change uses values from our app.url config instead


taylorotwell:
I'm not sure this is needed? Personal access token's don't have redirects.

PR #554, opened 2017-11-01
Support X-XSRF-TOKEN header.

n0f:
Fix to be able to authenticate not only with X-CSRF-TOKEN header but also with encrypted X-XSRF-TOKEN header.

As stated in https://laravel.com/docs/5.5/csrf#csrf-x-xsrf-token, X-XSRF-TOKEN is a header automatically generated by the JavaScript framework. The VerifyCsrfToken middleware on Laravel already supports this.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -200,8 +200,16 @@ class TokenGuard
      */
     protected function validCsrf($token, $request)
     {
-        return isset($token['csrf']) && hash_equals(
-            $token['csrf'], (string) $request->header('X-CSRF-TOKEN')
-        );
+        if (!isset($token['csrf'])) {
+            return false;
+        }
+
+        $headerToken = $request->header('X-CSRF-TOKEN');
+
+        if (!$headerToken && $header = $request->header('X-XSRF-TOKEN')) {
+            $headerToken = $this->encrypter->decrypt($header);
+        }
+
+        return hash_equals($token['csrf'], (string)$headerToken);
     }
 }


taylorotwell:
I believe this introduces a security risk as I've discussed several times before I believe on the issues.

GJordan904:
I am confused as to why this introduces a security risk. I may be way off, but by looking at it I assumed using the encrypted cookie, that's automatically set by Laravel and then automatically picked up by my front-end, would be more secure than exposing the csrf token in a meta tag or setting it to a global variable in a script tag.

The biggest state of my confusion lies in the [docs for x-xsrf](https://laravel.com/docs/5.2/routing#csrf-x-csrf-token) where it states: 
> Laravel stores the current CSRF token in a XSRF-TOKEN cookie that is included with each response generated by the framework. You can use the cookie value to set the X-XSRF-TOKEN request header.

>This cookie is primarily sent as a convenience since some JavaScript frameworks and libraries, like Angular and Axios, automatically place its value in the X-XSRF-TOKEN header.

So if I can use the cookie to set the X-XSRF-TOKEN request header, then why not check for that header as well? By adding a couple lines to the TokenGaurd I am able to realize the convenience referred to in the docs. I don't have to add a script tag with the csrf to my index page and I don't have to write the front-end code to handle  intercepting requests and retrieving and adding the token as my front end conveniently does that out of the box and Laravel conveniently provides the cookie out of the box.

Forgive me if I am missing something and please point me in the right direction to better understand if I am but checking for the cookie seems like the right way to handle this.

PR #566, opened 2017-11-09
Hide some error messages on production

themsaid:


m-bymike:
This will at least hide Exceptions from pop up in production. Though it will return text, even for an accept header with `application/json`.

themsaid:
@m-bymike yeah it was always a plain text response, don't think we can change that in a minor release.

m-bymike:
@themsaid agree, thanks for fixing!

Issue #398, opened 2017-06-05
Auth events not fired


renanwilliam:
I'm developing a multi-tenant application that requires to catch authenticated requests to determine the context according with the logged user and to archive it I'm created a shared object across the application. We determined to use Passport with Password grant type as authentication strategy for SPA.

The problem is when a API request is made with passport the [Auth events ](https://laravel.com/docs/5.4/authentication#events) are not triggered. Looking the SessionGuard, it is done [here](https://github.com/laravel/framework/blob/5.4/src/Illuminate/Auth/SessionGuard.php#L134): 
```php
if (! is_null($id)) {
   if ($user = $this->provider->retrieveById($id)) {
      $this->fireAuthenticatedEvent($user);
   }
}      
```

But looking in [TokenGuard of Passport](https://github.com/laravel/passport/blob/c87deee105c3e4d7d71d578019aca00db1f1133a/src/Guards/TokenGuard.php#L102), these events are not triggered anytime.

I think these events are basic for any application and only `Laravel\Passport\Events\AccessTokenCreated` and `Laravel\Passport\Events\RefreshTokenCreated` are not enough.

CupOfTea696:
Just ran into this issue as well. I use those events to prepare authenticated API calls to an external API with tokens I have stored in my database...

jbbr:
I have the same issue. I worked around it by replacing the `auth` middleware `Illuminate\Auth\Middleware\Authenticate` with a slightly modified version:

```php
// App/Http/Middleware/Authenticate.php
...
class Authenticate extends AuthenticateBase
{
    protected function authenticate(array $guards)
    {
        if (empty($guards)) {
            return $this->auth->authenticate();
        }

        foreach ($guards as $guard) {
            if ($this->auth->guard($guard)->check()) {
                $res = $this->auth->shouldUse($guard);
                
                // fire auth event for passport/api guard
                if ($guard === 'api') {
                    event(new Authenticated($this->auth->guard($guard)->user()));
                }
                return $res;
            }
        }

        throw new AuthenticationException('Unauthenticated.', $guards);
    }
}
```

The `auth` middleware alias has to be changed to this class.

```php
// App/Http/Kernel.php
...
    protected $routeMiddleware = [
        // 'auth' => \Illuminate\Auth\Middleware\Authenticate::class,
        'auth' => \App\Http\Middleware\Authenticate::class,
...
```

This matches the guard by the alias defined in `config/auth.php`. This is usually `api` for the passport driver.

[...]

PR #568, opened 2017-11-15
Add authenticated event for bearer tokens

pdata:
I've added in a new Authenticated event to Passport which gets fired whenever a Bearer token is successfully authenticated. There's an open issue about this at #398 where some users would like to be able to set contexts or undertake other actions.

### DIFF
--- a/src/Events/Authenticated.php
+++ b/src/Events/Authenticated.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Laravel\Passport\Events;
+
+class Authenticated
+{
+    /**
+     * The newly created token ID.
+     *
+     * @var string
+     */
+    public $tokenId;
+
+    /**
+     * The user associated with the token.
+     *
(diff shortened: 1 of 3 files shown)


taylorotwell:
No plans to add this to Passport right now.

jbbr:
@taylorotwell could you explain the reasons behind this decision?
Do you generally prefer to not have an authenticated event fired in the passport token guard or are there just problems with the implementation in this PR?
Thanks!

cansozeri:
@taylorotwell I think this should be implemented. I hope you consider to add this :)

bastienu:
So, where are w on this ? Still no explanation provided ?

danhkhh:
I agree. This enhancement would be really helpful.

bastienuh:
So more than two years later, what is your position @taylorotwell ?
You never answered our questions :-(

PR #569, opened 2017-11-17
Change index name, prevent for duplicate name

overbid:
Some database can't handle duplicate index name.


taylorotwell:
You can modify the migrations in your own application if you want.

Issue #592, opened 2018-01-04
Frontend Quickstart: passport-components - vue dependencies


LaughingCabbage:
The files produced from running the command `php artisan vendor:publish --tag=passport-components` seem to require additional JS imports not mentioned in the documentation. The dependencies I've identified so far: 

[axios](https://github.com/laravel/passport/blame/4.0/resources/assets/js/components/Clients.vue#L274) @themsaid 
[underscore ](https://github.com/laravel/passport/blame/4.0/resources/assets/js/components/Clients.vue#L336)

What is the proper way to handle these imports?

Should I just add to the top of each script?
```
var axios = require('axios')
var _ = require('underscore')
```

LaughingCabbage:
In this instance I was not properly importing the bootstrap.js file in my app.js file. My code read:

app.js
`require('bootstrap')`

the fix was
app.js
`require('./bootstrap')`
which is the Laravel default.

PR #593, opened 2018-01-04
Client.vue: frontend-quickstart vue import fix

LaughingCabbage:
Axios and underscore libraries referenced but not included in script. I'm not sure if this is the proper way to handle these imports or if it was initially done by the authors at a higher level. However, the vanilla code requires these imports to suppress errors. #592

### DIFF
--- a/resources/assets/js/components/Clients.vue
+++ b/resources/assets/js/components/Clients.vue
@@ -215,6 +215,8 @@
 </template>
 
 <script>
+    var axios = require('axios');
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think we assume the user will not load them before Passport is used, since they are loaded in the default Laravel framework JavaScript bootstrapping.

LaughingCabbage:
You're right. I wasn't importing the bootstrap.js file properly in my app.js so the libraries weren't available to use.

PR #594, opened 2018-01-04
Remove decrypt from TokenGuard.

cdstarling:
After following the instructions for "Consuming Your API with JavaScript" the api requests return 401 unauthorized. This was because the JWT cookie being added to the site was not encrypted using laravels encryption service, however when it was being processed in the TokenGuard it was decrypted using laravels encryption service which threw an error.
This could either be fixed by encrypting the cookie before it is sent or removing the decryption code when it is received. I chose the latter as the token is encoded in the php-jwt package.
I apologise if I have missed something and this is the correct behaviour, or if I did the fix the wrong way round.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -186,7 +186,7 @@ class TokenGuard
     protected function decodeJwtTokenCookie($request)
     {
         return (array) JWT::decode(
-            $this->encrypter->decrypt($request->cookie(Passport::cookie())),
+            $request->cookie(Passport::cookie()),
             $this->encrypter->getKey(), ['HS256']
         );
     }
--- a/tests/TokenGuardTest.php
+++ b/tests/TokenGuardTest.php
@@ -98,10 +98,10 @@ class TokenGuardTest extends PHPUnit_Framework_TestCase
         $request = Request::create('/');
         $request->headers->set('X-CSRF-TOKEN', 'token');
         $request->cookies->set('laravel_token',
-            $encrypter->encrypt(JWT::encode([
+            JWT::encode([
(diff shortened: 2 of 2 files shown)


taylorotwell:
Most likely something is being misused on your side. This would mean Passport has been entirely broken since its release, which I know is not the case.

cdstarling:
That's what I thought.
But I've looked through the code and something doesn't add up.

The `ApiTokenCookieFactory` creates a cookie like this:
```
return new Cookie(
    Passport::cookie(),
    $this->createToken($userId, $csrfToken, $expiration),
    $expiration,
    $config['path'],
    $config['domain'],
    $config['secure'],
    true
);
```

And the `createToken` function looks like this:

```
return JWT::encode([
    'sub' => $userId,
    'csrf' => $csrfToken,
    'expiry' => $expiration->getTimestamp(),
], $this->encrypter->getKey());
```

The cookie is being decrypted and then decoded in the `TokenGuard`.

```
protected function decodeJwtTokenCookie($request)
{
    return (array) JWT::decode(
        $this->encrypter->decrypt($request->cookie(Passport::cookie())),
        $this->encrypter->getKey(), ['HS256']
    );
}
```

The encrypter is decrypting the key before letting `JWT` decode it, but nowhere can I see that it is being encrypted _after_ `JWT` has encoded it.

It works if I either remove the call to `decrypt` in the `TokenGuard` or add `encrypt` to the `createToken` function in the `ApiTokenCookieFactory`.

cdstarling:
I'm a real dummy.

Sorry to have wasted your time.

I put the `CreateFreshApiToken` middleware before `EncryptCookies`.
I thought there must be something.

PR #607, opened 2018-01-14
Fix void returns

carusogabriel:


taylorotwell:
Rather not make these changes if they aren't fixing bugs.

PR #610, opened 2018-01-15
Get user_id for new token from model

### LINKED ISSUE #609: AccessTokenRepository->persistNewAccessToken fails when user identifier is other than id
https://github.com/laravel/passport/blob/v4.0.3/src/Bridge/AccessTokenRepository.php#L58

When creating a new access token, the `user_id` is set as `$accessTokenEntity->getUserIdentifier()`.

However, `getUserIdentifier()` could be set to something other than the primary key for the User Model, if `User->getAuthIdentifierName` has been set, it could be something like the user's email, or another unique value.

A better way might be to replicate the functionality in `UserRepository->getUserEntityByUserCredentials`.

goodevilgenius:
Fixes #609

### DIFF
--- a/src/Bridge/AccessTokenRepository.php
+++ b/src/Bridge/AccessTokenRepository.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport\Bridge;
 
 use DateTime;
+use RuntimeException;
 use Laravel\Passport\TokenRepository;
 use Illuminate\Contracts\Events\Dispatcher;
 use Laravel\Passport\Events\AccessTokenCreated;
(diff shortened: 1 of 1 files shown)


taylorotwell:
All PRs need a full description in the PR and this also does not follow the formatting conventions of the framework for variables.

goodevilgenius:
I couldn't find any contributing guidelines in this repo, so I just looked at former merged PRs. Many of them had little to no description, so I assumed it was ok just to reference the issue, which has full details there.

Also, could you point me to where the formatting conventions are? I'm not sure what was wrong with my variable names.

PR #625, opened 2018-01-31
Added caching support for client and token repositories

pjnorberg:
Just wrapped the repository functions with standard Laravel Cache facades, so if using caching in env it will cache tokens and clients for 1 minute. Invalidates on delete and update methods too.


taylorotwell:
No plans to add this right now.

PR #628, opened 2018-02-04
Fix oauth_clients redirect to be stored as a native array

### LINKED ISSUE #627: Laravel\Passport\Client::redirect should be a native array
Issue #621 addressed an issue where the incorrect redirect URL was used if an authorization request was denied, but it brought to light that the underlying issue is that the redirect property of the Client model is not being stored as a native array, and instead the Bridge client explodes that field on commas to create an array. This means that without encoding the URL first, it is impossible to properly store redirect URLs that may contain a comma.

sleavitt:
Fixes #627 - Store oauth_clients redirect as a native array in the Client model rather than exploding on commas to create an array.

### DIFF
--- a/database/migrations/2016_06_01_000006_fix_oauth_clients_table_redirect_column.php
+++ b/database/migrations/2016_06_01_000006_fix_oauth_clients_table_redirect_column.php
@@ -1,0 +1,51 @@
+<?php
+
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class FixOauthClientsTableRedirectColumn extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        if (Schema::hasTable('oauth_clients')) {
+            $clients = DB::select('SELECT id, redirect FROM oauth_clients');
+
+            foreach($clients as $client) {
+                if (is_null(json_decode($client->redirect, true))) {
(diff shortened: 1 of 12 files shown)


taylorotwell:
There isn't nearly enough explanation about the breaking impacts of this PR for me to merge this.

sleavitt:
What level of explanation would you need in order to feel comfortable merging this?

The Bridge/Client already explodes on commas to create an array of redirect URLs, which is supported by League\OAuth2\Server\Entities\ClientEntityInterface::getRedirectUri() (https://github.com/thephpleague/oauth2-server/blob/8bbd21885635e132aaeca900e8221d8ca87cb6b9/src/Entities/ClientEntityInterface.php#L28)

This simply moves the support for it being an array into the Client model as a native array where it belongs, and removes the need to explode the URLs on commas, which allows redirect URLs that may happen to contain commas in them (for whatever reason) to be stored in the database that will allow them to work.

The only breaking impacts I encountered were with regard to the way the redirect URLs are stored currently, are those handled in the database migration which converts any redirect URLs that have not been stored in a JSON array format into that format so that they will work with this modification. So yes, a database migration is necessary in order to make this work without it being a breaking change, and that migration is part of this pull request.

All affected tests have been adjusted to pass in an array containing a string rather than just a string where necessary, and all tests pass.

The Clients.vue file has been updated to allow storing multiple redirect URLs, and the Laravel validation in the ClientController has been updated to take this into account as well.

I'll happily provide whatever I can in the form of documentation if that's what is lacking, but I tried to make this PR as seamless as possible otherwise. Please advise what if there is anything I can do to make this PR be considered for merging.

Issue #632, opened 2018-02-11
Infinite loop, stack overflow


iget-master:
Hi,

I have two projects, both using L5.5 and latest Laravel Passport.

Both projects runs under a docker environment, but the first one works, the last don't. 
Actually, both worked, but after some permission changes (my docker env was setting 777 for everything), it stopped working, but only one of them.

I've tried to revert the changes, clear install the environment, with no success.

The failing project just gives me 500 error, with exceeded execution time (30s).

Looking the logs, I've found the [following log, see the bin](https://pastebin.com/YubNmvkg), only for the last 3 entries, but the entire file grows very fast to 1GB.

Observe that this part repeats a lot of times, indicating an infinite loop:

> #16 [internal function]: Laravel\\Passport\\PassportServiceProvider->Laravel\\Passport\\{closure}(Object(Illuminate\\Http\\Request), NULL)
> #17 /var/www/vendor/laravel/framework/src/Illuminate/Auth/RequestGuard.php(58): call_user_func(Object(Closure), Object(Illuminate\\Http\\Request), NULL)
> #18 /var/www/vendor/laravel/framework/src/Illuminate/Auth/GuardHelpers.php(70): Illuminate\\Auth\\RequestGuard->user()
> #19 /var/www/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php(292): Illuminate\\Auth\\RequestGuard->id()
> #20 /var/www/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(221): Illuminate\\Auth\\AuthManager->__call('id', Array)
> #21 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Exceptions/Handler.php(152): Illuminate\\Support\\Facades\\Facade::__callStatic('id', Array)
> #22 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Exceptions/Handler.php(113): Illuminate\\Foundation\\Exceptions\\Handler->context()
> #23 /var/www/app/Exceptions/Handler.php(43): Illuminate\\Foundation\\Exceptions\\Handler->report(Object(League\\OAuth2\\Server\\Exception\\OAuthServerException))
> #24 /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php(143): App\\Exceptions\\Handler->report(Object(League\\OAuth2\\Server\\Exception\\OAuthServerException))
[...]

PR #634, opened 2018-02-12
Fixes infinite recursion on OAuthServerException

iget-esoares:
Hi there,

## The issue

As explained on issue #632, when an `OAuthServerException` occurs, an *infinite recursion* is triggered, causing the script to fail with PHP's `maximum execution time exceeded` error.

## The reason

The reason is that when failed, Laravel Passport [triggers the `ExceptionHandler->report()`](https://github.com/laravel/passport/blob/a04b9193054a8a7c50c26ff4f0519d3f7b5f5a30/src/Guards/TokenGuard.php#L140-L142), that will try to report the error giving the context, [that by the way calls the `Auth::user()` method](https://github.com/laravel/framework/blob/ac2c64ec61192771b106215296e6322a09221974/src/Illuminate/Foundation/Exceptions/Handler.php#L152) again, causing the recursion since `TokenGuard` doesn't keep track of failed authentication, causing multiple unnecessary (and expensive too) calls to the `authenticateViaBearerToken()` guard method.

## The proposed fix

My proposed fix is to [add an `failed` static](https://github.com/iget-master/passport/blob/e1b99f6c6d0ea74450f11b65d06a55b05b0e1f3b/src/Guards/TokenGuard.php#L57-L62) (must be static to work) boolean flag to the `TokenGuard`, and [raise it when the `OAuthServerException` occurs](https://github.com/iget-master/passport/blob/e1b99f6c6d0ea74450f11b65d06a55b05b0e1f3b/src/Guards/TokenGuard.php#L154). 

This allows that our `user()` method [checks if a failure occured](https://github.com/iget-master/passport/blob/e1b99f6c6d0ea74450f11b65d06a55b05b0e1f3b/src/Guards/TokenGuard.php#L95-L100) and skip the call to the `authenticateViaBearerToken()` since we already know that the authentication failed previously, and anything changed, so the result will be always the same.

I really expect that this could be fixed, since our application is broken due to this.

## Breaking impacts

This should not have breaking impacts since it just optimizes an failing process.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -55,6 +55,13 @@ class TokenGuard
     protected $encrypter;
 
     /**
+     * The failed authentication flag.
+     *
+     * @var boolean
+     */
+    static $failed;
+
+    /**
      * Create a new token guard instance.
      *
      * @param  \League\OAuth2\Server\ResourceServer  $server
@@ -85,6 +92,13 @@ class TokenGuard
      */
     public function user(Request $request)
     {
+        // If the authentication process failed, we should'nt try to
+        // get the user. This avoids an infinite recursion caused by
+        // the error handler trying to get the user to store context.
+        if (static::$failed) {
+            return;
+        }
+
         if ($request->bearerToken()) {
             return $this->authenticateViaBearerToken($request);
         } elseif ($request->cookie(Passport::cookie())) {
@@ -137,6 +151,8 @@ class TokenGuard
 
             return $token ? $user->withAccessToken($token) : null;
         } catch (OAuthServerException $e) {
+            static::$failed = true;
+
             Container::getInstance()->make(
                 ExceptionHandler::class
             )->report($e);
--- a/tests/TokenGuardTest.php
+++ b/tests/TokenGuardTest.php
@@ -64,6 +64,13 @@ class TokenGuardTest extends TestCase
         );
 
         $this->assertNull($guard->user($request));
+
+        // This assert that the second call after failed doesn't
+        // call the `validateAuthenticatedRequest` again.
+        $this->assertNull($guard->user($request));
+
+        // Avoids this test from interfering on other tests
+        TokenGuard::$failed = false;
     }
 
     public function test_null_is_returned_if_no_user_is_found()


taylorotwell:
I'm not sure why the static is necessary. The authentication driver manager caches already created drivers. In this case the Passport driver would have already been created so the same instance would be used.

iget-esoares:
@taylorotwell Yeah, I thought the same about the static, but for some reason the instances are not the same so not being static avoids the fix to work. I'll investigate the reason.

iget-esoares:
See the test:

- Add a var_dump() to constructor
- Add a var_dump() to the `catch` on the `authenticateViaBearerToken()` method with `spl_object_hash($this)`

The result will be:

> /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php:80:string 'constructed' (length=11)
> /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php:158:string '00000000522d4bab0000000056e91728' (length=32)
> /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php:80:string 'constructed' (length=11)
> /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php:158:string '00000000522d4bcb0000000056e91728' (length=32)
> /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php:80:string 'constructed' (length=11)
> /var/www/vendor/laravel/passport/src/Guards/TokenGuard.php:158:string '00000000522d4beb0000000056e91728' (length=32)
> 

This will be repeated many times. For some reason the instance is being destroyed and reconstructed (that's why the object hash is the same) instead of being reused.

taylorotwell:
Please re-submit when you have a non-static option. It should not be needed. @themsaid can you reivew?

iget-master:
@taylorotwell I'm trying to create a repository to reproduce the problem, also trying to explain the reason why static is necessary.

taylorotwell:
OK shoot me an e-mail if you can do that. taylor at laravel.

nmfzone:
@taylorotwell Hey, I've another approach to solve this issue. I'll submit a fix.

iget-master:
@taylorotwell What about if we reset the `Authorization` header on OAuth2 failure, so we break the cycle? Please see #666 (hell yeah)

PR #639, opened 2018-02-14
Added firebase 5.0 as possible dependency

duxthefux:
Hi,

I added the missing firebase 5.0 dependency which was already in 4.0.


duxthefux:
@taylorotwell would be possible that you create a new tag with `5.0.1` so composer recognizes the update?

thanks!

PR #642, opened 2018-02-20
TokenGuard did not check for x-xsrf-token header

itsUndefined:
If you are using an SPA and trying to consume your own API, it's impossible to use the blade engine to get the unencrypted CSRF token. Other middleware uses the header X-XSRF-TOKEN to get the encrypted token from the cookie.  I did the same in the token guard. The PR for laravel with code [#4128](https://github.com/laravel/laravel/pull/4128) had questioned whether or not passport should get updated to support this but no one got into it.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -161,7 +161,7 @@ class TokenGuard
         }
 
         // We will compare the CSRF token in the decoded API token against the CSRF header
-        // sent with the request. If the two don't match then this request is sent from
+        // sent with the request. If the two don't match then this request is not sent from
         // a valid source and we won't authenticate the request for further handling.
         if (! $this->validCsrf($token, $request) ||
             time() >= $token['expiry']) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
You're inserting a security vulnerability.

itsUndefined:
Oh. I think I missed something. Shouldn't this cookie be safe from getting read from another domain? The only way someone could indeed do damage would be with XSS but then that would also risk getting the unencrypted CSRF token.

itsUndefined:
Does that mean that the web routes, that have this X-XSRF-TOKEN header implemented, are vulnerable?

itsUndefined:
@taylorotwell I would like to ask why the angular js team believes that this is a safe way to implement CSRF protection as seen in their [documentation](https://docs.angularjs.org/api/ng/service/$http#cross-site-request-forgery-xsrf-protection):  and why it should be avoided and is a security vulnerability. If so should this not be used for web routes too?

PR #645, opened 2018-02-21
Call custom exceptions handling if exists for JSON APIs

igorjacauna:
Inserting a custom method call int `HandlesOAuthErrors.php@withErrorHandling` that allow to handle exceptions that not fire the custom laravel exception handler in `\App\Exceptions\Handler`

`return $this->exceptionHandler()->renderForApi($response);`

Will try call `renderForApi` method in `\App\Exceptions\Handler` if it exists. If not exists will return the `$reponse` handled by Passport 

Calling custom function allow to use localization in Exceptions messages and make some traits if wish before throw the exception.

Without this we cannot change messages or localize the messages if we use the Passport in a JSON API

### DIFF
--- a/src/Http/Controllers/HandlesOAuthErrors.php
+++ b/src/Http/Controllers/HandlesOAuthErrors.php
@@ -24,10 +24,11 @@ trait HandlesOAuthErrors
      */
     protected function withErrorHandling($callback)
     {
+        $response = $this->getResponse($callback);
         try {
             return $this->exceptionHandler()->renderForApi($response);
         } catch(Exception $e) {
-            return $this->getResponse($callback);
+            return $response;
         }
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Why did you comment out some lines from a test?

taylorotwell:
Also, I would just use method_exists instead of try / catch.

igorjacauna:
Ok. I changed to `method_exists` and uncommented the test lines.

I commented out the lines in test becase the  `get` from `$config` was not called anymore. If I did some mistake I need a little help here to understand this scenario. 

Now the `phpunit` results show ony one fail:

```
AccessTokenControllerTest::test_exceptions_are_handled
Mockery\Exception\InvalidCountException: Method get(<Any Arguments>) from Mockery_0 should be called
 exactly 1 times but called 0 times.
```

taylorotwell:
Going to hold off on this for now.

PR #659, opened 2018-03-07
Use Authenticatable::getAuthIdentifier() in place of Model::getKey()

RemiCollin:
This PR enhance interoperability by using the `Authenticatable` contract's methods when Passport require an user's identifier, thus adding the ability to use passport with custom user classes / authentication drivers.

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -64,7 +64,7 @@ trait HasApiTokens
     public function createToken($name, array $scopes = [])
     {
         return Container::getInstance()->make(PersonalAccessTokenFactory::class)->make(
-            $this->getKey(), $name, $scopes
+            $this->getAuthIdentifier(), $name, $scopes
         );
     }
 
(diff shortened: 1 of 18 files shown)


taylorotwell:
Could be a breaking change on 5.0 branch.

RemiCollin:
Is there a case where `$request->user()` won't return an `Authenticatable` instance ? 

That's the only breaking change I can think of and seems very unlikely to me.

RemiCollin:
Should I target `dev-master` instead ?

tbuteler:
@taylorotwell Do you think it's feasible to have this in 6.0? I've also bumped into this issue because we have a custom user provider in our Laravel app.

It's a potential breaking change, of course, but the likelihood of impact is low, as most Laravel implementations should have the `Authenticatable` trait in their `User` model, and by default the `getAuthIdentifier` method will return `getKey`, just like previous Passport versions.

It's worth noting that [`UserRepository::getUserEntityByUserCredentials`](https://github.com/laravel/passport/blob/6.0/src/Bridge/UserRepository.php#L58) already uses the `getAuthIdentifier` method to identify a `Bridge\User` entity, so this seems to be consistent with the existing approach.

PR #663, opened 2018-03-08
Create clients only if migrations are being used

TitasGailius:
When you have defined `Passport::ignoreMigrations();` and call `php artisan passport:install` it obviously fails to create the clients because the database table does not exist. 

This simply checks whether migrations are being run before creating initial clients with `php artisan passport::install` command.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Passport\Console;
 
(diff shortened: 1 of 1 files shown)


TitasGailius:
I completely misunderstood the purpose of `Passport::ignoreMigrations()` which is to skip the default migrations and use something else rather than not to use it at all.

Sorry for the inconvenience

Issue #674, opened 2018-03-20
Variable scopes for different token types


NastuzziSamy:
Hi everyone !

Thx for the work, but i was thinking about an issue, how can i specify an array of scopes destined for Issuing Access Tokens and another array for Client Credentials ?

I know that in the implementation, it could be managed with the different middlewares but what about application who asks the user to access to some scopes, including the one who allows it to access to "all data users" which the original user can't give.

Like i was worried not to be that clear, I put my example in that screenshot: 
![2018-03-20_01-12-47](https://user-images.githubusercontent.com/24639904/37628880-e1258816-2bdb-11e8-9b93-f063ce9271f9.png)

Thx again ! :)

driesvints:
Hi there,

Looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs or problems. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

PR #675, opened 2018-03-20
Possibility to create variable scopes for different authorization modes

NastuzziSamy:
Like I explained in this issue #674, it is now not possible to make the difference for the scopes destined to be used by application using a certain authorization mode or another.

For example, if we take theses scopes:
- get-user: allow an application to get information about the current connected user
- get-users: allow an application to get information about all users (connected or not)

It is reasonnable not to access to the scope "get-users" when issuing authorization code with user connection needed and conversely, "get-user" would only be accessible by an application with access allowed by the connected user.

Obviously, this feature is backported.

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -63,7 +63,6 @@ class AuthorizationController
             $authRequest = $this->server->validateAuthorizationRequest($psrRequest);
 
             $scopes = $this->parseScopes($authRequest);
-
             $token = $tokens->findValidToken(
                 $user = $request->user(),
                 $client = $clients->find($authRequest->getClient()->getIdentifier())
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/AuthorizationController.php
```diff
@@ -90,14 +89,16 @@ public function authorize(ServerRequestInterface $psrRequest,
...
-            collect($authRequest->getScopes())->map(function ($scope) {
-                return $scope->getIdentifier();
-            })->all()
-        );
-    }
+		return Passport::scopesFor(
+			collect($authRequest->getScopes())->map(function ($scope) {
+				return $scope->getIdentifier();
+			})->all(),
+			$authRequest->getGrantTypeId(),
```
NastuzziSamy: Specify the specific grant type

src/Passport.php
```diff
@@ -44,9 +45,16 @@ class Passport
      * @var array
      */
     public static $scopes = [
-        //
+		//
```
NastuzziSamy: Scopes variable with all scopes is needed for backport but also for Oauth2 to work

taylorotwell:
No plans to add right now.

PR #678, opened 2018-03-23
Adding optional configuration.

MatthewSH:
I use Passport for almost every project I have. It makes certain tasks fairly easy and it makes people integrating my site into theirs fairly easy. However, recently in one of my projects, I found the need to adjust a couple things in the `Token` model. There is no easy way to do this. So I propose, with this PR, that an optional configuration is added so if we need to use our own token model, we have the ability too.

I need to test a couple more things to make sure it works properly. I would like to go ahead and submit this PR to get some feedback on it and to see if this is a small enough change to where it can be easily merged into Passport.

Update: Just tested this PR on the project I wrote it for. Works as expected.

### DIFF
--- a/src/TokenRepository.php
+++ b/src/TokenRepository.php
@@ -15,7 +15,7 @@ class TokenRepository
      */
     public function create($attributes)
     {
-        return Token::create($attributes);
+        return config('passport.token.model', Laravel\Passport\Token::class)::create($attributes);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/TokenRepository.php
```diff
@@ -15,7 +15,7 @@ class TokenRepository
      */
     public function create($attributes)
     {
-        return Token::create($attributes);
+        return config('passport.token.model', Laravel\Passport\Token::class)::create($attributes);
```
taylorotwell: I would suggest making a method that returns a new token instance... like `newTokenInstance`. So you don't have to repeat this config call so many times in this file.

MatthewSH:
@taylorotwell This last commit should take care of that.  Let me know if there's anything else.

taylorotwell:
None of your namespace references are correct and this needs tests.

PR #679, opened 2018-03-23
Support customized User::getAuthIdentifier()

oanhnn:
In some cases, you want customize `User::getAuthIdentifier()` method. 
Eg. in my case, I want add prefix `u` and `s` for `User` and `Staff` model, and both two models authenticate by the guard `web`.
This pull request is support those cases.

Sorry, my English is not good.


oanhnn:
Yes. Please make it your self on next version. 
I think `$request->user()` should return Authenticatable Contract or Null .

PR #710, opened 2018-05-15
Use dropIfExists() instead of drop()

adriaanzon:
I was getting errors about this in my tests, which start with an empty database.


Sephster:
Shouldn't this have been merged into version 6 branch, not 5? @taylorotwell @adriaanzon

adriaanzon:
@Sephster Yes, it should have been merged into `6.0`. `5.0` was this repository's default branch at the time I made this pull request. This is also true for #704.

@taylorotwell should a new pull request be made to the `6.0` branch?

Issue #122, opened 2016-10-07
Support to Uuid


jimmylagp:
I have a problem with Passport and my Uuid users. Can you give me a solution for that problem?.

nikkuang:
Since Passport uses integer for user id, If you want to use uuid for your users you have to modify the script for migration. If you have the latest version of Passport, you can add this code `Passport::ignoreMigrations();` on your `AuthServiceProvider@register` then run 

`php artisan vendor:publish --tag=passport-migrations`

and just change migration scripts just change all user_id from integer to uuid and run your updated migration script

livingos:
I think the client id should also be a Uuid by default and not just an integer. 

If you are making your API available to external users, Uuid makes more sense.

heisian:
Yes I'm looking for this as well - it seems pretty standard in any OAuth scheme to use UUID or some sort of random string instead of normal integers for a client ID.

heisian:
Introduced UUID here:
https://github.com/laravel/passport/pull/168

ssuhat:
is this has been merge?. can't apply passport because of integer user_id

heisian:
@ssuhat no, not yet.  My pull request is working from the `master` branch while the current stable release is `1.0`.  So there may be more changes Taylor makes to `master` that I'll be pulling in.

Maybe this will make it to the next version.  My advice is to take my pull request and merge it into your own fork for now.

themsaid:
@ssuhat 

> can't apply passport because of integer user_id

You can alter the migration if you want.

ssuhat:
@themsaid i won't do that. it going take too much work to change my user_id to integer.

themsaid:
@ssuhat  No, I mean Passport's migrations, you can alter the key type to be string.

aliasdoc:
Hi when I try to publish migration files, it fails:
```php
php artisan vendor:publish --provider="Laravel\Passport\PassportServiceProvider" --tag=passport-migrations
Nothing to publish for tag [passport-migrations].
```

I added service provider in `config/app.php` : 
```php
/*
* Package Service Providers...
*/

 Laravel\Passport\PassportServiceProvider::class,
```

I forgot something ??
Thanks

heisian:
I get the same issue too. not sure why the migrations won't publish. will try with a completely fresh installation

jimmylagp:
[...]

PR #714, opened 2018-05-19
Made model classes configurable

robinvdvleuten:
With this PR, one can change and extend any model classes provided by Passport in a similar fashion as Spark. If people need stuff like UUIDs (#122), they can just add this functionality themselves without touching the Passport codebase. 

⚠️ Please note that this PR does not contain the complete change yet. But I would check with @taylorotwell and @themsaid first to see if they like this idea before putting more effort in it.

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -12,7 +12,7 @@ class ClientRepository
      */
     public function find($id)
     {
-        return Client::find($id);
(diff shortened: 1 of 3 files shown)


Sephster:
I think the default branch needs updating as this is going to version 5, not version 6, which I believe is the latest stable

PR #715, opened 2018-05-21
Make model classes configurable

robinvdvleuten:
With this PR, one can change and extend any model classes provided by Passport in a similar fashion as Spark. If people need stuff like UUIDs (#122), they can just add this functionality themselves without touching the Passport codebase.


robinvdvleuten:
Now it's the correct branch 👍

PR #724, opened 2018-05-31
Protect client credentials grant

willtj:
As in https://tools.ietf.org/html/rfc6749#section-4.4:
> The client credentials grant type MUST only be used by confidential clients.

This adds a check that the given client is confidential (has a secret), which allows public clients to be created without the risk that they will be used to obtain a client_credentials grant.

Also adds tests for the changed function.


willtj:
I noticed from the failing tests in #725 that I'd missed an existing test class that should be updated rather than the new one in this commit - I'll update and commit again.

PR #725, opened 2018-05-31
Restrict grant types

willtj:
Adds the ability for clients to be restricted to specific grant types. Clients without a `grant_types` array will still be allowed subject to the existing restrictions.


willtj:
Broken tests - I'll fix.

taylorotwell:
Can you re-submit this with further explanations of why this is needed and how you are using it? Sorry, I'm not following based on the short description.

PR #729, opened 2018-05-31
Restrict grant types

willtj:
Following on from #725.

Adds the ability for OAuth clients to be restricted to specific grant types. Clients without a `grant_types` array will still be allowed any grant type, subject to the existing restrictions.

For example, existing checks allow any third party client to request an `authorization_code` grant. This change would allow a client to be set up in such a way that it may be used for a `client_credentials` grant type but not `authorization_code`. A similar implementation can be seen in:
- Auth0: [docs](https://auth0.com/docs/applications/application-grant-types#how-to-edit-the-client-grant_types-property)
- The bshaffer/oauth2-server-php library: [docs](http://bshaffer.github.io/oauth2-server-php-docs/overview/grant-types/#restricting-grant-types-to-clients) and [code](https://github.com/bshaffer/oauth2-server-php/blob/8856aed1a98d6da596ae3f9b8095b5c7a1581697/src/OAuth2/Storage/MongoDB.php#L109)


taylorotwell:
Thanks. Would it make sense to add grant_types as a cast on the Client Eloquent model?

willtj:
Good point - PR coming up.

mro95:
This feature is missing documentation. Also the grant_types column is missing in the migration.

PR #736, opened 2018-06-12
Private Key Length

amrfayad:
Add private key length in .env file to control the length generated private key


taylorotwell:
I would rather just see key length be an argument. `env` helper is not intended to be called outside of configuration files.

amrfayad:
i have change it to be inside configuration files 

[https://github.com/laravel/passport/pull/737](url) .

PR #742, opened 2018-06-19
Private Key Length

amrfayad:
Add private key length as argument to the CLI command


amrfayad:
@taylorotwell    why this pull request fails ?

Issue #691, opened 2018-04-10
Client credentials are not secure for documented use case


soundsgoodsofar:
Reading the documentation (https://laravel.com/docs/5.6/passport#client-credentials-grant-tokens) makes it sound like a token issued for a client grant would be secure to use for internal machine-to-machine authentication.

However the CheckClientCredentials middleware does *not* validate that a token is actually a client credential.  It will accept a token generated for anything--password grants included.  This means that if these routes are reachable, anyone could call these internal endpoints doing who knows what.

The documentation is misleading and dangerous.  It makes it sound like these tokens can be used to secure API calls from internal services, but implementing it as documented does nothing of the sort.

I thought maybe I could use scopes to limit the permissions on a per-client basis but that doesn't work either!  Any client can request any scopes.  Client credentials don't have an associated user, so I can't do user-based permissions either.  Seems like client credentials aren't a fully baked concept.

soundsgoodsofar:
For anyone else trying to sort out how to use this, I duplicated CheckClientCredentials.php to my own class, SecureCheckClientCredentials.php.  Change the middleware registration in Kernel.php.

Then you can add something like this after the $psr try-catch block.

```
$client = Client::findOrFail($psr->getAttribute('oauth_client_id'));

if ($client->password_client || $client->personal_access_client) {
    throw new AuthorizationException;
}
```

You could just check the client ID directly if you want to restrict to a specific client, or do whatever you want.  Personally I'd like to have an attribute, Client->internal_client, the same way we have it for password and personal_access, but I'm trying to do as little surgery as possible and this works for my purposes.

robbytaylor:
@soundsgoodsofar, thanks for that workaround. Did you try opening a PR to put that change into Passport's `CheckClientCredentials`?

taylorotwell:
Hmm, maybe I'm confused but it seems to me the different grants are just different ways of obtaining a token. It is still up to you to make sure whatever given token can perform a given action? 

It feels like I would probably do this by checking the token directly in the route...

Something maybe like (from within user model):

```php
public function canDoSomething()
{
     return $this->accessToken && $this->accessToken->client->something();
}
```

robbytaylor:
> Hmm, maybe I'm confused but it seems to me the different grants are just different ways of obtaining a token. It is still up to you to make sure whatever given token can perform a given action?

[...]

PR #747, opened 2018-06-24
Added ability to restrict scopes for a client

robbytaylor:
This is a partial fix for https://github.com/laravel/passport/issues/691.

With this change the scopes which a client can issue tokens for is limited to it's own array of `scopes`. This allows scopes to be used to protect internal machine-to-machine requests from being accessed by users with valid password grant tokens.

This adds functionality which exists in other PHP OAuth server implementations. e.g. https://bshaffer.github.io/oauth2-server-php-docs/overview/scope:

> The scope(s) available to a client are controlled by a combination of the scope field in the client storage, and the list of scopes available, as defined in the scope storage.
>
> When the client has a list of scopes configured alongside it, the client is restricted to use only those scopes. When there are no scopes configured, the client is not restricted in what scopes it may use, it is able to use all scopes available within the authorization server.

This implementation behaves in the same way.

One thing I noticed is that following https://github.com/laravel/passport/pull/729 the migration to create the `oauth_clilents` table hasn't been updated to include the `grant_types` column. If this PR gets merged I'll therefore do another PR to add both the `grant_types` and `scopes` columns, if that's helpful.

### DIFF
--- a/src/Bridge/ScopeRepository.php
+++ b/src/Bridge/ScopeRepository.php
@@ -2,12 +2,20 @@
 
 namespace Laravel\Passport\Bridge;
 
+use Laravel\Passport\ClientRepository;
 use Laravel\Passport\Passport;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 use League\OAuth2\Server\Repositories\ScopeRepositoryInterface;
 
 class ScopeRepository implements ScopeRepositoryInterface
 {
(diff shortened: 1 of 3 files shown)


taylorotwell:
It feels like we are adding all kinds of code before addressing my original question and simple solution in the original PR? I would like to see that discussed first and then we can re-visit these.

PR #748, opened 2018-06-24
Update CheckClientCredentials middleware to fail if client is a first party client

robbytaylor:
This fixes the issue discussed in https://github.com/laravel/passport/issues/691 by ensuring the `CheckClientCredentials` middleware doesn't grant access by tokens granted by personal access or password clients.

This is a slightly modified version of the workaround proposed by @ soundsgoodsofar in https://github.com/laravel/passport/issues/691#issuecomment-380209374


taylorotwell:
My comment was never really addressed in the original PR.

PR #758, opened 2018-07-10
Fix key length issue on passport:install

Donovan-H:
`passport:install` is failing with:
 `openssl_pkey_new(): private key length is too short; it needs to be at least 384 bits, not 0`

Type casting the `passport:key` `--length` argument to an integer fixes this issue.

The RSA::createKey method requires an integer to be passed.
https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/RSA.php#L334

### DIFF
--- a/src/Console/KeysCommand.php
+++ b/src/Console/KeysCommand.php
@@ -32,7 +32,7 @@ class KeysCommand extends Command
      */
     public function handle(RSA $rsa)
     {
-        $keys = $rsa->createKey($this->input ? $this->option('length') : 4096);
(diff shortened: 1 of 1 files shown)


JosephNC:
Well for me, it doesn't. Only integers lower or equal to 256 works

rtuin:
Good stuff. Was actually just running into this!

drmmr763:
I had this issue and the CLI param fixed it partially; but it also manifested its self when using the OAuth Client Password Grant Type endpoint; can anyone confirm if this fix will address that or if more work is needed in that area?

PR #759, opened 2018-07-10
Fix openssl_pkey_new ErrorException

wuwx:
$this->option('length') return string type\
openssl_pkey_new function throw exception
openssl_pkey_new(): private key length is too short; it needs to be at least 384 bits, not 0


Donovan-H:
@wuwx not sure if you noticed but this is **_identical_** to the pull that's already open ;-)

PR #770, opened 2018-07-18
Update authorize.blade.php

sindicatoesp:
the application might not be in root folder


nmfzone:
@taylorotwell Is it make sense to put Laravel Foundation helpers inside Passport?

PR #781, opened 2018-07-26
authIdentifierName should not be static for password grants

duruld:
authIdentifierName column should not be static, should be same as stated in the Model.


duruld:
I've realized the purpose of findForPassport now, this is not a necessity. I'm closing this.

PR #783, opened 2018-07-31
Respect session's same_site configuration

robinvdvleuten:
This PR makes sure that the cookies created by Passport use the same configuration as the session cookies. Currently they do not respect the `same_site` config option.


taylorotwell:
Will break if that option isn't set. Use the null coalesce operator to set the default.

robinvdvleuten:
I thought that it already had a null default from the session's configuration as the others don't have a default as well. I've added the coalesce operator to make sure it always has a value 👍

PR #787, opened 2018-08-01
closes #786, timestamps parameter's value is set to true

### LINKED ISSUE #786: $timestamps = false is set as default on Token model
I have created the index page for personal access tokens in my application and tried to output the time every record was created like this `$token->created_at->diffForHumans()` since I expect that `created_at` and `updated_at` should be mutated to Carbon instances like documented. But it requires `$timestamps = true` at first so those fields will be mutated:

https://github.com/laravel/framework/blob/16983a689d15333e7101457c3e0c0d0b7da01d69/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L803-L810

`Token` model with current `$timestamps = false` value: https://github.com/laravel/passport/blob/86af0267016a076d58a780a86cbcdf0448ccc95e/src/Token.php#L54

Migration for `access_tokens` table implies that table has timestamps: https://github.com/laravel/passport/blob/3492cc0641c166b18023959329243543477f56d4/database/migrations/2016_06_01_000002_create_oauth_access_tokens_table.php#L23

It is a very quick fix and I will make a PR easily but I am aware that maybe it has a purpose by setting `$timestamps` to false.

justinasposiunas:
Detailed description is  there #786

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -51,7 +51,7 @@ class Token extends Model
      *
      * @var bool
      */
-    public $timestamps = false;
+    public $timestamps = true;
 
     /**
      * Get the client that the token belongs to.


taylorotwell:

Thanks!

PR #796, opened 2018-08-09
Adapt for laravel breaking change

themsaid:
This PR adds the new option `Passport::dontUnserializeCookies();` that you can place it in `AuthServiceProvider` to instruct passport that tokens shouldn't be unserialized.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -187,7 +187,7 @@ class TokenGuard
     protected function decodeJwtTokenCookie($request)
     {
         return (array) JWT::decode(
-            $this->encrypter->decrypt($request->cookie(Passport::cookie())),
+            $this->encrypter->decrypt($request->cookie(Passport::cookie()), Passport::$unserializesCookies),
             $this->encrypter->getKey(), ['HS256']
         );
     }
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -118,6 +118,13 @@ class Passport
     public static $runsMigrations = true;
 
     /**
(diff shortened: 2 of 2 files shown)


amacsmith:
Works for me now. Thanks!

ThomHurks:
So Laravel changed the default to not serializing, but passport still unserializes by default and you have to manually turn that off? That's going to trip some people up...

sfox55:
@ThomHurks agreed. I lost almost an entire day debugging this because I ran a 'composer update'. Shame on me I suppose....

amacsmith:
I wonder if it could have handled differently. It was done to fix a vunerability just wonder if could been an artisan command to make the edits. Then in 5.7 automate it. However I'm not sure how it should be handled. It would be interesting to know.

yaquawa:
@ThomHurks Agree! This should align with Laravel framework. Serialize or not, should be always the same default value. This is really confusing, especially people those who new to Laravel or Passport.

ThomHurks:
In the mean time a new major version was released (7.0) which does have the same default as Laravel itself, but you do need to edit your composer.json in order to get that version. So, definitely not as smooth as it could have gone but the projects are aligned again I guess...

luisdalmolin:
@themsaid this is going to be backported to the Laravel Passport version compatible with Laravel 5.5? I can submit the PR if is possible to tag a new version with this.

qiutuleng:
I hope it will be released to 4.0.4 as soon as possible.

matijakovacevic:
in the end, will this be tagged as 4.0.4 ?

PR #797, opened 2018-08-10
Backport pull request #796

clnt:
Backport pull request #796 into laravel/passport 4.0


simondavies:
Not sure if fixed yet but today just running composer update did not download the new entry 4.0

So as mentioned earlier i had to do `composer require laravel/passport:4.0.x-dev` for now to download the update.

But it did sort out the issue :-)

PR #811, opened 2018-09-05
Fix return type of `TokenGuard::user`

yaquawa:
The signature of `TokenGuard::user` should match to [Illuminate\Auth\RequestGuard::user](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Auth/RequestGuard.php#L46).


yaquawa:
@taylorotwell Yes, I see. Nothing wrong with result it self, but you can still change `@return mixed` to `@return \Illuminate\Contracts\Auth\Authenticatable|null` explicitly to denote the return type more clearly. Although I'm always expect passport move to PHP 7 so we can type return type natively.

PR #820, opened 2018-09-11
Use a Key object when instantiating AuthorizationServer for 200x-300x faster encryption of refresh token

EricTendian:
Right now, it can be quite slow to generate a new access and refresh token due to having to use the PBKFD2 hash function, which can account for over 50% of time spent generating an access token in most cases. I found an issue on the oauth2-server package which describes this issue in more detail: https://github.com/thephpleague/oauth2-server/issues/812

Since the encryption key Laravel uses already comes from the `random_bytes` function, it is secure enough to use as-is. We are also storing the key server-side, so there isn't a big need to use a password hashing key generation function when trying to encrypt the refresh token.

The PR https://github.com/thephpleague/oauth2-server/pull/814 implements a fix (released on version 7.x of the library), by allowing the authorization server to be instantiated with a Key object instead of a string. By taking advantage of this and passing in an instantiated key, I was able to get a 290x speed improvement over 1000 iterations of encrypting a refresh token (2+ mins to 0.5s). Other machines may have similar performance gains.

I have verified that Passport continues to work as normal with this change. This change can be backported to Passport 6.0 as that version also has version 7.x of the OAuth2 library. (I can make a separate PR for the backport once this one is approved)

### DIFF
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -3,6 +3,8 @@
 namespace Laravel\Passport;
 
 use DateInterval;
+use Defuse\Crypto\Key as EncryptionKey;
+use Defuse\Crypto\Encoding as EncryptionEncoding;
 use Illuminate\Auth\RequestGuard;
 use Illuminate\Auth\Events\Logout;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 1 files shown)


EricTendian:
@taylorotwell Good catch, in the current state of this PR it would be a breaking change. However, I have found a solution to preserve backwards-compatibility and have made a PR in the OAuth2 server repo for it: https://github.com/thephpleague/oauth2-server/pull/941

As such, it might be worth waiting on that other PR to get merged in and a new version of the package to be released before this PR is merged.

taylorotwell:
OK. Feel free to re-submit this when that is resolved.

EricTendian:
@taylorotwell based on the feedback in the other PR, it doesn't seem like this will be able to be a backwards-compatible change. Do you want to reconsider this for the next major release?

PR #825, opened 2018-09-18
Clean up & reformate code & fix phpdoc errors

qWici:


taylorotwell:
Some of these changes aren't really Laravel standard. For example, we never explicitly return "null". We just return, etc.

Issue #406, opened 2017-06-14
Trying to get property of non-object in ClientRepository.php (line 80)


prabhashkutti:
After some time when trying to create personal access token using vue components, it throws **Trying to get property of non-object in ClientRepository.php (line 80)**

It works normal again after running the command **php artisan passport:install** 

But if I run **php artisan migrate:refresh** , again got into the same issue..

prabhashkutti:
It works normal again after running  the command **php artisan passport:install** once again

alexbilbie:
Which version of Passport are you using?

Beaudinn:
same here, i use  Passport 3.0

borger:
I'm having the same issue, exactly the way you describe (also how to fix it), but to me, it says its on line #81. I'm on version ^3.0

alexbilbie:
Please can you paste some stack traces as there just isn't enough information to go on here

nfunwigabga:
I had the same issue when i upgraded to ^3.0. I ran `php artisan passport:install --force` to regenerate keys and it worked.

borger:
Hello @alexbilbie I apologize for the delay in providing more info, below is a stack trace from the log.

I believe this is just an error caused by the lack of the pre-requirement `php artisan passport:install`. 

To replicate the error, start a new laravel project, add passport on composer, and on `config/app.php` (ServiceProvider), and in my case, I've tried to create a new personal-access-token via the Vue component after a fresh migration.

```
[2017-08-09 06:07:15] local.ERROR: ErrorException: Trying to get property of non-object in /Users/username/Documents/Sites/laravel-project/vendor/laravel/passport/src/ClientRepository.php:81
Stack trace:
#0 /Users/username/Documents/Sites/laravel-project/vendor/laravel/passport/src/ClientRepository.php(81): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(8, 'Trying to get p...', '/Users/username/...', 81, Array)
#1 /Users/username/Documents/Sites/laravel-project/vendor/laravel/passport/src/PersonalAccessTokenFactory.php(71): Laravel\Passport\ClientRepository->personalAccessClient()
#2 /Users/username/Documents/Sites/laravel-project/vendor/laravel/passport/src/HasApiTokens.php(67): Laravel\Passport\PersonalAccessTokenFactory->make(1, 'test', Array)
[...]

PR #828, opened 2018-09-22
Display clear message when personal access client is missing

gauravmak:
As per [this issue](https://github.com/laravel/passport/issues/406), developers often forget to re-create personal access client after refreshing their migrations.

The error when they try to create a personal access token after that is quite vague: 'Trying to get property of non-object in ClientRepository.php'

This PR adds the code to throw an exception with the clear message in that case.

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Passport;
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
I'm not sure we need our own custom exception here. Can you re-submit this just using a built-in PHP exception?

PR #844, opened 2018-10-13
Set $timestamps to true as these are in the default migration

SolStis86:
Sets timestamps to true in Token model as timestamps are present in the default migration.


taylorotwell:
Not changing this on a minor version.

driesvints:
@taylorotwell this is also related to https://github.com/laravel/passport/issues/786. Can you perhaps review that one?

Issue #599, opened 2018-01-07
[Feature Request] Add route names to Passport::routes()


d8vjork:
Hi all :raising_hand_man:

As title says, modifying [passport/src/RouteRegistrar.php](https://github.com/laravel/passport/blob/4.0/src/RouteRegistrar.php) file:
```
$this->router->post('/token', [
        'uses' => 'AccessTokenController@issueToken',
	'as' => 'oauth.token',
        'middleware' => 'throttle',
]);
```

nejtr0n:
Same issue. it would be greate to have names of route to use them in custom controllers like
```
        $request = Request::create('/oauth/token', 'POST', [
            'grant_type' => 'password',
            'client_id' => config('services.vue_client.id'),
            'client_secret' => config('services.vue_client.secret'),
            'username' => $request->email,
            'password' => $request->password,
        ]);
```
Proble is in
```
        $request = Request::create('/oauth/token', 'POST', [
```
This route could be prefixed with api/v1 for example, 
and route names will help to avoid
hardcoding of it.

driesvints:
I'm not sure if it'll get merged but you can always try to send in a PR. I can see this being valuable.

PR #846, opened 2018-10-16
[7.0] Add names to routes for reusability

d8vjork:
As proposed in #599


taylorotwell:
Why `oauth` and not `passport`?

d8vjork:
> Why `oauth` and not `passport`?

Yes, is another option 🤔 I'll update it later

Issue #143, opened 2016-10-15
Associate client_credentials grant to a user


OscarRPR:
Hello

Currently, I'm building an API where my users will be able to create their API keys from my platform and then use them to perform requests, I look for OAuth 2.0 (given that in the future I'll support the Authorization Code and I have a mobile application that uses the password grant), and for this case I want to follow PayPal authentication system (https://developer.paypal.com/docs/integration/direct/paypal-oauth2/).

But I have an issue trying to use the client_credentials grant, based on the RFC 6749 of OAuth 2.0:

> 4.4.  Client Credentials Grant
> 
>    The client can request an access token using only its client
>    credentials (or other supported means of authentication) when the
>    client is requesting access to the protected resources under its
>    control, or those of another resource owner that have been previously
>    arranged with the authorization server (the method of which is beyond
>    the scope of this specification).

If I understand correctly, the client credentials can be associated to a resource owner or can have a null user instance to access any public resources available.

In my case, I want that the client_credentials are associated to a user in my system, and then, the access token can be related to it, but currently, the ClientCredentials grant sets the user instance as null.

I thought that one solution was create a new Service Provider based on the PassportServiceProvider, create a custom client credentials grant and then in the new service provider, enable the grant, but I don't want to keep maintaining the new ServiceProvider with the latest changes of PassportServiceProvider.

The other solution was to create a new middleware to check if the access token has the client_credentials grant, then, I'll look up for the client_id and then, get the user id, but I found this not useful because the JWT token should have this information.

I'm not sure which path to follow or if I miss something.

Thanks

dreferreira:
I've done a little bit of investigating into this using the password grant as noted below. I hope you find it insightful. **Warning it's fairly long**
#151

themsaid:
@OscarRPR the client_credentials requires no user, it gives access to publish resources that aren't owned by any user, so a user permission is not required.

OscarRPR:
@dreferreira Thank you, I'll keep an eye on it to check for a solution.

@themsaid But, if you read the OAuth 2.0 specification in the original post says:

The client can request an access token using only its client
credentials (or other supported means of authentication) when the
client is requesting access to the protected resources under its
control,

Which if I understand correctly, it means that the client can request access to his own resources, then, we will need a way to configure it, no?

Or, as I mentioned in the post, PayPal says that he uses OAuth 2.0 with client_credentials grant, and with the keys I generate I can perform calls for my account on protected resources, then, they're not really using OAuth 2.0?

[...]

PR #850, opened 2018-10-19
[8.0] Add ability to pull user from client

mattkoch614:
- Added a new relationship to the ``Client`` class to easily retrieve a user.
- Added a method to the ``ClientRepository`` class to retrieve the client based on a token.
- This PR is in support of discussion in #143

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -3,18 +3,9 @@
 namespace Laravel\Passport;
 
 use Illuminate\Database\Eloquent\Model;
-use Laravel\Passport\Bridge\User;
-use Laravel\Passport\Bridge\UserRepository;
 
 class Client extends Model
 {
-
-
-    public function __construct(array $attributes = [])
-    {
-        parent::__construct($attributes);
-
-    }
 
     /**
      * The database table used by the model.
--- a/src/Client.php
+++ b/src/Client.php
@@ -3,9 +3,19 @@
 namespace Laravel\Passport;
 
 use Illuminate\Database\Eloquent\Model;
+use Laravel\Passport\Bridge\User;
+use Laravel\Passport\Bridge\UserRepository;
 
 class Client extends Model
 {
+
+
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct($attributes);
+
+    }
+
     /**
      * The database table used by the model.
      *
@@ -69,5 +79,17 @@ class Client extends Model
     public function firstParty()
     {
         return $this->personal_access_client || $this->password_client;
+    }
+
+    /**
+     * Return the user associated with the client.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function user()
+    {
+        $provider = config('auth.guards.api.provider');
+
+        return $this->belongsTo(config('auth.providers.'.$provider.'.model'));
     }
 }
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Passport;
 
+use Lcobucci\JWT\Parser;
+
 class ClientRepository
 {
     /**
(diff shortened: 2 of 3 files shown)


mattkoch614:
I need some help getting my test to pass - the issue is that the ``ClientRepository`` class is trying to connect to a database which doesn't exist.

src/Client.php
```diff
@@ -6,6 +6,7 @@
 
 class Client extends Model
 {
+
```
driesvints: Remove this line

src/Client.php
```diff
@@ -70,4 +71,16 @@ public function firstParty()
...
+     * Return the user associated with the client.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function user()
+    {
+        $provider = config('auth.guards.api.provider');
+
+        return $this->belongsTo(config('auth.providers.'.$provider.'.model'));
+    }
```
driesvints: Move this above the `firstParty` method.

tests/ClientRepositoryTest.php
```diff
@@ -0,0 +1,33 @@
...
+            ->setId('4f1g23a12aa', true) // Configures the id (jti claim), replicating as a header item
+            ->setIssuedAt(time())
+            ->setHeader('typ', 'JWT')
+            ->setHeader('alg', 'HS256')
+            ->setExpiration(time() + 604800)
+            ->sign(new Sha256(), 'test.jwt.signature')
+            ->getToken();
+    }
+
+}
```
driesvints: Remove this test. There wasn't a `ClientRepository` test probably because we can't setup a DB connection atm.

src/ClientRepository.php
```diff
@@ -47,6 +49,20 @@ public function findForUser($clientId, $userId)
...
+     *
+     * @param $bearerToken
+     * @return \Laravel\Passport\Client|null
+     */
+    public function findByToken($bearerToken)
+    {
+        $clientId = (new Parser())->parse($bearerToken)->getClaim('aud');
+
+        return $this->find($clientId);
+    }
```
driesvints: I don't think it's a good idea to move this logic to the `ClientRepository`. You're mixing parsing tokens with retrieving clients which are two separate things. The `ClientRepository` only job is to retrieve client models.

A better place for this method would be the `TokenGuard` class. Try to see if you can implement a `client` method like the `user` method:

```
    /**
     * Get the user for the incoming request.
     *
     * @return mixed
     */
    public function user(Request $request)
    {
        if ($request->bearerToken()) {
            return $this->authenticateViaBearerToken($request);
        } elseif ($request->cookie(Passport::cookie())) {
            return $this->authenticateViaCookie($request);
        }
    }
```

driesvints:
If you move the logic to the `TokenGuard` class you can add a test in the `TokenGuardTest` class: https://github.com/laravel/passport/blob/7.0/tests/TokenGuardTest.php

driesvints:
Btw this is a suggestion but can you send the method with the relationship on the client model as a separate pr to the 7.0 branch? It doesn't has anything to do with the other functionality.

I'm also going to close this PR. Feel free to resend it once you've worked through the feedback above. You can also send the PR for retrieving the client as a PR to the 7.0 branch as it has no breaking changes.

mattkoch614:
@driesvints No problem. I'll review your feedback and re-submit.

nolanpro:
@mattkoch614 Thanks for your help on this. We need this functionality too. I tested this in TokenGuard  like @driesvints suggested and it works for me.

In authenticateViaBearerToken function:
```php
$user = $this->provider->retrieveById(
    $psr->getAttribute('oauth_user_id')
);

// Added this:
if (! $user) {
    $bearerToken = $request->bearerToken();
    $clientId = (new \Lcobucci\JWT\Parser())->parse($bearerToken)->getClaim('aud');
    $user = $this->provider->retrieveById(\Laravel\Passport\Client::find($clientId)->user_id);
}

if (! $user) {
    return;
}
```

PR #854, opened 2018-10-22
[7.0] Add the ability to retrieve current client

driesvints:
This commit adds the ability to retrieve the current client from the request, just like the current user is retrieved. I chose to do this in the TokenGuard class since this class uses the same logic to retrieve the current active user.

Usage could be as follows:

    $client = $tokenGuard->client($request);

One of the biggest reasons for adding this is that there isn't any way at the moment to retrieve the current client from the request. By adding this, users can use the client to further perform authorization actions or check the creator of the client and subsequently limit resources bases on either one of those two.

This is especially helpful for client credentials grant requests where you simply don't have an active/authed user. This way you can still limit resources if you want based on either the client or its creator.

This commit is fully BC. No methods have been renamed, only added. The ones that have been modified still behave in the same way as before but only have some parts extracted to other methods so their code could be re-used.

This solves the following long outstanding issue: https://github.com/laravel/passport/issues/143

PS. I'm sorry if the changes are too long or confusing. Feel free to ask questions and I will gladly clarify and explain further.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -93,6 +93,29 @@ class TokenGuard
     }
 
     /**
+     * Get the client for the incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return mixed
+     */
+    public function client(Request $request)
+    {
+        if ($request->bearerToken()) {
+            if (! $psr = $this->getPsrRequestViaBearerToken($request)) {
+                return;
+            }
+
+            return $this->clients->findActive(
+                $psr->getAttribute('oauth_client_id')
+            );
+        } elseif ($request->cookie(Passport::cookie())) {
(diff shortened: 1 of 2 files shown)


TimWolla:
@driesvints Can you clarify how I would pull the client from the guard in the real world? It looks like it does not properly `mixin` the `client` method into the `RequestGuard`:

```
$client = Auth::guard('api')->client($request);
```
fails with a “Method Illuminate\\Auth\\RequestGuard::client does not exist.”. I should not need to construct the `TokenGuard` manually, no?

Am I doing this wrong?

rizaldywirawan:
The only think that i do not understand is, Laravel has a super cool documentation, always make dev easy to use it and find something they really need. 

This one little problem already haunted lots of developers out there just to find a client id, why is this not in the documentation?

PR #863, opened 2018-11-03
Add client avatar when creating clients

richard457:


richard457:
Added ability to add client_avatar as image string url for nice branding of client created.

PR #866, opened 2018-11-08
[7.0] Fixed inconsistency of where('user_id', $userId)

ArlonAntonius:
- Changed the places that use `where('user_id', $userId)` to `whereUserId($userId)` to make the use consistent


taylorotwell:
It's faster to pass the column explicitly.

Issue #869, opened 2018-11-09
Client Grant requires users table


sasa-b:
Currently, the __Client Grant__ implementation requires the presence of the `users` table, otherwise if the table is not present and we try to authorize we get the following exception:

```json
{ 
    "message": "SQLSTATE[42S02]: Base table or view not found: 1146 Table 'my_db.users' doesn't exist (SQL: select * from `users` where `id` =  limit 1)",
    "exception": "Illuminate\\Database\\QueryException",
    ...
}
```

Since the Client Grant should be used for machine to machine communication the `users` table should not be mandatory.

The issue happens in the `Laravel\Passport\Guards\TokenGuard`:
```php
$user = $this->provider->retrieveById(
      $psr->getAttribute('oauth_user_id')
 );

if (! $user) {
     return;
}
```
The provider tries to fetch the user and everything blows up. We could easily circumvent this with an additional if statement, namely, if a `user_id/oauth_user_id` is null in the `oauth_tokens` table, don't  try to fetch the user. Something like this:

```php
$userId = $psr->getAttribute('oauth_user_id');

if ($userId) {
     $user = $this->provider->retrieveById($userId);

      if (! $user) {
          return;
      }
}
```

I'd be happy to make a pull request if that's OK with the maintainers?

driesvints:
I'm not really sure how this can happen. A client credentials grant shouldn't attempt to retrieve a user in the first place I believe? Or am I missing something?

sasa-b:
@driesvints Well it shouldn't, but here is the catch. Laravel requires the `auth.php` config `driver` and `provider` to be set. If you use `passport` as a driver or even any other, when you use the `\Laravel\Passport\Http\Middleware\CheckClientCredentials::class` for the Client Access Token the `Laravel\Passport\Guards\TokenGuard` is used and then happens what I described previously.

I'll make a PR, hope that is OK with you?

driesvints:
I'm a bit concerned with edge cases and what this will break in people's setups. I think we need to investigate this more first.

sasa-b:
I don't think it will break anything, because when we use the Client Grant there is no `user_id` associated with the token, and guard provider should not try to fetch the user. 

In Client Grant scenario we are only concerned if the token is valid, e.g. if it has not expired or if it has not been revoked etc.

When we use other scenarios, the `user_id` is set on the tokens and the provider should and will try to fetch the user.

Then again, I might have jumped the gun, I was just eager to make my first open source contribution :D 

[...]

PR #871, opened 2018-11-09
#869 Update to TokenGuard.php

sasa-b:
Client Grant no longer requires the users' table to be present.


taylorotwell:
No plans on merging.

PR #877, opened 2018-11-12
Compatible with DB::first() return array

szuwwy:
Compatible with DB::first() return array

reason:

App/Providers/EventServiceProvider.php boot method add code as follow to support global object to array

//PDO::FETCH_ASSOC
Event::listen(StatementPrepared::class, function ($event) {
            $event->statement->setFetchMode(\PDO::FETCH_ASSOC); 
});

thanks.


driesvints:
Please branch off master instead of 7.0

Issue #884, opened 2018-11-18
auth:api 401 Unauthorized - TokenGuard::decodeJwtTokenCookie should not decrypt cookie values


benallfree:
If anyone is seeing the `auth:api` guard failing with 401 Unauthorized and you have already played with cookie serialization settings, don't worry - you are not crazy.  I think I know why.

I'm not an expert in this area of the framework, but I believe there is a bug in [this line](https://github.com/laravel/passport/blob/7.0/src/Guards/TokenGuard.php#L241) of `TokenGuard` because performs unnecessary decryption. Cookie decryption is handled by `EncryptCookie` middleware, so this code results in duplicate decryption:

```
    protected function decodeJwtTokenCookie($request)
    {
        return (array) JWT::decode(
            $this->encrypter->decrypt($request->cookie(Passport::cookie()), Passport::$unserializesCookies),
            $this->encrypter->getKey(), ['HS256']
        );
    }
```

Should be

```
    protected function decodeJwtTokenCookie($request)
    {
        return (array) JWT::decode(
            $request->cookie(Passport::cookie()),
            $this->encrypter->getKey(), ['HS256']
        );
    }
```

I think this bug was introduced because `TokenGuardTest` generates its own test values rather than using values from `ApiTokenCookieFactory::createToken`. If the test suite consumed values created by `ApiTokenCookieFactory`, this bug would become apparent in testing.

**Problem**

`auth:api` guard fails with 401 Unauthorized

**Diagnosis**

`TokenGuard` is attempting to decrypt cookie values generated by `ApiTokenCookieFactory`, which does not encrypt them in the first place.

**Solution**

1. Update `TokenGuardTest` to use test values produced by `ApiTokenCookieFactory`
2. Remove decryption from `TokenGuard`

See PR #886 for the fix. Previous attempt and discussion in PR #885.

benallfree:
#885

benallfree:
Reopening based on #885 discussion.

driesvints:
FWIW: if I find some time I'll deep dive into this but feel free to add remarks and insights if you find some.

benallfree:
Following up and closing. The issue is that `EncryptCookies` should not be used in the `api` group because it’s supposed to be stateless. 

I’d prefer to see a more compatible solution where cookie middleware can be used, but for now that’s the answer.

PR #888, opened 2018-11-20
[8.0] auth:api 401 Unauthorized - TokenGuard::decodeJwtTokenCookie should not decrypt cookie values 

benallfree:
This PR has corrected passing tests to follow up #885 (discussion in #884), hopefully we can leave it open until the issue is resolved.

@driesvints See how I updated `TokenGuardTest` to consume real tokens from `ApiTokenCookieFactory` like it would in a full Laravel App. You can now see that the additional decryption when decoding JWTs in `TokenGuard` would fail.

I hope this better demonstrates the problem :)

One note: I didn't see any tests for unserializing cookies, and since the serialization option is no longer used, I removed it from the `Passport` class as well. I think this as well does not belong in Passport and should be handled by middleware. Cookie value encrypt/decrypt and serialize/unserialize are middleware tasks right?

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -187,7 +187,7 @@ class TokenGuard
     protected function decodeJwtTokenCookie($request)
     {
         return (array) JWT::decode(
-            $this->encrypter->decrypt($request->cookie(Passport::cookie()), Passport::$unserializesCookies),
+            $request->cookie(Passport::cookie()),
             $this->encrypter->getKey(), ['HS256']
         );
     }
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -118,13 +118,6 @@ class Passport
     public static $runsMigrations = true;
 
     /**
-     * Indicates if Passport should unserializes cookies.
-     *
-     * @var bool
-     */
-    public static $unserializesCookies = false;
-
-    /**
      * Enable the implicit grant type.
      *
      * @return static
@@ -510,30 +503,6 @@ class Passport
     public static function ignoreMigrations()
     {
         static::$runsMigrations = false;
-
-        return new static;
-    }
-
-    /**
-     * Instruct Passport to enable cookie serialization.
-     *
-     * @return static
-     */
-    public static function withCookieSerialization()
-    {
-        static::$unserializesCookies = true;
-
-        return new static;
-    }
-
-    /**
-     * Instruct Passport to disable cookie serialization.
-     *
-     * @return static
-     */
-    public static function withoutCookieSerialization()
-    {
-        static::$unserializesCookies = false;
 
         return new static;
     }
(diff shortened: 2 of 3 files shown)


driesvints:
I really believe we should be more careful with making changes like these and gather opinions from other users first. This *looks* like twtg but I'm not sure or convinced of it yet to give this my full support. I'll need more time to figure this out and test certain scenarios. I don't have time at this point but will have a look at a later moment.

taylorotwell:
Yeah, the explanation makes sense that there should not be any reason to decrypt that cookie on a default Laravel 5.7 installation.

taylorotwell:
I'm unable to recreate this bug on latest Laravel 5.7 and latest Passport 7.0:

image

image

benallfree:
@driesvints Agree, definitely a breaking change. Hopefully this PR is good for demonstration purposes.

@taylorotwell Add `EncryptCookies` to the `api` middleware and things will break as expected. When would it make sense to run `EncryptCookies` on just `web` but not `api`? Or is that the source of my entire misunderstanding?

driesvints:
@benallfree that's it. You shouldn't encrypt cookies on the api middleware group. Your API is stateless and doesn't know anything about cookies. The encryption here is performed when using tokens through the web middleware.

benallfree:
@driesvints Ah, interesting. So the intent is that any code run as a result of an `api` entry point does not access cookies? At first thought that would seem a surprising assumption. `CreateFreshApiToken` and the `auth:api` guard are at least one exception to that. I see now why Passport has manual decryption in it :)

Would the cleaner solution be to use `EncryptCookies` on `api`?

driesvints:
No `EncryptCookies` doesn't makes any sense because cookies aren't used for api requests. I also don't understand how this can be a problem really since the `authenticateViaCookie` is only called when the request has a cookie set. The `$request->bearerToken()` check is performed first and should be called first when you have a bearer token with your api request.

diegofelix:
Hi.

I'm having exactly the same problem, but with Laravel 5.5. Is that a problem in using Passport 7.0 with Laravel 5.5? I compared Laravel 5.5 EncryptCookie@encrypt with Laravel 5.7 and seems pretty the same.

benallfree:
@diegofelix do you have `EncryptCookies` running on your `api` group?

diegofelix:
@benallfree No.

diegofelix:
I tried to to add on the api group, add on EncryptCookies$except, but the only solution is removing the `$this->encrypter->decrypt` part of the TokenGuard.

I cant even upgrade laravel to 5.7 for now because we use mongo and have so much things to do to upgrade =(

benallfree:
The `api` group is supposed to be stateless, so cookie related functions and middleware lik `EncryptCookies` should no be used in `api` code paths.

diegofelix:
@benallfree Yeah, I know, I'm not using it on api group. I just tried to clear the possibilities.

PR #892, opened 2018-11-27
Fix(saving personal client was malformed)

richard457:


driesvints:
Please don't commit your `.idea` settings. You've also accidentally included a commit which changes the branch alias.

Issue #912, opened 2018-12-20
Client ID is unsigned, client_ids are signed


tentus:
Wait wait wait, this isn't yet another of those requests about changing the key type. I just wanted to point out that the migrations, by default, create _signed_ integers for the `client_id` columns. This is out of line with the `increments` of `id` field in the `Client`, which is unsigned.

We someone to apply foreign keys after the fact, they'd need to change the columns to realign the possible values. This is as simple as calling `$table->unsignedInteger('client_id')->change();`, but it's a step that shouldn't be necessary.

driesvints:
Ah, you're right. Good catch 👍 

I'll try to send in an update for this in.

driesvints:
This was merged on 8.0 and cherry picked on 7.0. We'll probably release a patch soon, in the meantime you can overwrite the migrations.

PR #913, opened 2018-12-20
[8.0] Use unsignedInteger column type for client_id columns

driesvints:
This is because the client id autoincrements. We don't add foreign key indexes ourselves but if you try to add them afterwards you'll have to first change the column type of the client_id columns because otherwise they don't match the increments column. We can immediately use the proper type in the migrations here.

Sending this to the 8.0 branch because I wasn't too sure if this would be breaking or not so better play it safe.

Fixes https://github.com/laravel/passport/issues/912


taylorotwell:
You could probably send this to 7.0 branch too.

driesvints:
@taylorotwell cherry picked on 7.0

PR #921, opened 2019-01-03
[7.0] Added redirect_uri and user_id options to cli

matthewsuan:
When using `passport:client` commands, add `redirect_uri` and `user_id` options. Useful for automated builds.


driesvints:
Can you give a specific example how this would be useful?

matthewsuan:
I found myself wanting this feature for my ci builds and testing scripts.

I have deploy scripts where I am creating new oauth client on the fly and this is not directly achievable using current setup.

I found out `--name` option is added so why not `--user_id` and `--redirect_uri`.

matthewsuan:
cool thanks!

PR #926, opened 2019-01-10
[7.0] Make name an optional question

ankurk91:
The command ask a required question  but allows user to skip; which cause failure in mysql insert query.
Just press enter when the command asks
"What should we name the client?:"

```
php artisan passport:client --client
```

This PR will auto guess the name when no name is specified or input by user.


ankurk91:
We should also put some validation on `createAuthCodeClient`

https://github.com/laravel/passport/blob/514e2d551af6b87baf182dfa2dff7f203997e376/src/Console/ClientCommand.php#L122

Issue #917, opened 2018-12-27
Refresh tokens can't be used after access token is expired and pruned?


FlsZen:
My app has an access token (4 hour expiry) and refresh token (1 year expiry). Once the access token expires I see it is removed from the `oauth_access_tokens` table. My refresh token is still in the `oauth_refresh_tokens` table since it hasn't expired (and `revoked=0`).

When my app makes the `refresh_token` call to `oauth/token`, instead of getting a new access token and refresh token, it gets:

```
{
    "error": "invalid_request",
    "message": "The refresh token is invalid.",
    "hint": "Token has been revoked"
}
```

After this response, I can still see my refresh token is in the `oauth_refresh_tokens` table with `revoked=0` and the expiry date is still about a year out.

The RFC seems to indicate refresh tokens are the way to go since they can be used to get a new access token after a previous access token expires, but that isn't the behavior I'm seeing.

As a sanity check, I have verified that I can use the refresh token to get a new access and refresh token pair up until the point that the access token is pruned. If the access token is expired but not yet pruned it still works. It goes sideways once it's been pruned.

Is this the expected behavior? If so, is there something I can do to use a refresh token after the access token has expired & been pruned?

driesvints:
Afaik I think you should indeed still be able to use the refresh token even though the access token is pruned. Maybe @sephster knows what the correct behavior is?

Sephster:
Yep, that's correct. You should be able to use the refresh token regardless. I did try to take a quick look into why this is occurring but didn't get very far. 

It should be the case that the refresh token has all the information required to generate a new access token.

Sephster:
Hey @driesvints. I think it is fine to use the refresh token, even if the access token no longer exists. We actually store the original access token entity in the refresh token so we can compare scopes etc so there shouldn't be a need to keep it (if I am remembering correctly).

The problem seems to be the function `isRefreshTokenRevoked` in the `RefreshTokenRepository` which is also checking if the access token is revoked. I don't think this is the correct behaviour.

driesvints:
Okay. I'll have a look at this soonish. Thanks for helping out @Sephster! :)

driesvints:
Sent in a PR for this: https://github.com/laravel/passport/pull/927

TimWolla:
@driesvints I was the person that changed this in #391 as part of another fix (deleted entries are valid for ever). Please consider the following:

[...]

PR #927, opened 2019-01-10
[7.0] Fix revoked refresh token check

driesvints:
When a check if performed to determine if a refresh token is revoked or not, it's unnecessary to also check if its corresponding access token is revoked. It's perfectly possible to issue a new access token if the old access token is either revoked or pruned as long as the refresh token is valid.

Fixes https://github.com/laravel/passport/issues/917


driesvints:
Btw: I've removed the `AccessTokenRepository` dependency from the `RefreshTokenRepository`. Unless someone is extending this class it shouldn't be breaking.

Issue #289, opened 2017-02-22
How to work with OAuth errors


lotarbo:
Hi, what is the right way to custom or work with errors? I mean when we have invalidClient error, this auth popup is bad for me, mb better redirect to client site with some message?
And how change error? I read  http://www.kingpabel.com/oauth2-exception-custom-error-message/
but it works only for clean League\OAuth2, passport has own HandlesOAuthErrors, can somebody help?

sebastiaanluca:
Running into the same issue. Thought my exception handler was botched, but turns out Passport intercepts all errors and returns a plain text string. That's not really useful, to just output raw exceptions to the user without a chance to format these as a developer?

@themsaid @taylorotwell What's the course here to disable/extend this custom Passport exception handler?

Similar issue: https://github.com/laravel/passport/issues/97

deividaspetraitis:
Not sure what's wrong to catch and handle these in Hander.php:

`    /**
     * Determine if the given exception is an OAuth2 exception.
     *
     * @param Exception $e
     *
     * @return bool
     */
    public function isOAuthException(Exception $e)
    {
        return $e instanceof \League\OAuth2\Server\Exception\OAuthServerException
                ||
                $e instanceof AuthenticationException;
    }`

In case you want to determine specific exception you can use `$e->getErrorType()`.

EdwardNelson:
The exceptions do not bubble up to the exception handler. Passport catches them

deividaspetraitis:
Yea, seems you guys are right. 

If I'm correct error handling happens using predefined controllers that calls  `$this->withErrorHandling()` function that refers to `HandlesOAuthErrors` trait.

Such handling is not useful not only in case when you do need to catch exception but there is some headache  when you do need to return these errors in some other format for example JSON, XML, etc.

What I have done so far in order to solve this I just created my own controllers and removed such handling. 
That means that you must to:

-  Emit `Passport::routes();` from your `AuthServiceProvider ` class `boot` method,  
-  Copy original contoller that will issue tokens ( in this example let's consider it will be `AccessTokenController` ) and remove `withErrorHandling()`
- Define manually routes for your created controller

For example: Please see original `AccessTokenController`: https://github.com/laravel/passport/blob/2.0/src/Http/Controllers/AccessTokenController.php and my re-defined: https://gist.github.com/deividaspetraitis/7c3958381d33a06b511d17e2e22b8bb5

erkanarslan:
[...]

PR #937, opened 2019-01-17
[8.0] Rework HandlesOAuthErrors trait to middleware

driesvints:
The benefit this gives is that it allows developers to overwrite the middleware and implement their own handling of errors if they want to. At the moment they were stuck with the implementation being a trait so the only way to overwrite the implementation was to rebind their own controllers.

With the new implementation you could bind a new middleware which extends the current one like so:

```php
$this->app->bind(HandleOAuthErrors::class, function () {
    // return your own implementation...
});
```

At least it'll become way easier for developers to customize this than before. It also makes much more sense to handle the error handling in a middleware than through a trait. Given, the previous implementation was probably just an old one and just needed a refresh. 

Added tests for the new middleware and removed the old one. Also removed the tests for the controllers (which aren't needed anymore now).

Sent this in to master because of the obvious breaking changes.

Fixes https://github.com/laravel/passport/issues/289


taylorotwell:
What will the upgrade path for existing Passport users?

TimWolla:
@driesvints Was this PR tested from within a Laravel app? We upgraded to `dev-master` passport within an Laravel 5.8 application to get PKCE support and noticed that OAuth errors result in a HTTP 500 now.

The issue being that Laravel itself catches Exceptions happening in the controller within the middleware pipeline (even in `master` of laravel/framework): https://github.com/laravel/framework/blob/9097786caae30e9ca243874e5c807609f1f2728e/src/Illuminate/Pipeline/Pipeline.php#L133, so they don't actually reach the `HandleOAuthErrors` middleware.

driesvints:
@TimWolla hmm that's unfortunate. I don't have time myself to look at this atm so feel free to open up an issue for this.

TimWolla:
@driesvints ack. I've filed #1062 to track this issue.

PR #938, opened 2019-01-20
Replaced some helpers with Blade directives

andrewgorpenko:
I've replaced "massive" helpers with Blade's new directives (@csrf and @method). I think it is much cleaner.


taylorotwell:
May be best to send this to master in case people are using Passport with a version of Laravel that doesn't have these?

andrewgorpenko:
May I recreate this Pull Request one more time applying to the master?

PR #940, opened 2019-01-21
[7.0] Add ext-json dependency

driesvints:
Was missing apparently.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     ],
     "require": {
         "php": ">=7.1",
+        "ext-json": "*",
         "firebase/php-jwt": "~3.0|~4.0|~5.0",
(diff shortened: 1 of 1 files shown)


taylorotwell:
Is it even possible to have PHP 7.1 without the JSON extension? Isn't it built in?

driesvints:
@taylorotwell yes but it's always a good idea to add them as a requirement to your `composer.json` to know exactly which libraries/extensions your project depends on. In the situations where you have a custom build of PHP this can be a good gotcha for your to know why you can't install a library like Passport. PHPStorm has started throwing warnings if these dependencies aren't added:

https://medium.com/@luis.barros.nobrega/ext-json-is-missing-in-composer-json-what-is-this-700a1fcad966

Issue #687, opened 2018-04-06
AuthorizedAccessTokenController->destroy returns void on success


goodevilgenius:
https://github.com/laravel/passport/blob/bd5ff54598d461f89cb8eaf803f4906f77b559bf/src/Http/Controllers/AuthorizedAccessTokenController.php#L61

When a token is successfully deleted, a valid response should be returned (probably `new Response('', 204)`.

However, nothing is returned, resulting in an error.

goodevilgenius:
Sorry about this. My frontend guys told me there was an error, and I didn't double-check it. This actually does return a valid response.

michaelklopf:
Similar situation here, Ember expects a response, but the one coming is no fit for Ember.

https://stackoverflow.com/questions/20648372/what-does-ember-data-expect-in-response-to-deleting-a-record

driesvints:
Seems like this is indeed a problem.

PR #942, opened 2019-01-21
[7.0] Add responses for destroy methods

driesvints:
This adds proper responses and HTTP statuses for the destroy methods on success.

Fixes https://github.com/laravel/passport/issues/687

### DIFF
--- a/src/Http/Controllers/AuthorizedAccessTokenController.php
+++ b/src/Http/Controllers/AuthorizedAccessTokenController.php
@@ -60,6 +60,6 @@ class AuthorizedAccessTokenController
 
         $token->revoke();
 
-        return new Response('Token successfully revoked.', Response::HTTP_NO_CONTENT);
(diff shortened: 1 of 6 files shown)


src/Http/Controllers/AuthorizedAccessTokenController.php
```diff
@@ -59,5 +59,7 @@ public function destroy(Request $request, $tokenId)
         }
 
         $token->revoke();
+
+        return new Response('Token successfully revoked.', Response::HTTP_NO_CONTENT);
```
goodevilgenius: `204` is intended to indicate that there is no body in the response.

See https://httpstatuses.com/204:

> A 204 response is terminated by the first empty line after the header fields because it cannot contain a message body.

```suggestion
        return new Response('', Response::HTTP_NO_CONTENT);
```

driesvints: You're correct. Fixed.

PR #945, opened 2019-01-23
Configurable Refresh Tokens

donovanhare:
This adds a configurable `RefreshToken` model and a new repository. The changes hopefully bring the same flexibility access tokens, clients, authcodes etc. enjoy, making refresh tokens easier to tinker with.

These changes are particularly helpful when you want to use eloquent to query the refresh token table from your application or if you change the refresh token table name.

### DIFF
--- a/src/Bridge/RefreshTokenRepository.php
+++ b/src/Bridge/RefreshTokenRepository.php
@@ -4,6 +4,7 @@ namespace Laravel\Passport\Bridge;
 
 use Illuminate\Database\Connection;
 use Illuminate\Contracts\Events\Dispatcher;
+use Laravel\Passport\RefreshTokenRepository as RTokenRepository;
 use Laravel\Passport\Events\RefreshTokenCreated;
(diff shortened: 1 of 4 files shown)


donovanhare:
I'm having trouble writing _meaningful_ tests without essentially mocking all of the functionality of the repositories. If @driesvints can advise, I'd be grateful.

src/Bridge/RefreshTokenRepository.php
```diff
@@ -4,18 +4,19 @@
 
 use Illuminate\Database\Connection;
 use Illuminate\Contracts\Events\Dispatcher;
+use Laravel\Passport\RefreshTokenRepository as RTokenRepository;
```
donovanhare: I'm not happy with this naming here, would `TokenRepository` be more appropriate?

driesvints:
Heya, tests seems to be failing and no new tests were added for this. Besides that it's best to send this to the master branch as it contains some breaking changes. Can you do these changes and resend? Thanks.

PR #946, opened 2019-01-23
Make refresh tokens configurable

donovanhare:
This adds a configurable `RefreshToken` model and a new repository. The changes hopefully bring the same flexibility access tokens, clients, authcodes etc. enjoy, making refresh tokens easier to tinker with.

These changes are particularly helpful when you want to use eloquent to query the refresh token table from your application or if you change the refresh token table name.

Let me know if you would like anything else changed.


driesvints:
Can you also add some tests for the new functionality?

Issue #516, opened 2017-09-16
Cant grant * scopes on Client Credentials grant


philcross:
I'm having an issue whereby If I try to grant the scopes `*` on a `client_credentials grant, the `ScopeRepository` is stripping out the scope.

It's referenced in this issue https://github.com/laravel/passport/issues/319 but from the look of it, it seems to have been resolved, but I cant work out how. 

I'm creating an access token with the following:

```
[
    'grant_type'    => 'client_credentials',
    'client_id'     => $client->id,
    'client_secret' => $client->secret,
    'scope'         => '*',
]
```

In `Laravel\Passport\Bridge\ScopeRepository`, in the `finalizeScopes()` method, the `scopes` argument contains an scope instance containing `*`, and the `grantType` property is `client_credentials`.

But it appears this method removes `*` from client credential grants.

As a workaround, I am doing:

```
[
    'grant_type'    => 'client_credentials',
    'client_id'     => $client->id,
    'client_secret' => $client->secret,
    'scope'         => implode(' ', Passport::scopeIds()),
]
```

But I think either the docs should explictly mention that `*` is not allowed on client credential grants, or by adding `client_credentials` to the array of grants which allow `*` to be specified.

philcross:
This test https://github.com/laravel/passport/blob/4.0/tests/BridgeScopeRepositoryTest.php#L33-L35 seems to make you believe that `*` shouldn't be allowed on client credentials?

driesvints:
I'm actually not 100% sure why this is blocked. I'll leave this open for review by Taylor.

Sephster:
For what it's worth, `*` is a valid character for scopes according to the OAuth spec. I checked just in case it was being incorrectly blocked by by the server but note that we don't even have restrictions on the scopes. I'm going to raise an issue to add restrictions so we conform to the spec better.

Sammyjo20:
Hey guys, I understand this is a merged PR but I wanted to bring this back up again because I'm currently trying to use the `authorization_code` grant and realised that I'm not allowed to pass in all scopes. @driesvints, is there a reason why the flows like "authorization_code" are blocked? If there's a specific reason, I'd understand, but I suggest we should remove it entirely, maybe as an enhancement let the clients be the ones to restrict access to scopes.

driesvints:
I have no memory of this. It's been more than 2 years.

Sammyjo20:
Apologies, I have a discussion open here https://github.com/laravel/framework/discussions/40365

PR #949, opened 2019-01-27
[7.0] Allow '*' scope to be used with Client Credentials

philcross:
As mentioned in https://github.com/laravel/passport/issues/516, using the `*` character to request all scopes for the Client Credentials grant type is valid.

This PR will allow the character to be used when requesting an access token with the Client Credentials Grant.

### DIFF
--- a/src/Bridge/ScopeRepository.php
+++ b/src/Bridge/ScopeRepository.php
@@ -25,7 +25,7 @@ class ScopeRepository implements ScopeRepositoryInterface
         array $scopes, $grantType,
         ClientEntityInterface $clientEntity, $userIdentifier = null)
     {
-        if (! in_array($grantType, ['password', 'personal_access'])) {
+        if (! in_array($grantType, ['password', 'personal_access', 'client_credentials'])) {
             $scopes = collect($scopes)->reject(function ($scope) {
                 return trim($scope->getIdentifier()) === '*';
             })->values()->all();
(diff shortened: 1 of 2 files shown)


billriess:
Isn't `*` sort of implied by the nature of how `client_credentials` work in the current version of passport? Any client can request any scope and be given it automatically. I guess this just furthers that some but there is no back-end check to validate the client and it's allowed scopes without adding your own logic.

philcross:
As far as I can tell from the documentation, providing `'scopes' => ['*']` is meant to allow the access token all registered scopes.

For other grant types, like password grant, it's correctly stored in the `oauth_access_tokens` table as `["*"]`, but if you try to pass `*` in when using client credentials, it's stored as `[]` and therefore no scopes are given to the access token.

There's nothing in the documentation that I can see that indicates client credentials _shouldn't_ be allowed `*`, and there was another test (https://github.com/laravel/passport/blob/7.0/tests/CheckClientCredentialsForAnyScopeTest.php#L26) which gave me the impression `*` is valid for `CC` (along with apparently `*` being a valid scope character in oauth specs (mentioned in https://github.com/laravel/passport/issues/516))

billriess:
@philcross Under https://laravel.com/docs/5.7/passport#password-grant-tokens under "Requesting All Scopes" it does say `This scope may only be assigned to a token that is issued using the password grant:`

philcross:
Ha, I completely missed that. I'm not sure _why_ it's restricted though, especially not to Client credentials - I'll leave the PR as it is, if it's refused then fair enough, but feels a bit weird to have to manually specify all scopes for client credentials, which I would assume is a more trusted grant type than Password Grant

billriess:
I agree it should be allowed on this scope and I think the PR is good. I just think there are some fundamental issues with Passport in general that should also be addressed. `client_credentials` should have a whitelist of scopes. `client_id` should not be sequential... and so on.

PR #954, opened 2019-02-07
Make user_id option a little more useful

warrickbayman:
Currently the `user_id` option is only used when creating authorization code clients, otherwise it's `null`. This will allow the `user_id` option to be used when creating personal access, password grant and client credentials grant clients from the cli.


driesvints:
This isn't wanted because you're not creating clients in a user context here. Especially not for the client credentials client which doesn't belongs to any user.

warrickbayman:
Okay thanks. What about password grants, though?

driesvints:
Also not wanted. Only AuthCode clients are created directly by users.

PR #965, opened 2019-02-20
 Support overriding logic of automatic token approval 

TimWolla:
My use case for this is storing the information which clients have been approved by the user explicitly instead of implicitly by the existence of a token that expires automatically if the application fails to renew the token until its expiry.

This change allows me to re-use the bulk logic of `AuthorizationController::authorize()` without copying and pasting it into my own controller.

I'm not sure about the correct target branch. By moving the injection of the repositories into the controller the method signature of `authorize` changed which breaks compatibility with users that extend `AuthorizationController`. Please advise.


taylorotwell:

PR #974, opened 2019-02-26
[7.0] Add matrix to travis to tests against all supported version.

crynobone:


driesvints:
Hmm, interesting. I was thinking of doing this but thought it might be a little overkill 😅

.travis.yml
```diff
@@ -1,35 +1,39 @@
...
+  - 7.3
+
 env:
   global:
     - SETUP=stable
+  matrix:
+    - laravel=5.6.*
+    - laravel=5.7.*
+    - laravel=5.8.*
+    - laravel=dev-master
```
driesvints: Please uppercase env variables.

crynobone:
Travis will be slower on handling each build since now it's 24 builds per commit instead of 6 builds but if we really want to avoid introducing anything that might not be available in previous releases or introduce any regression bug this is the only way to ensure that.

driesvints:
Thanks for sending this in btw 👍

.travis.yml
```diff
@@ -1,35 +1,39 @@
...
-      env: SETUP=nightly
-    - php: 7.2
-      env: SETUP=nightly
-    - php: 7.3
-      env: SETUP=nightly
+    - env: laravel=dev-master
+    - env: SETUP=nightly laravel=5.6.*
+    - env: SETUP=nightly laravel=5.7.*
+    - env: SETUP=nightly laravel=5.8.*
+    - env: SETUP=nightly laravel=dev-master
```
crynobone: We can reduce the number of `laravel=dev-master` build if we use `matrix.include` instead of `env.matrix`, let me know what you think is acceptable to test `dev-master`.

driesvints: Yeah we'll need to prevent testing unnecessary version combinations. PHP 7.1 + dev-master for example.

.travis.yml
```diff
@@ -1,35 +1,39 @@
...
 
 sudo: false
 
 before_install:
   - phpenv config-rm xdebug.ini || true
 
 install:
+  - composer require "laravel/framework=${laravel}" --prefer-dist --no-interaction --no-suggest
```
driesvints: Hmm. I'm not sure this is the wtg (see failing tests)

driesvints:
I'll take another look at this tomorrow.

.travis.yml
crynobone: Fixed.

.travis.yml
crynobone: Updated, pending travis to run the build.

PR #975, opened 2019-02-28
[7.0] Change wasRecentlyCreated to false in actingAs

FatBoyXPC:
The core framework fixed this same bug with the core actingAs method but it seems it was missed for Passport! Check this PR for more details:

https://github.com/laravel/framework/pull/25873

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -389,6 +389,8 @@ class Passport
 
         $user->withAccessToken($token);
 
+        $user->wasRecentlyCreated = false;
+
         app('auth')->guard($guard)->setUser($user);
 
         app('auth')->shouldUse($guard);


driesvints:
Don't we need the same code with the if statement? Not sure here, genuine question.

taylorotwell:
Also unsure why you tried to port the feature but changed the code.

FatBoyXPC:
@driesvints I don't think we need the conditional, and certainly not the truthy check of it on the right side. Setting something to false that's already false won't have any negative effects. The `User` that's passed to the auth guard is expecting an `Authenticatable` which also extends `Model`, which should always have `wasRecentlyCreated`.

I look at this as "Hey look we need this to be false, why do we care if it's true first?"

@taylorotwell That also explains why I changed the code - I only removed the conditional, but the important bit is still there.

driesvints:
@FatBoyXPC can you resubmit with this explanation?

PR #976, opened 2019-02-28
AccessTokenRevoked Event

centrus:
Dispatch event when access token is manualy revoked.


centrus:
I wrote fixes and teseted on Travis CI and passed, which I should do the first time.

PR #977, opened 2019-02-28
AccessTokenRevoked Event 

centrus:
I want to propose dispatching event when access token is manually revoked on authorization server, so it can be implemented listener to alert resource servers that token is not valid any more


driesvints:
This seems to break some tests. Please try to resubmit with a passing test and perhaps also a real-app example so it's clear why this is needed.

PR #979, opened 2019-02-28
[7.0] Change wasRecentlyCreated to false

FatBoyXPC:
Resubmitting with explanation as requested by @driesvints in #975 

The core framework fixed this same bug with the core actingAs method but
it seems it was missed for Passport! Check this PR for more details:
https://github.com/laravel/framework/pull/25873

I don't think we need the conditional, and certainly not the truthy check of it on the right side. Setting something to false that's already false won't have any negative effects. The `User` that's passed to the auth guard is expecting an `Authenticatable` which also extends `Model`, which should always have `wasRecentlyCreated`.

I look at this as "Hey look we need this to be false, why do we care if it's true first?"

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -389,6 +389,10 @@ class Passport
 
         $user->withAccessToken($token);
 
+        if (isset($user->wasRecentlyCreated) && $user->wasRecentlyCreated) {
+            $user->wasRecentlyCreated = false;
+        }
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Please include the exact conditional check that is in the framework.

FatBoyXPC:
@taylorotwell Done! Did I miss something about how this conditional works or is this a consistency thing?

PR #986, opened 2019-03-11
[7.0] Tweak Travis setup

driesvints:
This PR removes the nightly builds as they don't provide too much value at this point. I've also not seen too much other OS projects that do this on a dependency level (besides the nightly language builds themselves which we could do or not).

Also tweaked the script so it's not much more clear and concise. If this is accepted then I'll start adding these changes to all the other libraries as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: php
 
+sudo: false
+
 php:
+  - 7.1
   - 7.2
   - 7.3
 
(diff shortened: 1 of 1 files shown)


driesvints:
@crynobone Feel free to leave your remarks if you have any!

crynobone:
> I've also not seen too much other OS projects that do this on a dependency level

I do this all the times, It helps me to detect possible breaking change in Laravel patch version in the past. But totally depends on the project maintainer as what they see fit.

However, this changes bring a different installation setup when running locally to say send a PR since it doesn't implements `"prefer-stable": true` in `composer.json` which would leads to different version of deps between local and Travis-CI.

crynobone:
With `"prefer-stable": true`, you can clean up Travis configuration file even further and able to utilise `composer outdated` command built-in with Composer.

Screen Shot 2019-03-12 at 8 02 09 AM

This feature only work well with `"prefer-stable": true` or `"minimum-stability": "stable"`

driesvints:
@crynobone Thanks Mior! Makes total sense. I've updated the Travis and composer file with this.

PR #989, opened 2019-03-13
update zendframework/zend-diactoros version

Christophvh:
Update version of zend-diactoros to support sentry-laravel.


driesvints:
@Christophvh this will prevent installs of ~1.0 for current users. I'll try to send in a commit that allows installs of both.

Christophvh:
@driesvints Ah yes ofc, my mistake, I would've changed it, but I see you already made the commit. Thanks for the fast response. I assume this does not warrant a minor tag-release?

driesvints:
@Christophvh new release will be tagged next week tuesday.

PR #1001, opened 2019-04-12
[7.0] Fix incorrect comment in TokenGuard::getTokenViaCookie

matt-allan:
Just changes the comment from saying the request **is** from a valid source when the CSRF token doesn't match to saying that it **is not** from a valid source.


driesvints:
Build fails because of this: https://github.com/thephpleague/oauth2-server/pull/1006

I'll push a commit which will fix this.

driesvints:
@matt-allan pushed. Can you rebase?

matt-allan:
Thanks @driesvints, I just rebased.

PR #1013, opened 2019-05-09
[7.0] Enhanced Client User relationship

T0miii:
Enhanced Client User relationship, by adding foreignKey name so it still works if the Column name is not default "ID".
I think this might be important in some legacy systems ,where its not possible to change the column names that easily.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -49,7 +49,8 @@ class Client extends Model
     public function user()
     {
         return $this->belongsTo(
-            config('auth.providers.'.config('auth.guards.api.provider').'.model')
+            config('auth.providers.'.config('auth.guards.api.provider').'.model'),
+            resolve(config('auth.providers.'.config('auth.guards.api.provider').'.model'))->getKeyName()
         );
     }
 


src/Client.php
```diff
@@ -49,7 +49,8 @@ class Client extends Model
     public function user()
     {
         return $this->belongsTo(
-            config('auth.providers.'.config('auth.guards.api.provider').'.model')
+            config('auth.providers.'.config('auth.guards.api.provider').'.model'),
+            resolve(config('auth.providers.'.config('auth.guards.api.provider').'.model'))->getKeyName()
```
driesvints: A model doesn't needs to be resolved from the container so let's not use the resolve method here but do something like this instead:

```suggestion
           (new config('auth.providers.'.config('auth.guards.api.provider').'.model'))->getKeyName()
```

T0miii: I can't do `new` on a `config()`helper one solution could be to save the config() in a variable and then make it like this:

````
    public function user()
    {
        $class = config('auth.providers.' . config('auth.guards.api.provider') . '.model');

        return $this->belongsTo($class, (new $class)->getKeyName());
    }
````  

or i could store it in a member variable like in the Passport.php class.

````
    protected $userModel = config('auth.providers.' . config('auth.guards.api.provider') . '.model');
````

driesvints: Ah yeah, sorry, my bad 🤦‍♂ 

The first option is the better solution I think.

driesvints:
Actually: I just realized that the solution is to override the model yourself and do whatever you want in your override model. See the docs on that here: https://laravel.com/docs/5.8/passport#overriding-default-models

T0miii:
yess, thats what i actually did, just wasnt sure if thats a reason enough to create a Custom Client model to override.

But anyhow thanks for the support.

driesvints:
@T0miii the reason why the overrides were created was to accommodate for almost every use-case people had for customization so this is the way to go. Thanks for reporting though. Good luck with your project 👍

PR #1016, opened 2019-05-15
Create Token Revoked event

GabrielDeveloper:
This PR implement a token revoke event , so we can implement listeners to alert when token is revoked


taylorotwell:
This event will fire every time a revoked token is updated in the database.

GabrielDeveloper:
@taylorotwell what about dispatch the event into revoked method in token model?

PR #1018, opened 2019-05-17
[7.0] Created access token revoked events

GabrielDeveloper:
When the access token is revoked on AuthorizedAccessTokenController an event is dispatched to every listener of this event. 

* Correcting the PR #977

### DIFF
--- a/src/Http/Controllers/AuthorizedAccessTokenController.php
+++ b/src/Http/Controllers/AuthorizedAccessTokenController.php
@@ -4,12 +4,21 @@ namespace Laravel\Passport\Http\Controllers;
 
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
+use Illuminate\Events\Dispatcher;
+use Laravel\Passport\Events\AccessTokenRevoked;
 use Laravel\Passport\TokenRepository;
 
 class AuthorizedAccessTokenController
 {
     /**
      * The token repository implementation.
+     *
+     * @var use Illuminate\Events\Dispatcher;
+     */
+    protected $events;
+
+    /**
+     * The events dispatcher
      *
      * @var \Laravel\Passport\TokenRepository
      */
@@ -21,9 +30,10 @@ class AuthorizedAccessTokenController
      * @param  \Laravel\Passport\TokenRepository  $tokenRepository
      * @return void
      */
-    public function __construct(TokenRepository $tokenRepository)
+    public function __construct(TokenRepository $tokenRepository, Dispatcher $events)
     {
         $this->tokenRepository = $tokenRepository;
+        $this->events = $events;
     }
 
     /**
@@ -59,6 +69,13 @@ class AuthorizedAccessTokenController
         }
 
         $token->revoke();
+
+        $this->events->dispatch(new AccessTokenRevoked(
+            $token->id,
+            $token->user_id,
+            $token->client_id
+        )
+    );
 
         return new Response('', Response::HTTP_NO_CONTENT);
     }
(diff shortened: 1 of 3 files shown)


driesvints:
Can you add a test for this and explain how this would be used in a real-life example? Thanks

src/Http/Controllers/AuthorizedAccessTokenController.php
```diff
@@ -60,6 +70,13 @@ public function destroy(Request $request, $tokenId)
 
         $token->revoke();
 
+        $this->events->dispatch(new AccessTokenRevoked(
+            $token->id,
+            $token->user_id,
+            $token->client_id
+        )
+    );
```
driesvints: Please indent this correctly.

src/Http/Controllers/AuthorizedAccessTokenController.php
```diff
@@ -4,13 +4,22 @@
...
+use Illuminate\Events\Dispatcher;
 use Laravel\Passport\TokenRepository;
+use Laravel\Passport\Events\AccessTokenRevoked;
 
 class AuthorizedAccessTokenController
 {
     /**
      * The token repository implementation.
      *
+     * @var use Illuminate\Events\Dispatcher;
```
driesvints:
```suggestion
     * @var \Illuminate\Events\Dispatcher
```

tests/AuthorizedAccessTokenControllerTest.php
```diff
@@ -18,6 +19,11 @@ class AuthorizedAccessTokenControllerTest extends TestCase
      */
     protected $tokenRepository;
 
+    /**
+     * @var \Mockery\Mock|Illuminate\Events\Dispatcher
```
driesvints:
```suggestion
     * @var \Mockery\Mock|\Illuminate\Events\Dispatcher
```

tests/AuthorizedAccessTokenControllerTest.php
```diff
@@ -73,6 +83,8 @@ public function test_tokens_can_be_deleted()
     {
         $request = Request::create('/', 'GET');
 
+        $this->events->shouldReceive('dispatch')->once();
```
driesvints: It's probably better to assert the actual `AccessTokenRevoked` was dispatched.

GabrielDeveloper:
Of course, I have a system that use the [laravel-users-online](https://github.com/highideas/laravel-users-online) package and I listen the AccessTokenCreate to put the user on users online list and with the new event (AcessTokenRevoked) I will listen him and pull off of users online list.

tests/AuthorizedAccessTokenControllerTest.php
GabrielDeveloper: Ready, I changed to use the Dispatcher contratc.

tests/AuthorizedAccessTokenControllerTest.php
GabrielDeveloper: Ready.

src/Http/Controllers/AuthorizedAccessTokenController.php
GabrielDeveloper: Ready, I changed to use the Dispatcher contratc.

src/Http/Controllers/AuthorizedAccessTokenController.php
GabrielDeveloper: Ready.

taylorotwell:
To be honest this doesn't feel like a good use case for that. A token may not be revoked for weeks even though the user is no longer using the application.

Issue #1010, opened 2019-05-08
Allow first party apps to skip the consent dialog


matt-allan:
## Description

If I am running two apps, `app.my-app.test` and `admin.my-app.test`, both of them are first party applications, and I am using the auth code grant or implicit grant to authorize `admin.my-app.test` with the Passport server running on `app.my-app.test`, the user will currently see a consent dialog like this:

![authorization-request](https://user-images.githubusercontent.com/9440455/57378897-b39b7c00-7173-11e9-8d24-60afeab969ab.png)

Since both of these applications are first party apps I would like to be able to skip the consent dialog.

A good example of this is Google.  If you go to `mail.google.com` and are not logged in, you will be redirected to `accounts.google.com`.  You will not see a consent screen but instead will continue to the normal login screen.

Another example of this is Auth0 - they [let you skip the consent dialog](https://auth0.com/docs/api-auth/user-consent) for first party apps.

Otka [lets you specify what scopes should require consent](https://developer.okta.com/docs/how-to/add-user-consent-to-flow/#how-to-require-user-consent).

**edit:** Looks like [doorkeeper in ruby](https://doorkeeper.gitbook.io/guides/configuration/skip-authorization), [Django OAuth Toolkit](https://django-oauth-toolkit.readthedocs.io/en/latest/advanced_topics.html#skip-authorization-completely-for-trusted-applications), and [IdentityServer in .NET](https://identityserver4.readthedocs.io/en/latest/topics/consent.html) support this too.

The OAuth 2.0 IETF doc [allows this explicitly](https://tools.ietf.org/html/rfc6749#section-4.1.1):

> If the request is valid,
   the authorization server authenticates the resource owner and obtains
   an authorization decision (by asking the resource owner **or by
   establishing approval via other means**).

## Rationale

Developers (understandably) don't want to show a consent screen for their own apps.  I've seen developers reach for the password grant instead because there is no consent screen involved.

The password grant doesn't allow you to prompt user for multi factor authentication, use social login, etc, doesn't support single sign on, and [trains users to get phished](https://www.scottbrady91.com/OAuth/Why-the-Resource-Owner-Password-Credentials-Grant-Type-is-not-Authentication-nor-Suitable-for-Modern-Applications).  It's less secure and shouldn't really be used.

Skipping the consent screen is already supported by other major OAuth providers.  It doesn't create a security risk as long as it's limited to first party clients.

[...]

PR #1022, opened 2019-05-21
[7.0] Allow first party clients to skip the authorization prompt

matt-allan:
## Overview

This PR adds a new method `Client::skipsAuthorization()`. If the method returns true the client is allowed to skip the authorization prompt.

## Rationale

It's considered best practice to use the Authorization code grant for first party mobile apps and SPAs but Passport currently requires you to show an authorization prompt to the user ("{client} is requesting permission...") with this grant type.

It doesn't really make sense to show the authorization prompt for first party clients.  If Google asked you to grant permission to Google every time you logged in to gmail it would be pretty confusing.

Allowing first party clients to skip the authorization prompt is supported by the majority of OAuth servers: Auth0, Otka, Doorkeeper, Django OAuth Toolkit, IdentityServer, Keycloak, Ory Hydra, and others.  It's also explicitly allowed by the OAuth specification.

It's difficult to add this feature yourself.  You have to override the Authorization controller which requires copying ~30 lines of code.  You then need to register the route override in the `AuthServiceProvider` after the `Passport::routes()` call and make sure to add the same middleware.

There is some more background info in #1010.

## Usage

To use this feature you need to extend the `Client` and override the `skipsAuthorization` method. You also need to tell Passport to use your extended model in the `AuthServiceProvider`.

How you implement `skipsAuthorization` is up to you. For example, you may add a `first_party` boolean column to the database and check the value.  Alternatively you might want to check the redirect URL and allow any client for a given domain.

```php
use App\Models\Passport\Client;
use App\Models\Passport\Token;
use App\Models\Passport\AuthCode;
use App\Models\Passport\PersonalAccessClient;

/**
 * Register any authentication / authorization services.
 *
 * @return void
 */
public function boot()
{
    $this->registerPolicies();

    Passport::routes();

    Passport::useClientModel(Client::class);
}
```

```php
namespace App\Passport;

use Laravel\Passport\Client as BaseClient;

class Client extends BaseClient
{
    /**
     * Determine if the client should skip the authorization prompt.
     *
     * @return bool
     */
    public function skipsAuthorization()
    {
        return $this->first_party;
    }
}
```

## Backwards Compatibility

This feature is opt-in and does not do anything unless you override the `skipsAuthorization` method to return true.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -82,4 +82,14 @@ class Client extends Model
     {
         return $this->personal_access_client || $this->password_client;
     }
+
+    /**
+     * Determine if the client should skip the authorization prompt.
+     *
+     * @return bool
+     */
+    public function skipsAuthorization()
+    {
+        return false;
+    }
 }
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -67,7 +67,8 @@ class AuthorizationController
                 $client = $clients->find($authRequest->getClient()->getIdentifier())
             );
 
-            if ($token && $token->scopes === collect($scopes)->pluck('id')->all()) {
+            if (($token && $token->scopes === collect($scopes)->pluck('id')->all()) ||
+                $client->skipsAuthorization()) {
                 return $this->approveRequest($authRequest, $user);
             }
 
--- a/tests/AuthorizationControllerTest.php
+++ b/tests/AuthorizationControllerTest.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport\Tests;
 
 use Exception;
+use Laravel\Passport\Client;
 use Mockery as m;
 use Laravel\Passport\Token;
 use Illuminate\Http\Request;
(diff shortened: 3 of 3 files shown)


driesvints:
Instead of doing this in the settings which might be a little more overhead, why don't we do this in the model itself?

```php
public function skipsAuthorization()
{
    return false;
}
```

And then:

```php
if (($token && $token->scopes === collect($scopes)->pluck('id')->all()) ||
    $client-> skipsAuthorization()) {
    return $this->approveRequest($authRequest, $user);
}
```

This way it's done on a model base without an additional setting. If people want to override they can override the model and method and still build in custom support for anything specific if they want.

matt-allan:
👍 sounds good to me, I will push an update shortly.

matt-allan:
@driesvints I implemented the change, updated the PR description, and rebased.

RomainLanz:
Hey all, 👋 

Are you going to provide a PR for the documentation too @matt-allan?

matt-allan:
Thanks for the reminder @RomainLanz, I forgot. I just opened a PR: https://github.com/laravel/docs/pull/5226

PR #1029, opened 2019-05-31
Added the ability for finer route definition

KieronWiltshire:
Now it is possible to pass a `\Illuminate\Contracts\Routing\Registrar` instance through to each method which allows you to group middleware as you see fit. Below is an example of how passing the registrar is useful.

```php
Passport::routes(function ($registrar) {
    $registrar->router()->group(['middleware' => 'role:developer'], function ($router) use ($registrar) {
        $registrar->forClients($router); // pass the sub-router
    });
});
```

You can additionally pass through options which can include middleware to each route which is done as shown below:

```php
Passport::routes(function ($registrar) {
    $registrar->forClients(null, [
        'passport.clients.store' => [
            'middleware' => 'role:developer'
        ]
    ]);
// or
    $registrar->router()->group(['middleware' => 'role:developer'], function ($router) use ($registrar) {
        $registrar->forClients($router, [
            'passport.clients.store' => [
                'middleware' => 'role:even-bigger-developer'
            ]
        ]);
    });
});
```

This added functionality doesn't break any of the current functionality as the new functionality assumes the old methods if the specified parameters are not passed.

### DIFF
--- a/src/RouteRegistrar.php
+++ b/src/RouteRegistrar.php
@@ -51,6 +51,10 @@ class RouteRegistrar
                         $options['middleware'] = explode('|', $options['middleware']);
                     }
 
+                    if (!is_array($subOptions['middleware'])) {
+                        $subOptions['middleware'] = explode('|', $subOptions['middleware']);
+                    }
+
                     $options['middleware'] = array_merge($options['middleware'], $subOptions['middleware']);
 
                     unset($subOptions['middleware']);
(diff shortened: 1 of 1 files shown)


taylorotwell:
Now I see it would have been simpler to just mention you should copy the routes from the service provider into your own application 😬

KieronWiltshire:
whoops :P do you not think it's kind of counter-intuitive to redefine the routes?

PR #1034, opened 2019-07-01
[7.0] Change server Property type in CheckClientCredentialForAnyScope

alshf89:
I change declaration $server property in CheckClientCredentialForAnyScope from __private__ to __protected__, this is useful if anyone wants to extends this middleware.

### DIFF
--- a/src/Http/Middleware/CheckClientCredentialsForAnyScope.php
+++ b/src/Http/Middleware/CheckClientCredentialsForAnyScope.php
@@ -16,7 +16,7 @@ class CheckClientCredentialsForAnyScope
      *
      * @var \League\OAuth2\Server\ResourceServer
      */
-    private $server;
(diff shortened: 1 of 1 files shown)


driesvints:
Unfortunately this is a breaking change if anyone is overwriting the property: https://3v4l.org/i7s5h

alshf89:
I think if someone wants to overwrite the property they must use another name and use dependency injection and service container to assign `League\OAuth2\Server\ResourceServer` object to that property.

and why should they overwrite it when they can access it from parent class ?

Actually I wanted to extends my custom middleware with `CheckClientCredentialForAnyScope` and needed `server` Property to do something in my extended middleware. but I cant cuz it is private. 

currently I can extend CheckClientCredential  middleware because it has `protected` server property.

PR #1038, opened 2019-07-03
[7.0] Fix exception will thrown if token belongs to first party clients

### LINKED ISSUE #691: Client credentials are not secure for documented use case

alshf89:
I made a pull request for issue [#691](https://github.com/laravel/passport/issues/691).

So when someone passes a token in their request which belongs to first party clients (such as `Password Credentials`), our middlewares will throw an `Illuminate\Auth\AuthenticationException` Exception.

So only those who have issued access token with `Client Credentials` grant type can go through `CheckClientCredential` & `CheckClientCredentialForAnyScope` Middlewares.

### DIFF
--- a/src/Http/Middleware/CheckClientCredentials.php
+++ b/src/Http/Middleware/CheckClientCredentials.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport\Http\Middleware;
 
 use Closure;
+use Laravel\Passport\TokenRepository;
 use League\OAuth2\Server\ResourceServer;
 use Illuminate\Auth\AuthenticationException;
 use Laravel\Passport\Exceptions\MissingScopeException;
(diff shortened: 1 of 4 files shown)


themsaid:
This is a breaking change, please submit it to the master branch.

alshf89:
Ok i will do it, tnx for your response

PR #1039, opened 2019-07-04
[7.0] Merge default Passport configuration

hackel:
The PassportServiceProvider is lacking a `mergeConfigFrom` call, which breaks the key check in `makeCryptKey` unless one first copies the default `config/passport.php` into their app's config directory.  Copying the config should only be required when one wants to override the default values.  I ran into this issue when trying to set my keys [using environment variables](https://laravel.com/docs/5.8/passport#deploying-passport) in a new project.

### DIFF
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -78,6 +78,8 @@ class PassportServiceProvider extends ServiceProvider
      */
     public function register()
     {
+        $this->mergeConfigFrom(__DIR__.'/../config/passport.php', 'passport');
+
         $this->registerAuthorizationServer();
 
         $this->registerResourceServer();


taylorotwell:
I think we would only do this if the configuration is not cached?

drbyte:
> I think we would only do this if the configuration is not cached?

Agreed! 

But it seems strange that you're suggesting Laravel doesn't already do that by default.

If it's that important then I'm surprised that the [docs](https://laravel.com/docs/5.8/packages#configuration) for mergeConfigFrom don't mention any recommendation to put the burden on packages to first ask the framework to do a conditional check for whether config caching is enabled and only if it is not then do the merge.

Is this a documentation problem? or a framework problem?

(I'm indifferent to this PR, but as a package maintainer I'm very interested in the comment you made, as it suggests something big is missing somewhere.)

taylorotwell:
Right now the framework does not check if the configuration is cached. If you tell it to merge config it will merge config. So, you will need to check that in this PR. Whether that should be the behavior or not is probably a separate discussion.

PR #1040, opened 2019-07-06
Fix exception will thrown if token belongs to first party clients

alshf89:
I made a pull request for issue [#691](https://github.com/laravel/passport/issues/691).

So when someone passes a token in their request which belongs to first party clients (such as `Password Credentials`), our middlewares will throw an `Illuminate\Auth\AuthenticationException` Exception.

So only those who have issued access token with `Client Credentials` grant type can go through `CheckClientCredential` & `CheckClientCredentialForAnyScope` Middlewares.

### DIFF
--- a/src/Http/Middleware/CheckClientCredentials.php
+++ b/src/Http/Middleware/CheckClientCredentials.php
@@ -5,6 +5,7 @@ namespace Laravel\Passport\Http\Middleware;
 use Closure;
 use Zend\Diactoros\StreamFactory;
 use Zend\Diactoros\ResponseFactory;
+use Laravel\Passport\TokenRepository;
 use Zend\Diactoros\UploadedFileFactory;
 use League\OAuth2\Server\ResourceServer;
 use Zend\Diactoros\ServerRequestFactory;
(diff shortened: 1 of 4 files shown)


alvirbismonte:
Hi @JuanDMeGon ,

I believe the reason behind this change is because they separated the authentication between the user credentials(password grant type) and client credentials (client grant type). As mentioned from [API Authentication (Passport)](https://laravel.com/docs/5.7/passport#client-credentials-grant-tokens) which says that 

> client credentials grant is suitable for machine-to-machine authentication

JuanDMeGon:
Thanks. I think it is not correct. The purpose of the client credentials grant is to validate the access to the API from valid clients when there is not User data involved.
Basically, it is what stands in OAuth2 official site: 
"The Client Credentials grant is used when applications request an access token to access their own resources, not on behalf of a user." [Source](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)
I opened an issue (#1125) to discuss this, maybe you would like to comment on your opinion on this.
Best wishes :)

PR #1044, opened 2019-07-09
[8.0] Fix auth codes table customization

afilippov1985:
Current implementation does not respect `$connection` property in custom AuthCode model and always use Laravel's default connection.


driesvints:
Bit torn on this on wether we should send this to the current release branch because technically it's a bug fix. But I guess the breaking changes from this PR warrent that it's better to do the in the next release. Thanks for sending this in 👍

PR #1049, opened 2019-07-17
[7.0] Make passport:keys only generate keys when needed.

MrMage:
Benefit: This can significantly speed up tests that call `artisan passport:install` by avoiding work whose results will be discarded anyway (see https://stackoverflow.com/a/50243105/153354 for an example where this is recommended).

No changes in observable behavior, we only avoid unnecessary work.

I guess this should also be forward-ported to 8.0 once merged, or will 7.0 automatically be merged into 8.0?


driesvints:
@MrMage yes it will be.

PR #1052, opened 2019-07-22
[5.8] Support public clients

matt-allan:
This pull request adds support for public clients.

The league/oauth2-server now supports using both public and confidential clients with the authorization code grant. The recommended flow for single page applications and native apps is the auth code grant with a public client and PKCE. Supporting public clients makes it possible for developers to use the authorization code grant with PKCE instead of the insecure implicit grant.

Previously all clients were required to have a secret. If you used the implicit grant the secret just wasn't checked. We can't just skip checking the secret for the auth code grant like we were doing with the implicit grant because we need to check the secret if it was issued.

Instead we need to offer the ability to create public clients, and only check the credentials if the client is confidential.

The UI looks like this:

Screenshot from 2019-07-17 16-28-13

The `oauth_clients.secret` column was updated to be nullable and the `POST /clients` endpoint was updated to allow passing a `confidential` param (defaults to true).

A public client can only be used with the implicit grant and the authorization code grant. The client credentials and personal access grants are only usable with a confidential client.

The password grant is supposed to be usable by public clients and this used to be possible but it's currently not working in league/oauth2-server, see https://github.com/thephpleague/oauth2-server/issues/889#issuecomment-512554693.g

If you attempt to use the auth code grant with a client was assigned a secret and don't provide the secret it will fail, because the League server [calls `ClientRepository::validateClient` if `Client::isConfidential` returns `true`](https://github.com/thephpleague/oauth2-server/blob/e1dc4d708c56fcfa205be4bb1862b6d525b4baac/src/Grant/AuthCodeGrant.php#L103).

### DIFF
--- a/database/migrations/2019_07_17_000001_make_client_secret_nullable.php
+++ b/database/migrations/2019_07_17_000001_make_client_secret_nullable.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class MakeClientSecretNullable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('oauth_clients', function (Blueprint $table) {
+            $table->string('secret', 100)->nullable()->change();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('oauth_clients', function (Blueprint $table) {
+            $table->string('secret', 100)->change();
+        });
+    }
+}
--- a/src/Bridge/ClientRepository.php
+++ b/src/Bridge/ClientRepository.php
@@ -37,7 +37,7 @@ class ClientRepository implements ClientRepositoryInterface
         }
 
         return new Client(
-            $clientIdentifier, $record->name, $record->redirect, ! is_null($record->secret)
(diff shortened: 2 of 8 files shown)


database/migrations/2019_07_17_000001_make_client_secret_nullable.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class MakeClientSecretNullable extends Migration
```
matt-allan: The migration isn't 100% necessary as we could just use the empty string for public clients but it seems more explicit to use `NULL`.

src/Bridge/ClientRepository.php
```diff
@@ -75,8 +75,6 @@ protected function handlesGrant($record, $grantType)
                 return $record->personal_access_client;
             case 'password':
                 return $record->password_client;
-            case 'client_credentials':
```
matt-allan: This check was moved up to `validateClient` as the League server should only call `validateClient` for confidential clients, regardless of the grant type.

matt-allan:
I'm going to close this for now, I need to clarify some behavior with the League server first: https://github.com/thephpleague/oauth2-server/issues/1034

AlfaSchz:
Hello @matt-allan,
Are you still working on the public client for Passport? I am interested in a Oauth2 Authentication Code + PKCE solution and being able to use Passport for it would be great : )
Thanks for the awesome work!

driesvints:
@AlfaSchz Public Clients landed in Passport 8.0 thanks to @matt-allan

PR #1054, opened 2019-08-02
[7.0] Add copy token button to Access Token Modal

thamerbelfkihthamer:
The current implementation has a copy button for the **Access Token Modal** within **PersonalAccessTokens.vue** file, we will be able to copy the long token string quickly instead of selecting it.


driesvints:
Can you add a screenshot how this would look?

thamerbelfkihthamer:
@driesvints  look at it please
token

taylorotwell:
That alert has bad grammar and doesn't look good.

PR #1060, opened 2019-08-07
Update RefreshTokenRepository.php

narwy:
Return the bool instead of a `$refreshToken->revoked`

This method supports multiple databases, including document db structures such as MongoDB.

Otherwise it will throw this error:
```
Trying to get property 'revoked' of non-object at /var/www/vendor/laravel/passport/src/Bridge/RefreshTokenRepository.php:80)
```


taylorotwell:
Tests fail and we already don't access ->revoked if the object is null so I'm not sure what the problem is?

narwy:
@taylorotwell It's 100% not working with fresh MongoDB install from `jenssegers/laravel-mongodb` only on the token refresh api call. This seemed to fix it.

Issue #1062, opened 2019-08-09
master: HandlesOAuthErrors middleware unable to actually handle errors


TimWolla:
- Passport Version: dev-master with commit 7959c6c4d181090de5f064971a9fca035f759513
- Laravel Version: 5.8.30
- PHP Version: 7.3.4 (Debian Buster)
- Database Driver & Version: MySQL 8.0.17

### Description:

The new `HandlesOAuthErrors` middleware is unable to actually handle errors, because Laravel intercepts `Exception`s of the final destination of the middleware pipeline:

https://github.com/laravel/framework/blob/9097786caae30e9ca243874e5c807609f1f2728e/src/Illuminate/Pipeline/Pipeline.php#L133

see also my comment in the PR introducing the middleware: https://github.com/laravel/passport/pull/937#issuecomment-519923621

### Steps To Reproduce:

Attempt to use Passport within a Laravel application and open `/oauth/authorize` with an invalid `client_id`. You'll see an HTTP 500 instead of a proper JSON response explaining the issue.

driesvints:
The PR for this was merged into master.

donmbelembe:
I'm still getting a generic error on passport 8.0

PR #1066, opened 2019-08-11
[8.0] Use a renderable exception for OAuth errors

matt-allan:
This PR updates the Passport controllers to use a renderable exception instead of middleware for rendering OAuth errors.

Originally Passport handled exception reporting and rendering itself, but developers wanted to be able to control the rendering so the behavior was moved to middleware in #937.

The problem with using middleware is the pipeline catches exceptions before they reach middleware (#1062).

The only reason Passport needed to handle exceptions in the first place was to convert the League exception's PSR response to something Laravel can render, so this PR wraps the League exception in a renderable Passport exception.

If the user wants to change how the exception is rendered they can do that in their application's exception handler.

I don't see any reason that Passport needs to override 500 error rendering so these fall through to the application's normal rendering.

**Upgrading:**

If you are explicitly handling `League\OAuth2\Server\Exception\OAuthServerException` in your exception handler's `report` method you will need to check for an instance of `Laravel\Passport\Exceptions\OAuthServerException` instead.

If for some reason your OAuth client's are relying on 500 errors returning the string `Error.` in the response you will need to override the rendering for `/oauth/authorize` and `/oauth/token`, i.e.

```php
public function render($request, Exception $e)
{
	if ($request->is('/oauth/authorize', '/oauth/token')) {
		return response('Error.', 500);
	}
}
```


src/Exceptions/OAuthServerException.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        parent::__construct($e->getMessage(), $e->getCode(), $e);
+
+        $this->response = $response;
+    }
+
+    /**
+     * Render the exception into an HTTP response.
+     *
+     * @param  \Illuminate\Http\Request
```
driesvints:
```suggestion
     * @param  \Illuminate\Http\Request  $request
```

src/Http/Controllers/HandlesOAuthErrors.php
```diff
@@ -0,0 +1,33 @@
...
+
+trait HandlesOAuthErrors
+{
+    use ConvertsPsrResponses;
+
+    /**
+     * Perform the given callback with exception handling.
+     *
+     * @param  \Closure  $callback
+     *
```
driesvints:
```suggestion
```

tests/HandlesOAuthErrorsTest.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function testShouldHandleOAuthServerException()
+    {
+        $controller = new HandlesOAuthErrorsStubController;
+
+        $exception = new LeagueException('Error', 1, 'fatal');
+
+        $e = null;
+        try {
```
driesvints:
```suggestion

        try {
```

driesvints (changes requested):

driesvints:
Also rebase once to resolve the conflicts.

matt-allan:
Hey @driesvints, thanks for the review. I made the requested changes and rebased.

src/Exceptions/OAuthServerException.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Laravel\Passport\Exceptions;
+
+use Exception;
+use Illuminate\Http\Response;
+use League\OAuth2\Server\Exception\OAuthServerException as LeagueException;
+
+class OAuthServerException extends Exception
+{
+    protected $response;
```
driesvints: Add a DocBlock here.

src/Exceptions/OAuthServerException.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function __construct(LeagueException $e, Response $response)
```
driesvints: Add a DocBlock here like on other constructors.

driesvints (changes requested):
Heya, just some DocBlocks and we're good.

matt-allan:
@driesvints thanks, I added the docblocks and squashed.

src/Exceptions/OAuthServerException.php
driesvints: We use fully qualified namespaces in docblocks everywhere and two spaces before and after a type for `@param` annotations.

driesvints:
@matt-allan I'm going to re-open this if that's okay because we'll need this for the next release. Planning to work on Passport 8 in September so I'll take a look at this then.

matt-allan:
Sure, no problem.

swh-cgm:
@driesvints I am using passport 11.10.5. Can I still catch and render Passport's Laravel\Passport\Exceptions\OAuthServerException from Handler.php? 
I want to make a JSON response if the access token is expired. But I can only catch League\OAuth2\Server\Exception\OAuthServerException only on reportable function in Handler.php.

biodunbamigboye:
I faced this same issue today, and came out with a very simple though not syntactically elegant solution, but it works like magic
Note that my project is running on laravel 8 and might not be able to benefit from most recent fixes 

Inside Handler.php  add League\OAuth2\Server\Exception\OAuthServerException to $dontReport array then in render method, based on the message Log a custom error
`
    use League\OAuth2\Server\Exception\OAuthServerException;

    protected $dontReport = [
        OAuthServerException::class
    ];

    public function render($request, Throwable $exception)
     {
          if ($exception->getMessage() === 'Unauthenticated.') {
            Log::error('error message, it could be json encoded array, string, or any thing you desire'));
            }
     }
`

PR #1068, opened 2019-08-16
[7.0] Let Passport support inherited parent scopes.

lucasmichot:
This PR introduces the concept of inherited scopes.

By setting `Passport::$withInheritedScopes = true;` in your `AppServiceProvider.php` file you can enable this new feature (which is disabled by default).

Once enabled it allows the access to be granted or not depending on the presence of parent scopes.

This is a common pattern which is used by [Github](https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes) for their API scopes - but not only.

A user will have the possibility to *also* execute an action requiring the `admin:webhooks:create` scope, if he already has already been granted the following scopes `admin:webhooks`, `admin`.

---

Note:

If merged, additional documentation should be added to https://laravel.com/docs/5.8/passport.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -139,6 +139,13 @@ class Passport
     public static $unserializesCookies = false;
 
     /**
+     * Indicates the scope should inherit its parent scope.
+     *
+     * @var bool
+     */
+    public static $withInheritedScopes = false;
+
+    /**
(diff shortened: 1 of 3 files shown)


diego-lipinski-de-castro:
Docs on this?

Issue #982, opened 2019-03-04
Passport Multi-Auth (Implementation)


billriess:
Since https://github.com/laravel/passport/issues/161 has been locked we are no longer able to discuss the issue. Before I submit a PR I want to make sure we have a common ground on what the expected behavior is for multi-provider support in Passport. More specifically, should the provider be defined at the oauth client level? This would limit that client id/secret pair to always return from a specific provider. If the gate requested another provider than what the client has defined it would return a 401.

This is the functionality I have built into my current project and I would be more than willing to issue a PR for it but I want to make sure that is what everyone is expecting.

EDIT: I should note, that if no provider is set on the oauth client then it will follow the same logic as what is in place now, so it should be backward compatible outside of the migration.

driesvints:
I agree that this issue might be a good place to restart the discussion on how to implement this. I currently don't have time myself to look into this but feel free to provide thoughts here.

Warning: I'll be removing all comments which aren't related to looking for an actual implementation to prevent this issue from de-railing like the other one. Please only discuss the possible implementation for this.

billriess:
I think to get this started we should first determine if the multi-auth implementation should be enforced or not. Meaning, should the gate simply return the user against its own provider or should the client have a provider that enforces the user must come from a specific provider?

pramanikriju:
Or to make it more dynamic, the provider can be supplied in the request and the guard check can be made against that request parameter to make it compatible with as many models as required

afilippov1985:
Hello everyone.
Take a look at my proposal #987

rugiguru:
Still there is no update for implementing multi auth ??

billriess:
> Still there is no update for implementing multi auth ??

This discussion is meant to figure out the best approach to add multi-auth. There are a bunch of different ways this could be implemented all with different outcomes.

billriess:
Personally, I think each `oauth_client` should be limited to 1 provider but I can see how this may be limiting in some use cases. If we open it up to multiple providers per client then the developer will always be responsible for returning the correct user against any of the allowed providers.

pramanikriju:
> Personally, I think each `oauth_client` should be limited to 1 provider but I can see how this may be limiting in some use cases. If we open it up to multiple providers per client then the developer will always be responsible for returning the correct user against any of the allowed providers.

But isn't that the point of having multi-auth, to have the flexibility of choosing a provider but at the same time, using fallbacks for default behavior, like using the User model as is the case now. 

[...]

PR #1072, opened 2019-08-29
Initial support for multi-auth.

billriess:
This PR is to setup multi-auth support. To achieve this, the oauth_client has a new column called `provider` which is nullable. The `TokenGuard` now checks to see if a provider is defined for a client, if one is it will validate it against the requested provider, if those match it will continue through as normal. If not, it will return early as the request is invalid.

https://github.com/laravel/passport/issues/982


billriess:
This is my first PR to the Laravel community. I was unsure how to handle the migrations for this case. Would it be preferred to have a new migration to add the column or just update the existing one?

Issue #968, opened 2019-02-25
Can't use custom Token class that doesn't extend Laravel\Passport\Token


alasdairmackenzie:
My code is attempting to call `$user->createToken()` via the `Laravel\Passport\HasApiTokens\HasApiTokens` trait whilst using a custom Token class that extends a non Laravel model class.  However, because the `save()` function inside the `TokenRepository` expects it to be a ` Laravel\Passport\Token`, this is not possible.

```php
    L77: public function save(Token $token)
````

Instead I get an exception

```
"Argument 1 passed to Laravel\Passport\TokenRepository::save() must be an instance of Laravel\Passport\Token, instance of XXX\Mongodb\Passport\Token given, called in XXX/vendor/laravel/passport/src/PersonalAccessTokenFactory.php on line 77
```

I can see two options:
* Option 1:  create an interface/contract for `Token` and use that instead then it would be possible to resolve a custom `TokenRepository` and  `AccessTokenRepository` that uses that interface on the save function.
* Option 2:  remove the type hint in the `save()` parameter

While I prefer option 1, I don't see any interfaces being used internally by the package and there are a lot of parameters that aren't type-hinted.  This leaves me leaning towards option 2.

I'm happy to create a PR but unsure which direction to take or if a change would be accepted for this. 

Can you provide some input on the direction I should take? Thanks

driesvints:
Heya. For some reason this issue slipped past by my, sorry for that. 

Can you please provide a thorough explanation why you need this?

jmosul:
We need this as we use MongoDB so our Token class extends a different base Model.

This causes an issue when the TokenRespository save function is called.  It is type hinted to only accept the Token from laravel-passport.  Which we can't override in our own extension of Laravel Passport designmynight/laravel-mongodb-passport

Can I suggest a Token interface as the type hint?

driesvints:
Since Taylor rejected the PR I don't think we'll provide support for this soon, sorry.

uriel2707:
@alasdairmackenzie @jmosul  https://packagist.org/packages/webdecero/webdecero-package-laravel-passport

PR #1073, opened 2019-08-29
Refactor Token to use Token Contract

jmosul:
By using a TokenContract as type hint inside the TokenRepository it allows custom Tokens to be used which is necessary when using a different database such as MongoDB as this requires the Token to extend a different base Model.

For #968


taylorotwell:
Typically we do not suffix interfaces with things like "Contract".

jmosul:
Force of habit with our own naming conventions.

Changed to match the following convention:
https://github.com/laravel/framework/blob/58360ffe7094be96e57c7c2628e3c080222867d0/src/Illuminate/Cache/Lock.php

jmosul:
Fair point, thanks will do

PR #1075, opened 2019-08-30
Allow Token model to update timestamp

vigneshgurusamy:
The current model property does not set to update timestamp in the column `oauth_access_tokens`.`updated_at` on token revoke action.


taylorotwell:
No plans on changing this.

PR #1076, opened 2019-08-30
Token contract

jmosul:
Following up from laravel/passport#1073

> By using a Token interface as type hint inside the TokenRepository it allows custom Tokens to be used which is necessary when using a different database such as MongoDB as this requires the Token to extend a different base Model.
> 
> For #968

Merging into master as breaking change

### DIFF
--- a/src/Contracts/Token.php
+++ b/src/Contracts/Token.php
@@ -9,7 +9,7 @@ use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Routing\UrlRoutable;
 use Illuminate\Contracts\Queue\QueueableEntity;
 
-interface TokenContract extends ArrayAccess, Arrayable, Jsonable, JsonSerializable, QueueableEntity, UrlRoutable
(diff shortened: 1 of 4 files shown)


taylorotwell:
So, does this contract actually make sense for your Mongo implementation? Two of the methods in the interface specify an Eloquent return type.

jmosul:
Yes, we use jenssegers/laravel-mongodb which extends the base Eloquent classes

taylorotwell:
Honestly I think I will hold off on this for now. To me this gives an implicit promise that Passport is supposed to work with other arbitrary ORMs and even different types of databases such as NoSQL databases. I don't want to commit to providing that kind of support. Passport is a first-party Laravel package and as such is intended to work with our first-party ORM.

If you want a Mongo version of Passport I suggest forking it or creating your own version of its functionality in a way that is tuned for Mongo.

jmosul:
I don't think it says passport supports other ORMs/databases, I think it just allows extendability.  If someone replaces Token that's their own risk. Most of Laravel is extendable this just continues in that vein.

PR #1090, opened 2019-10-01
Use config for session name

narwy:
Uses config for the session name not exposing the `laravel` tags by default instead of using `Passport::cookie('custom_token')`

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -80,7 +80,7 @@ class Passport
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


driesvints:
Dd you actually test this? Because a config call in a property can never work. We'll have to refactor this differently I think.

narwy:
> Dd you actually test this? Because a config call in a property can never work. We'll have to refactor this differently I think.

Fair point!

PR #1102, opened 2019-10-31
[8.x] Allow to customize passport auth guard.

dkulyk:
Not always need use `api` guard for passport authorization.

This PR adds ability to change passport guard in config.


driesvints:
@dkulyk heya, thanks for checking into this. I'm not sure this is the proper approach regarding multi-auth support? With these changes you're still only able to set one guard if I see it correctly.

https://github.com/laravel/passport/issues/982

taylorotwell:

Thanks!

Issue #981, opened 2019-03-01
Generate cookie name out of the app's name by default


andriihorpenko:
## Brief overview
The main purpose of this feature is to make Laravel Passport more "unified" with the main application.
## Detailed
The session token name is generating with the following formula:
`Str::slug(env('APP_NAME', 'laravel'), '_').'_session'`
And this is how the Passport is "making" the name:
`public static $cookie = 'laravel_token';`

Why not implement this thing to the Passport by default? This will likely remove most "cookie name override" needs.

Thanks in advance.

driesvints:
Yeah, I could see the benefit here.

driesvints:
Closing as the related PR was closed.

PR #1103, opened 2019-11-01
[9.x] Generate default cookie name from app name

cmorbitzer:
This implements the proposal in #981 to set the default cookie name using the app name.

I'm not sure if this should target current major version as it is a potentially breaking change.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -5,6 +5,7 @@ namespace Laravel\Passport;
 use Carbon\Carbon;
 use DateInterval;
 use DateTimeInterface;
+use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Str;
 use League\OAuth2\Server\ResourceServer;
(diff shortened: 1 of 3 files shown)


driesvints:
@cmorbitzer tests seem to be failing.

driesvints:
Thanks for sending this in but unfortunately this is a breaking change. Every cookie name will be different for apps because it uses the app name. We should do this on master for the next release the soonest I believe.

cmorbitzer:
@driesvints Tests pass now. I'm happy to target master; I think that's a good idea.

taylorotwell:
Is there an actual end-user benefit to this or are we making a breaking change to soothe our own internal need for "consistency" between Laravel and Passport's cookie name generation?

cmorbitzer:
@andriihorpenko speculated that this would eliminate most cookie override needs and I was sympathetic because I override the cookie in this way too.

As far as breaking change goes, this is pretty low impact. It’s a problem if you already use a cookie by the new name or were for some reason counting on the name being what it was. Most applications will carry on as if nothing happened and the old cookie will expire after 2 hours of disuse.

taylorotwell:
Why do you override the cookie?

taylorotwell:
Closing pending no answer to question above.

PR #1107, opened 2019-11-05
Allow to customize passport auth guard.

### LINKED ISSUE #982: Passport Multi-Auth (Implementation)

hakhsin:
Fixes #982

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -17,4 +17,5 @@ return [
 
     'public_key' => env('PASSPORT_PUBLIC_KEY'),
 
+    'auth_guard' => Laravel\Passport\Bridge\AuthGuard::class,
 ];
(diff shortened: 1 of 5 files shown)


taylorotwell:
No plans to change this right now.

hakhsin:
Hi @taylorotwell, is it legal to custom this package and publish it on Packagist? I want to use it for my self.

driesvints:
@hakhsin yes as long as you keep and don't alter the license file.

PR #1119, opened 2019-11-16
[8.x] Fix 'actingAsClient' testing method

gdebrauwer:
The `Passport::actingAsClient()` is broken since version 8.0, because of the changes made by #1040. I updated the method to also mock the use of `TokenRepository` property.


driesvints:
@gdebrauwer Thanks! :)

PR #1120, opened 2019-11-16
[8.x] Presets

fritsstegmann:
WIP. Trying to build preset commands for Laravel Passport. If maintainers are interested I'm happy to follow any instructions to complete functionality.


driesvints:
Thanks but we already can do this with the `vendor:publish` command which was intended for this. See https://laravel.com/docs/6.x/passport#installation

PR #1127, opened 2019-11-24
[8.x] Add abstract CheckCredentials middleware and allows to create

crynobone:
custom implementation of middleware if needed.

Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

### DIFF
--- a/src/Http/Middleware/CheckClientCredentialsForAnyScope.php
+++ b/src/Http/Middleware/CheckClientCredentialsForAnyScope.php
@@ -2,104 +2,45 @@
 
 namespace Laravel\Passport\Http\Middleware;
 
-use Closure;
 use Illuminate\Auth\AuthenticationException;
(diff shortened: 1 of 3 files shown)


src/Http/Middleware/CheckClientCredentialsForAnyScope.php
```diff
@@ -2,104 +2,45 @@
...
 
         foreach ($scopes as $scope) {
             if ($token->can($scope)) {
-                return true;
+                return;
             }
         }
 
-        return false;
+        throw new MissingScopeException($scopes);
```
crynobone: Both `CheckClientCredentials` and `CheckClientCredentialsForAnyScope` are the same except

* `CheckClientCredentials` throws an exception of failure and return `void`.
* `CheckClientCredentialsForAnyScope` return `bool` and `handle` method convert `false` to an exception.

This changes remove the different (making both identical).

PR #1134, opened 2019-12-05
[9.x] Switch from getKey() to getAuthIdentifier() to match Laravel core

### LINKED ISSUE #1006: Use getAuthIdentifier() instead of getKey() to retrieve current user identifier
In core framework, `getAuthIdentifier()` is used to get user identifier for authentication.
However, in passport `getKey()` is used to get user identifier.

Below are the places where made us confused:
https://github.com/laravel/passport/blob/4e3390e67a5a74850698e1e991fd324b4dfa0ffa/src/Http/Controllers/AuthorizedAccessTokenController.php#L35-L42
https://github.com/laravel/passport/blob/55c1c546332db7bbbb95e0e4fd50ee8a690afc11/src/Http/Controllers/AuthorizationController.php#L107-L116

In our case, we used uuid as auth identifier but kept primary key as `$primaryKey` field for relationships.

Would you consider to change to use `getAuthIdentifier()` to be in sync with core framework design?

kaptk2:
Switch from getKey to getAuthIdentifier to better match the core and work better with other 3rd party packages. This pull request closes #1006

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -107,7 +107,7 @@ class AuthorizationController
      */
     protected function approveRequest($authRequest, $user)
     {
-        $authRequest->setUser(new User($user->getKey()));
+        $authRequest->setUser(new User($user->getAuthIdentifier()));
 
         $authRequest->setAuthorizationApproved(true);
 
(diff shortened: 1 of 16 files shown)


Legion112:
@driesvints, could you please give any estimate when will be the major release?

driesvints:
@Legion112 not at this time sorry

PR #1150, opened 2019-12-31
[9.x] Rename `validateForPassportPasswordGrant` to `validateForPassport`

gdebrauwer:
This PR renames the method `validateForPassportPasswordGrant` to `validateForPassport`.

I think it's a good idea to rename this method for a couple of reasons:

1. There is no real reason to specify 'PasswordGrant' in the method name. There are no other grant types that allow for similar customization.
2. It puts the method name in line with the 2 other Passport customization methods `findForPassport` and `findAndValidateForPassport`.
3. The name of the method is easier to remember.

I know it's a breaking change, but it's a very small one. It requires very little upgrade effort, while in the long run, it makes it clearer for people that want to use those methods.

### DIFF
--- a/src/Bridge/UserRepository.php
+++ b/src/Bridge/UserRepository.php
@@ -56,8 +56,8 @@ class UserRepository implements UserRepositoryInterface
 
         if (! $user) {
             return;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Pointless breaking change IMO.

PR #1152, opened 2020-01-02
[8.x] Replace deprecated package

JC5:
The Zend package is deprecated. Replaced it with the new one. According to the docs, the package are the same except for the name.

- https://github.com/laminas/laminas-diactoros
- https://github.com/zendframework/zend-diactoros


taylorotwell:
This doesn't update any namespaces.

PR #1154, opened 2020-01-04
[8.x] Use Hasher interface instead of HashManager

X-Coder264:
Currently there is no way to specify the hasher implementation because the hash manager would always return the default driver implementation. There is also no reason for the entire hash manager being there as we only need the hasher implementation and `$hasher->driver()` which was called in the constructor always returns the default one which is gonna be [injected by default anyway](https://github.com/laravel/framework/blob/v6.9.0/src/Illuminate/Hashing/HashServiceProvider.php#L22).

There are no breaking changes here as [the HashManager also implements the hasher interface](https://github.com/laravel/framework/blob/v6.9.0/src/Illuminate/Hashing/HashManager.php#L8).

### DIFF
--- a/src/Bridge/UserRepository.php
+++ b/src/Bridge/UserRepository.php
@@ -2,7 +2,7 @@
 
 namespace Laravel\Passport\Bridge;
 
-use Illuminate\Hashing\HashManager;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Let's do this on the next major release.

PR #1155, opened 2020-01-05
[8.x] Add tests for Passport::actingAs and Passport::actingAsClient

X-Coder264:
Currently Passport is being only unit tested which is great except that each method where Laravel global helpers are being used cannot be tested at all.

This is the case with the `Passport::actingAsClient` method which uses the `app` helper to resolve and bind stuff with the container. Because of that the method wasn't tested which resulted in it being broken from July (https://github.com/laravel/passport/pull/1040) till November (https://github.com/laravel/passport/pull/1119). The fix was made in November but as there are still no tests for it we won't know if it breaks again in the future.

This PR adds Orchestra Testbench so that we can write full functional tests with the whole framework and I've added tests for the `Passport::actingAs` and `Passport::actingAsClient` methods.

In doing so I had to fix the `storage_path` function conflict because now the foundation `storage_path` function exists.

I've also created an abstract `PassportTestCase` class, it does a few things so that we don't need to do them in almost every test (as they are gonna be needed):

1. It calls the `passport:keys` command which generates the oauth keys (if they don't exist already) - without the keys the `ResourceServer` class throws an exception on instantiation and the `ResourceServer` is probably gonna be instantiated at some point in 95% of the functional tests.
2. It sets the `passport` driver for the auth API guard.
3. It registers the `PassportServiceProvider` without which a bunch of bindings would be missing from the container.

This paves the way for further covering Passport functionality with feature/functional tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
     },
     "require-dev": {
         "mockery/mockery": "^1.0",
+        "orchestra/testbench": "^4.4|^5.0",
         "phpunit/phpunit": "^8.0"
     },
     "autoload": {
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -8,6 +8,7 @@ use DateTimeInterface;
 use Illuminate\Support\Facades\Route;
 use League\OAuth2\Server\ResourceServer;
 use Mockery;
+use Psr\Http\Message\ServerRequestInterface;
 
 class Passport
 {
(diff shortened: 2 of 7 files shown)


taylorotwell:
If you're requiring testbench why do you need to require laravel/framework?

X-Coder264:
@taylorotwell I required `orchestra/testbench-core` instead of `orchestra/testbench` the first time so I had to require `laravel/framework` manually. This is now fixed.

driesvints:
@X-Coder264 thanks!

X-Coder264:
@driesvints No problem :)

PR #1160, opened 2020-01-09
[8.x] Added support for API resources

jezzdk:
I added support for API resources since thats what I normally use. The feature wont mess with previous installs, since the feature is turned off by default. I just thought it would be a nice little addition :)


driesvints:
Heya, this seems to break the tests and the pr doesn't has tests of its own. Feel free to resubmit once you've added some tests and made the test suite pass. Thanks.

PR #1169, opened 2020-01-21
[8.x] Shouldn't the migrations use bigIncrements and indexes on relationships?

KieronWiltshire:
I was looking through the migrations and saw that some of them seem pretty dated and don't use bigIncrements and indexes on certain possible relationships like the `user_id` field [here][1].

[1]: https://github.com/KieronWiltshire/passport/blob/8.x/database/migrations/2016_06_01_000004_create_oauth_clients_table.php#L18

### DIFF
--- a/database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
+++ b/database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
@@ -15,8 +15,9 @@ class CreateOauthAuthCodesTable extends Migration
     {
         Schema::create('oauth_auth_codes', function (Blueprint $table) {
             $table->string('id', 100)->primary();
-            $table->bigInteger('user_id');
-            $table->unsignedInteger('client_id');
+            $table->bigInteger('user_id')->unsigned()->index();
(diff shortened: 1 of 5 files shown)


driesvints:
Have you tried these changes on all four supported DB drivers? Sqlite, MySQL, PostGres & SQL Srv?

KieronWiltshire:
@driesvints I have only tested this with MySQL, I was under the assumption that the point behind the Schema interface is that it abstracts the four supported drivers away

driesvints:
@KieronWiltshire have you checked into how the cascading deletes affect the underlying repository code? I suspect atm the code within Passport itself is currently handling the deletion of related records etc.

KieronWiltshire:
@driesvints I didn't see the point as I couldn't find any use cases as to where it would be damaging, for example... if a client is deleted, shouldn't access tokens etc be deleted along with? there would be no reason to keep them?

taylorotwell:
Remove the cascading foreign key you added.

clemblanco:
Same question for `oauth_personal_access_clients`.`client_id`?

PR #1171, opened 2020-01-24
[8.x] Add a Passport Client factory to Passport publishing

Orrison:
In the help documentation for 6.x and up there is reference in the testing section to test Client access with the method `Passport::actingAsClient` and that it takes a Client model created by a factory to be passed in. https://laravel.com/docs/6.x/passport#testing

```php
use Laravel\Passport\Client;

public function testGetOrders()
{
    Passport::actingAsClient(
        factory(Client::class)->create(),
        ['check-status']
    );

    $response = $this->get('/api/orders');

    $response->assertStatus(200);
}
```

Just trying that code out of the box fails since there is no published asset for a `Laravel\Passport\Client` factory. The user is left to make one on their own.
This will publish a Client factory that the user can use for this.

I don't believe it will break any existing features as it is just publishing a factory for the user. It could possibly overwrite one the user has already created if it was the same filename?
I think it will make testing this feature easier for Laravel users as they don't have to reference the `oauth_clients` table and create the factory themselves.

I don't have any tests included as I am not sure how to test the publish functionality and it seems the other publishes are not tested. If you would like me to write some tests please let me know. I could try harder to find a way to test that.

Thank you!

### DIFF
--- a/database/factories/PassportClientFactory.php
+++ b/database/factories/PassportClientFactory.php
@@ -1,0 +1,18 @@
+<?php
+
+/** @var \Illuminate\Database\Eloquent\Factory $factory */
+
+use Faker\Generator as Faker;
+use Laravel\Passport\Client;
+
+$factory->define(Client::class, function (Faker $faker) {
+    return [
+        'user_id' => null,
+        'name' => $faker->company,
+        'secret' => Str::random(40),
+        'redirect' => $faker->url,
+        'personal_access_client' => 0,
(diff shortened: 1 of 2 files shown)


driesvints:
If you bump the minimum Laravel components version to 6.12.0 we can use the new `loadFactoriesFrom` method. But then they can't be adjusted so it's probably best to do a combination of both.

https://github.com/laravel/framework/pull/31133

Issue #1173, opened 2020-01-28
Migrations fails


adumskis:
- Passport Version: v8.3.0
- Laravel Version: 6.13.1
- PHP Version: 7.3.14
- Database Driver & Version: MariaDB 10.4

### Description:
Migrations after fresh install fails:
`PDOException::("SQLSTATE[HY000]: General error: 1005 Can't create table db.oauth_auth_codes (errno: 150 "Foreign key constraint is incorrectly formed")")`

2016_06_01_000001_create_oauth_auth_codes_table.php contains a foreign key to `oauth_clients` table that will be created only in 2016_06_01_000004_create_oauth_clients_table.php.

### Steps To Reproduce:
Require passport via composer and `php artisan migrate`/`php artisan migrate:fresh`.

lloricode:
same as mine

lloricode:
```
Migrating: 2016_06_01_000001_create_oauth_auth_codes_table

   Illuminate\Database\QueryException  : SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint (SQL: alter table `oauth_auth_codes` add constraint `oauth_auth_codes_client_id_foreign` foreign key (`client_id`) references `oauth_clients` (`id`) on delete cascade)

  at /home/vagrant/code/laravels/laravel-boilerplate/vendor/laravel/framework/src/Illuminate/Database/Connection.php:669
    665|         // If an exception occurs when attempting to run a query, we'll format the error
    666|         // message to include the bindings with SQL, which will make this exception a
    667|         // lot more helpful to the developer instead of just the database's errors.
    668|         catch (Exception $e) {
  > 669|             throw new QueryException(
    670|                 $query, $this->prepareBindings($bindings), $e
    671|             );
    672|         }
    673|

  Exception trace:

  1   Doctrine\DBAL\Driver\PDOException::("SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint")
      /home/vagrant/code/laravels/laravel-boilerplate/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:123

  2   PDOException::("SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint")
      /home/vagrant/code/laravels/laravel-boilerplate/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:121

  Please use the argument -v to see more details.
[...]

PR #1174, opened 2020-01-29
[8.x] Fix Foreign key oauth_clients in oauth_auth_codes and oauth_access_t…

carlossantesp:
There is a failure in migration
**2016_06_01_000001_create_oauth_auth_codes_table** and **2016_06_01_000002_create_oauth_access_tokens_table** where a foreign key was created before creating the oauth_clients table, remove it and create two additional migrations where add the foreign keys to their respective tables **oauth_auth_codes** and **oauth_access_tokens**

PR #1175, opened 2020-01-29
Swapped migration execution order which fixes referencing issues

KieronWiltshire:
I broke the branch, it's only fair I fix it. Reference to #1173


KieronWiltshire:
I'm closing this as swapping the execution order will result in migrations being re-run from older versions.

PR #1174 is probably a much nicer solution as it will add the foreign keys to older migrations too.

gebrail:
Solve my mistake, I have changed the order of migrations and it worked.

Current Order

2016_06_01_000001_create_oauth_auth_codes_table
2016_06_01_000002_create_oauth_access_tokens_table
2016_06_01_000003_create_oauth_refresh_tokens_table
2016_06_01_000004_create_oauth_clients_table
2016_06_01_000005_create_oauth_personal_access_clients_table

Order correctly ordered

2016_06_01_000001_create_oauth_clients_table
2016_06_01_000002_create_oauth_auth_codes_table
2016_06_01_000003_create_oauth_access_tokens_table
2016_06_01_000004_create_oauth_refresh_tokens_table
2016_06_01_0000055_create_oauth_personal_access_clients_table

KieronWiltshire:
the only issue with swapping the order is that if someone upgrades, they will be running migrations twice because the names of the migrations have been changed.

PR #1181, opened 2020-02-08
[8.x] Add tests for getting the access token with the password grant

X-Coder264:
After adding the support for feature tests in #1155 I'm sending this PR which contains feature tests for getting the access token with the password grant.

If this gets merged I'll probably also add the tests for the other grants when I'll have the time.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -20,4 +20,8 @@
             <directory suffix=".php">./src/</directory>
         </whitelist>
     </filter>
+
+    <php>
+        <env name="APP_KEY" value="AckfSECXIvnK5r28GVIWUAxmbBSjTsmF"/>
+    </php>
 </phpunit>
--- a/tests/Feature/AccessTokenControllerTest.php
+++ b/tests/Feature/AccessTokenControllerTest.php
@@ -1,0 +1,189 @@
+<?php
+
+namespace Laravel\Passport\Tests\Feature;
+
+use Carbon\CarbonImmutable;
+use Illuminate\Contracts\Hashing\Hasher;
+use Illuminate\Database\Eloquent\Factory;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Laravel\Passport\Client;
+use Laravel\Passport\ClientRepository;
+use Laravel\Passport\HasApiTokens;
+use Laravel\Passport\Token;
+use Laravel\Passport\TokenRepository;
+use Lcobucci\JWT\Parser;
+
+class AccessTokenControllerTest extends PassportTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        Schema::create('users', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('email')->unique();
+            $table->string('password');
+            $table->dateTime('created_at');
+            $table->dateTime('updated_at');
+        });
+    }
+
+    protected function tearDown(): void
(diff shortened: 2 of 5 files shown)


phpunit.xml.dist
```diff
@@ -20,4 +20,8 @@
             <directory suffix=".php">./src/</directory>
         </whitelist>
     </filter>
+
```
driesvints:
```suggestion
```

driesvints:
This all looks pretty promising. Thanks for your work on this 👍

driesvints:
@X-Coder264 any reason in particular why you used mysql instead of sqlite?

X-Coder264:
@driesvints There's no particular reason, it was just so easy to setup mysql that I saw no reason not to (especially since sqlite is not a database that is used in production).

driesvints:
@X-Coder264 we use sqlite for all our testing in all libraries so it might be best that we switch that. I can't for example run the tests suite locally now out of the box without having to setup a mysql database. Can you send in a new PR to change that maybe?

X-Coder264:
I actually did this to be just like how laravel/framework is tested, but yeah, I see your point. I can send a PR for this sometime tomorrow probably as I'm swamped today.

driesvints:
I've updated the test suite to use sqlite instead.

X-Coder264:
@driesvints Sorry for not doing it myself, I had a ton of work, but I hadn't forgotten it, it was on my TODO list for the weekend. Thanks for taking care of it.

driesvints:
@X-Coder264 no problem at all :)

tests/Feature/AccessTokenControllerTest.php
```diff
@@ -0,0 +1,189 @@
...
+
+        $decodedResponse = $response->decodeResponseJson();
+
+        $this->assertArrayHasKey('token_type', $decodedResponse);
+        $this->assertArrayHasKey('expires_in', $decodedResponse);
+        $this->assertArrayHasKey('access_token', $decodedResponse);
+        $this->assertArrayHasKey('refresh_token', $decodedResponse);
+        $this->assertSame('Bearer', $decodedResponse['token_type']);
+        $expiresInSeconds = 31622400;
+        $this->assertEqualsWithDelta($expiresInSeconds, $decodedResponse['expires_in'], 5);
```
driesvints: @X-Coder264 this line was failing for us randomly. I had to update the value of the `$expiresInSeconds` variable to 31536000 instead. Any idea why? I also don't really know what we're testing here exactly? Why should the `expires_in` be the value of above?

X-Coder264: @driesvints This tests the expiration time of the issued access token. The expiration time for the password grant is being set here:

https://github.com/laravel/passport/blob/v8.4.0/src/PassportServiceProvider.php#L116

This calls `Passport::tokensExpireIn()` which then returns `new DateInterval('P1Y')` (because there is no `$date` provided and its default value is `null` and `static::$tokensExpireAt` is `null` too).

The difference between our numbers is 86400 seconds which is exactly one day. I guess there is some weird time related stuff going on with that `DateInterval` which causes a random fail of that specific assertion.

X-Coder264: Did the test maybe start to fail after 29.02.? If so then it is related to the leap year which is exactly why there was a one day difference between our expiration numbers 😄

driesvints: hah, that could be yeah

Issue #1190, opened 2020-02-12
Made lifetime for auth code easy customizable


prodigy7:
So far I hope understand right, auth code are currently valid for 10 minutes. The OAuth 2.0 spec recommends a maximum lifetime of 10 minutes and I want set a shorter lifetime (30-60 seconds).

So far I understand, the lifetime is set here (passport/src/PassportServiceProvider.php):
```
    protected function registerAuthorizationServer()
    {
        $this->app->singleton(AuthorizationServer::class, function () {
            return tap($this->makeAuthorizationServer(), function ($server) {
                $server->setDefaultScope(Passport::$defaultScope);

                $server->enableGrantType(
                    $this->makeAuthCodeGrant(), Passport::tokensExpireIn()
                );

                $server->enableGrantType(
                    $this->makeRefreshTokenGrant(), Passport::tokensExpireIn()
                );

                $server->enableGrantType(
                    $this->makePasswordGrant(), Passport::tokensExpireIn()
                );

                $server->enableGrantType(
                    new PersonalAccessGrant, Passport::personalAccessTokensExpireIn()
                );

                $server->enableGrantType(
                    new ClientCredentialsGrant, Passport::tokensExpireIn()
                );

                if (Passport::$implicitGrantEnabled) {
                    $server->enableGrantType(
                        $this->makeImplicitGrant(), Passport::tokensExpireIn()
                    );
                }
            });
        });
    }


    /**
     * Create and configure an instance of the Auth Code grant.
     *
     * @return \League\OAuth2\Server\Grant\AuthCodeGrant
     */
    protected function makeAuthCodeGrant()
    {
        return tap($this->buildAuthCodeGrant(), function ($grant) {
            $grant->setRefreshTokenTTL(Passport::refreshTokensExpireIn());
        });
    }
```
When I see it right, i could change the lifetime with Passport::tokensExpireIn($value) but it would also have a impact to other grant types like 'refresh' or 'password'. Is there any easy way, change it for auth code only without leaving default code?
I know, extend/override the service provider replacing the logic in this registerAuthorizationServer method with your own is possible, but also a step more own code only for a simple setting.

So I wish, the auth code has a own method for override default lifetime setting.

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

[...]

PR #1195, opened 2020-02-15
[8.x] Implement set auth codes expire as requested in #1190

prodigy7:
As requested by me in #1190 i implement the possibility, change the auth codes expires.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -71,6 +71,13 @@ class Passport
     public static $refreshTokensExpireAt;
 
     /**
+     * The date when auth codes expire.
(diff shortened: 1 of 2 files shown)


taylorotwell:
I think the default is fine.

prodigy7:
Hm, whats wrong with made auth code lifetime configurable? Currently rfc max lifetime is implemented but I want the rfc suggested lifetime. Its also possible, change other lifetime, so why not auth code lifetime?
At the moment it is not objectively comprehensible to me why decisions are made in this way.

PR #1201, opened 2020-02-20
[8.x] GitHub actions

driesvints:
Replaces Travis with Github Actions. Inspired by the workflow file from https://github.com/spatie/laravel-medialibrary/blob/master/.github/workflows/run-tests.yml


driesvints:
@iBotPeaches are you saying you were relying on `DiactorosFactory` directly? Because then you should reference psr-http-message-bridge in your composer.json directly.

iBotPeaches:
@driesvints  - That worked. Apologies for the message, not sure how I missed that deprecation, since it looks like it has been deprecated for quite a while.

driesvints:
No worries. Always make sure to add dependencies which your rely on directly to your `composer.json`.

PR #1205, opened 2020-03-20
[8.x] mergeConfigFrom() already check if app is running with config cached.

crynobone:
https://github.com/illuminate/support/blob/6.x/ServiceProvider.php#L60-L67

This solved issue with laravel/lumen-framework#938

### DIFF
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -87,9 +87,7 @@ class PassportServiceProvider extends ServiceProvider
      */
     public function register()
     {
-        if (! $this->app->configurationIsCached()) {
-            $this->mergeConfigFrom(__DIR__.'/../config/passport.php', 'passport');
-        }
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks!

lloricode:
Hi @driesvints, can you set tag with this?

Thank you

driesvints:
@lloricode release day is today

driesvints:
FYI: we don't officially support using Passport with Lumen. Please consider migrating to Laravel.

Exotelis:
> FYI: we don't officially support using Passport with Lumen. Please consider migrating to Laravel.

I'd like to mention that if you just want to build an API without sessions and all these stuff, lumen is the right choice. `laravel/passwort` work perfectly with Lumen when you use the `dusterio/lumen-passport` package, even after the upgrade to 7.x

PR #1207, opened 2020-03-26
[8.x] Fix resolveInheritedScopes.

lucasmichot:
Previously the first element returns by `resolveInheritedScopes` was an empty string.
This PR fixes this.


taylorotwell:
Do we need a test?

lucasmichot:
I've just added one, just in case: 92f729ff68490a3f2483d6686f900f8c3c109663

PR #1212, opened 2020-04-02
[7.x] Add option to customize ScopeModel

ItsZiroy:
Hi. 
This is my first pull request to a laravel project and I might have made a lot of foolish mistakes, but the idea is clear I suppose and I thought a pull request might be better than just creating an issue. 
Me and my team would like to have the possibility to provide a custom Scope model, because we feel kind of limited to just a name and a description.
The advantage you get from it is, that you can add fields like a title or a icon name, making it much easier to customize the authorization page and provide the user with a nice looking and descriptive authorization page.
The reason we find a customized scope model good, is that we are currently working with dozens of if requests to provide a easier to understand page, which is truly awful.
Please let me know, whether this is already enough change, or if I have missed something crucial. As mentioned, this is my first pull request.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -124,6 +124,13 @@ class Passport
      * @var string
      */
     public static $tokenModel = 'Laravel\Passport\Token';
+    
+     /**
+     * The scope model class name.
+     *
+     * @var string
+     */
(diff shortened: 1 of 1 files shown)


driesvints:
Please send this to the 8.x branch.

PR #1213, opened 2020-04-02
[8.x] Add option to customize Scope Model 

ItsZiroy:
Hi.
This is my first pull request to a laravel project and I might have made a lot of foolish mistakes, but the idea is clear I suppose and I thought a pull request might be better than just creating an issue.
Me and my team would like to have the possibility to provide a custom Scope model, because we feel kind of limited to just a name and a description.
The advantage you get from it is, that you can add fields like a title or a icon name, making it much easier to customize the authorization page and provide the user with a nice looking and descriptive authorization page.
The reason we find a customized scope model good, is that we are currently working with dozens of if requests to provide a easier to understand page, which is truly awful.
Please let me know, whether this is already enough change, or if I have missed something crucial. As mentioned, this is my first pull request.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -132,6 +132,13 @@ class Passport
      * @var string
      */
     public static $refreshTokenModel = 'Laravel\Passport\RefreshToken';
+    
+     /**
+     * The scope model class name.
+     *
+     * @var string
+     */
+    public static $scopeModel = 'Laravel\Passport\Scope';
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hello,

I don't see any code here for how Passport would actually use this new Scope model?

I think you could maybe work around this yourself by defining your own App\Scope class or something and then creating an instance of it from a given string... like Scope::from($scopeString)... then working with that to display icons or whatever else you need.

Issue #161, opened 2016-10-27
Passport Multi-Auth


SidharthRaveendran:
Would passport be implementing Multi-Auth system? or is there another way to incorporate the multi-auth into the app.

I have created Multiple Models which use Laravel's Auth system to implement proper multi-auth system. I'm not sure how to use passport for the same.

eepawan:
I am also looking for same. Anyone here to help?

jenky:
https://github.com/laravel/passport/blob/7ed1a0b80c1da9a3bf25eb180068ddd005a82068/src/Bridge/UserRepository.php#L35

This package always uses users provider model in `config/auth` to authenticate, so I guess there are no hacks for now?

belov91:
+1

gauravmak:
With #216 , it is dynamic now so multiple guards can be added and used with passport now I believe.

SidharthRaveendran:
Still we cannot implement multi-auth using laravel using this update.

zubair1:
I managed to get access tokens from different auth providers in laravel, but I had to modify:

-------------------------------
vendor\laravel\passport\src\Bridge\UserRepository.php
- Copy / Paste `getUserEntityByUserCredentials` to make a duplicate of it and name it `getEntityByUserCredentials`

- Then, in the new duplicated function, find the below:
`$provider = config('auth.guards.api.provider');`
and Replace it with:
`$provider = config('auth.guards.'.$provider.'.provider');`

-------------------------------
vendor\league\oauth2-server\src\Grant\PasswordGrant.php
- Find Line: 94
- Update with the following code:
```
        $user = $this->userRepository->getEntityByUserCredentials(
            $username,
            $password,
            $this->getIdentifier(),
            $client,
            $provider
        );
```

-------------------------------
After doing this you'll be able to pass an extra key/value pair to your access token request, like for example:
```
grant_type => password,
client_id => someclientid
client_secret => somesecret,
username => someuser,
password => somepass,
client_scope => *,
provider => api
```
In this example, I set provider to 'api', which is default. But now I can pass other auth providers as part of the request. So, just to show another example, I could do the following:

```
grant_type => password,
client_id => someclientid
client_secret => somesecret,
username => someuser,
password => somepass,
client_scope => *,
provider => api_admins
```

Now I used 'api_admins' auth provider to authenticate tokens.

P.S:
[...]

PR #1220, opened 2020-04-17
[9.x] Initial support for multiple providers

billriess:
This PR adds support to allow passport to use and enforce multiple providers.

### The issue:
Currently, Passport only uses the `UserProvider` passed into it from the `RequestGuard`. This can be problematic if you have more than one provider (config/auth.php) that needs to be authenticated. Since Passport has no knowledge of other `UserProvider` you can end up in situations where the wrong user is authenticated and returned.

### The solution:
This PR adds support to optionally set the provider that should be used on the `Passport\Client`. If a provider is defined on the `Passport\Client`, it will compare it against the `UserProvider` coming in from the `RequestGuard` and return the user if they match. This ensures that the user being returned is coming from the right model. If there is a mismatch, no user is returned, thus leaving you with an unauthenticated request. If no provider is defined on the `Passport\Client` then everything works as it always has.

### Breaking changes:
The only big breaking change here is that this requires a schema change to the `oauth_clients` table to add the `provider` column.

### DIFF
--- a/database/migrations/2016_06_01_000004_create_oauth_clients_table.php
+++ b/database/migrations/2016_06_01_000004_create_oauth_clients_table.php
@@ -18,6 +18,7 @@ class CreateOauthClientsTable extends Migration
             $table->unsignedBigInteger('user_id')->nullable()->index();
             $table->string('name');
             $table->string('secret', 100)->nullable();
+            $table->string('provider')->nullable();
             $table->text('redirect');
             $table->boolean('personal_access_client');
             $table->boolean('password_client');
--- a/src/Bridge/Client.php
+++ b/src/Bridge/Client.php
@@ -16,16 +16,21 @@ class Client implements ClientEntityInterface
      */
     protected $identifier;
 
+    /**
+     * The client's provider.
+     *
+     * @var string
+     */
     public $provider;
 
     /**
      * Create a new client instance.
      *
-     * @param  string  $identifier
-     * @param  string  $name
-     * @param  string  $redirectUri
-     * @param  bool  $isConfidential
-     * @return void
+     * @param string $identifier
+     * @param string $name
+     * @param string $redirectUri
+     * @param bool $isConfidential
+     * @param string|null $provider
      */
     public function __construct($identifier, $name, $redirectUri, $isConfidential = false, $provider = null)
     {
--- a/src/Bridge/Client.php
+++ b/src/Bridge/Client.php
@@ -16,6 +16,8 @@ class Client implements ClientEntityInterface
      */
     protected $identifier;
 
+    public $provider;
+
     /**
      * Create a new client instance.
      *
@@ -25,13 +27,14 @@ class Client implements ClientEntityInterface
      * @param  bool  $isConfidential
      * @return void
      */
-    public function __construct($identifier, $name, $redirectUri, $isConfidential = false)
+    public function __construct($identifier, $name, $redirectUri, $isConfidential = false, $provider = null)
     {
         $this->setIdentifier((string) $identifier);
 
         $this->name = $name;
         $this->isConfidential = $isConfidential;
         $this->redirectUri = explode(',', $redirectUri);
+        $this->provider = $provider;
     }
 
     /**
--- a/src/Bridge/Client.php
+++ b/src/Bridge/Client.php
@@ -26,11 +26,12 @@ class Client implements ClientEntityInterface
     /**
      * Create a new client instance.
      *
-     * @param string $identifier
-     * @param string $name
-     * @param string $redirectUri
-     * @param bool $isConfidential
-     * @param string|null $provider
+     * @param  string  $identifier
+     * @param  string  $name
+     * @param  string  $redirectUri
+     * @param  bool  $isConfidential
+     * @param  string|null  $provider
+     * @return void
      */
     public function __construct($identifier, $name, $redirectUri, $isConfidential = false, $provider = null)
     {
--- a/src/Client.php
+++ b/src/Client.php
@@ -3,7 +3,6 @@
 namespace Laravel\Passport;
 
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Support\Facades\Auth;
 
 class Client extends Model
 {
--- a/src/Client.php
+++ b/src/Client.php
@@ -49,8 +49,10 @@ class Client extends Model
      */
     public function user()
     {
+        $provider = $this->provider ?: config('auth.guards.api.provider');
+
         return $this->belongsTo(
-            config('auth.providers.'.$this->provider ?: config('auth.guards.api.provider').'.model')
+            config("auth.providers.$provider.model")
         );
     }
 
@@ -107,7 +109,7 @@ class Client extends Model
     /**
      * Get the client's provider.
      *
-     * @return mixed
+     * @return \Illuminate\Contracts\Auth\UserProvider|null
      */
     public function getProvider()
     {
--- a/src/Client.php
+++ b/src/Client.php
@@ -51,7 +51,7 @@ class Client extends Model
         $provider = $this->provider ?: config('auth.guards.api.provider');
 
         return $this->belongsTo(
-            config("auth.providers.$provider.model")
+            config("auth.providers.{$provider}.model")
         );
     }
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -105,14 +105,4 @@ class Client extends Model
     {
         return ! empty($this->secret);
     }
-
-    /**
-     * Get the client's provider.
-     *
-     * @return \Illuminate\Contracts\Auth\UserProvider|null
-     */
-    public function getProvider()
-    {
-        return $this->provider ? Auth::createUserProvider($this->provider) : null;
-    }
 }
--- a/src/Client.php
+++ b/src/Client.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Auth;
 
 class Client extends Model
 {
@@ -49,7 +50,7 @@ class Client extends Model
     public function user()
     {
         return $this->belongsTo(
-            config('auth.providers.'.config('auth.guards.api.provider').'.model')
+            config('auth.providers.'.$this->provider ?: config('auth.guards.api.provider').'.model')
         );
     }
 
@@ -101,5 +102,15 @@ class Client extends Model
     public function confidential()
     {
         return ! empty($this->secret);
+    }
+
+    /**
+     * Get the client's provider.
+     *
+     * @return mixed
+     */
+    public function getProvider()
+    {
+        return $this->provider ? Auth::createUserProvider($this->provider) : null;
     }
 }
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -101,12 +101,13 @@ class ClientRepository
     /**
      * Store a new client.
      *
-     * @param  int  $userId
-     * @param  string  $name
-     * @param  string  $redirect
-     * @param  bool  $personalAccess
-     * @param  bool  $password
-     * @param  bool  $confidential
+     * @param int $userId
+     * @param string $name
+     * @param string $redirect
+     * @param string|null $provider
+     * @param bool $personalAccess
+     * @param bool $password
+     * @param bool $confidential
      * @return \Laravel\Passport\Client
      */
     public function create($userId, $name, $redirect, $provider = null, $personalAccess = false, $password = false, $confidential = true)
@@ -147,9 +148,10 @@ class ClientRepository
     /**
      * Store a new password grant client.
      *
-     * @param  int  $userId
-     * @param  string  $name
-     * @param  string  $redirect
+     * @param int $userId
+     * @param string $name
+     * @param string $redirect
+     * @param string|null $provider
      * @return \Laravel\Passport\Client
      */
     public function createPasswordGrantClient($userId, $name, $redirect, $provider = null)
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -109,13 +109,14 @@ class ClientRepository
      * @param  bool  $confidential
      * @return \Laravel\Passport\Client
      */
-    public function create($userId, $name, $redirect, $personalAccess = false, $password = false, $confidential = true)
+    public function create($userId, $name, $redirect, $provider = null, $personalAccess = false, $password = false, $confidential = true)
     {
         $client = Passport::client()->forceFill([
             'user_id' => $userId,
             'name' => $name,
             'secret' => ($confidential || $personalAccess) ? Str::random(40) : null,
             'redirect' => $redirect,
+            'provider' => $provider,
             'personal_access_client' => $personalAccess,
             'password_client' => $password,
             'revoked' => false,
@@ -136,7 +137,7 @@ class ClientRepository
      */
     public function createPersonalAccessClient($userId, $name, $redirect)
     {
-        return tap($this->create($userId, $name, $redirect, true), function ($client) {
+        return tap($this->create($userId, $name, $redirect, null, true), function ($client) {
             $accessClient = Passport::personalAccessClient();
             $accessClient->client_id = $client->id;
             $accessClient->save();
@@ -151,9 +152,9 @@ class ClientRepository
      * @param  string  $redirect
      * @return \Laravel\Passport\Client
      */
-    public function createPasswordGrantClient($userId, $name, $redirect)
+    public function createPasswordGrantClient($userId, $name, $redirect, $provider = null)
     {
-        return $this->create($userId, $name, $redirect, false, true);
+        return $this->create($userId, $name, $redirect, $provider, false, true);
     }
 
     /**
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -101,13 +101,13 @@ class ClientRepository
     /**
      * Store a new client.
      *
-     * @param int $userId
-     * @param string $name
-     * @param string $redirect
-     * @param string|null $provider
-     * @param bool $personalAccess
-     * @param bool $password
-     * @param bool $confidential
+     * @param  int  $userId
+     * @param  string  $name
+     * @param  string  $redirect
+     * @param  string|null  $provider
+     * @param  bool  $personalAccess
+     * @param  bool  $password
+     * @param  bool  $confidential
      * @return \Laravel\Passport\Client
      */
     public function create($userId, $name, $redirect, $provider = null, $personalAccess = false, $password = false, $confidential = true)
@@ -148,10 +148,10 @@ class ClientRepository
     /**
      * Store a new password grant client.
      *
-     * @param int $userId
-     * @param string $name
-     * @param string $redirect
-     * @param string|null $provider
+     * @param  int  $userId
+     * @param  string  $name
+     * @param  string  $redirect
+     * @param  string|null  $provider
      * @return \Laravel\Passport\Client
      */
     public function createPasswordGrantClient($userId, $name, $redirect, $provider = null)
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -30,6 +30,7 @@ class InstallCommand extends Command
     public function handle()
     {
         $provider = in_array('users', array_keys(config('auth.providers'))) ? 'users' : null;
+
         $this->call('passport:keys', ['--force' => $this->option('force'), '--length' => $this->option('length')]);
         $this->call('passport:client', ['--personal' => true, '--name' => config('app.name').' Personal Access Client']);
         $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client', '--provider' => $provider]);
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -31,6 +31,6 @@ class InstallCommand extends Command
     {
         $this->call('passport:keys', ['--force' => $this->option('force'), '--length' => $this->option('length')]);
         $this->call('passport:client', ['--personal' => true, '--name' => config('app.name').' Personal Access Client']);
-        $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client']);
+        $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client', '--provider' => array_keys(config('auth.providers'))[0]]);
     }
 }
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -29,8 +29,9 @@ class InstallCommand extends Command
      */
     public function handle()
     {
+        $provider = in_array('users', array_keys(config('auth.providers'))) ? 'users' : null;
         $this->call('passport:keys', ['--force' => $this->option('force'), '--length' => $this->option('length')]);
         $this->call('passport:client', ['--personal' => true, '--name' => config('app.name').' Personal Access Client']);
-        $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client', '--provider' => array_keys(config('auth.providers'))[0]]);
+        $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client', '--provider' => $provider]);
     }
 }
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -98,7 +98,7 @@ class TokenGuard
         }
 
         // Determine if the client's provider and the request's provider have matching models.
-        return $client && $client->getProvider()->getModel() === $this->provider->getModel();
+        return $client->getProvider()->getModel() === $this->provider->getModel();
     }
 
     /**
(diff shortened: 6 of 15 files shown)


driesvints:
Hey @billriess. Thanks a lot for giving this another go! I've marked the PR as draft because some tests are still failing. Let's get those fixed first before we mark it as ready.

database/migrations/2016_06_01_000004_create_oauth_clients_table.php
```diff
@@ -18,6 +18,7 @@ public function up()
             $table->unsignedBigInteger('user_id')->nullable()->index();
             $table->string('name');
             $table->string('secret', 100)->nullable();
+            $table->string('provider')->nullable();
```
driesvints, line 21: I guess this would be a breaking change for anyone who already ran the migrations? Sounds like we need to move the pr to the master branch?

src/Bridge/Client.php
```diff
@@ -16,6 +16,8 @@ class Client implements ClientEntityInterface
      */
     protected $identifier;
 
+    public $provider;
```
driesvints, line 19: Needs a DocBlock here.

src/Bridge/Client.php
```diff
@@ -25,13 +27,14 @@ class Client implements ClientEntityInterface
      * @param  bool  $isConfidential
      * @return void
      */
-    public function __construct($identifier, $name, $redirectUri, $isConfidential = false)
+    public function __construct($identifier, $name, $redirectUri, $isConfidential = false, $provider = null)
```
driesvints, line 30: Param needs to be added to the DocBlock above.

src/ClientRepository.php
```diff
@@ -109,13 +109,14 @@ public function personalAccessClient()
      * @param  bool  $confidential
      * @return \Laravel\Passport\Client
      */
-    public function create($userId, $name, $redirect, $personalAccess = false, $password = false, $confidential = true)
+    public function create($userId, $name, $redirect, $provider = null, $personalAccess = false, $password = false, $confidential = true)
```
driesvints, line 112: This would definitely need to go to master because of the breaking change here.

src/ClientRepository.php
```diff
@@ -151,9 +152,9 @@ public function createPersonalAccessClient($userId, $name, $redirect)
      * @param  string  $redirect
      * @return \Laravel\Passport\Client
      */
-    public function createPasswordGrantClient($userId, $name, $redirect)
+    public function createPasswordGrantClient($userId, $name, $redirect, $provider = null)
```
driesvints, line 155: Needs to be added to the DocBlock above.

src/Console/InstallCommand.php
```diff
@@ -31,6 +31,6 @@ public function handle()
     {
         $this->call('passport:keys', ['--force' => $this->option('force'), '--length' => $this->option('length')]);
         $this->call('passport:client', ['--personal' => true, '--name' => config('app.name').' Personal Access Client']);
-        $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client']);
+        $this->call('passport:client', ['--password' => true, '--name' => config('app.name').' Password Grant Client', '--provider' => array_keys(config('auth.providers'))[0]]);
```
driesvints, line 34: Is it sensible to pick the first one here?

src/Guards/TokenGuard.php
```diff
@@ -82,6 +82,17 @@ public function __construct(ResourceServer $server,
...
 
+    /**
+     * Determine if the requested provider matches the client's provider.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return bool
+     */
+    protected function validateProvider(Request $request)
+    {
+        return $this->provider == $this->client($request)->getProvider();
```
driesvints, line 93: Best to use strict comparison here.

```suggestion
        return $this->provider === $this->client($request)->getProvider();
```

src/Guards/TokenGuard.php
```diff
@@ -90,6 +101,10 @@ public function __construct(ResourceServer $server,
      */
     public function user(Request $request)
     {
+        if (!$this->validateProvider($request)) {
```
driesvints, line 104:
```suggestion
        if (! $this->validateProvider($request)) {
```

database/migrations/2016_06_01_000004_create_oauth_clients_table.php
billriess: Yeah, most likely. Sorry, I didn't realize it was against the tagged release.

driesvints: No sorry needed. Feel free to send it to master :)

src/Guards/TokenGuard.php
billriess: Strict comparison won't work since we may be comparing different instances of the same object.

driesvints: @billriess aren't we comparing strings here?

src/Console/InstallCommand.php
billriess: Maybe not, I wasn't really sure what the best solution was. We could hardcode the default to "users" since that ships with Laravel. The right approach might be to attempt to use "users" if its an actual provider, otherwise default to null.

Thoughts?

src/Guards/TokenGuard.php
billriess: No, we're comparing implementations of Illuminate\Contracts\Auth\UserProvider.

src/Console/InstallCommand.php
driesvints: I can follow that.

billriess:
I fixed DocBlocks and StyleCI suggestions as well as updated the way the provider defaults when using the artisan command.

I will take a look at the failed tests later. I encourage anyone interested in this feature to review it and make suggestions as needed.

taylorotwell:
Can someone describe what "multi-provider" even means?

lucadegasperi:
@taylorotwell inside the `config/auth.php` you can define multiple `providers`.
The possibility to choose a provider is not currently exposed by passport.
Since adding request parameters to an oAuth2 call is generally not recommended, one possible way and probably the easiest due to the underlying league package, is to be able to tie a provider to an oauth 2 client.

billriess:
@taylorotwell as @lucadegasperi has explained the current implementation Passport is fairly restrictive when using things like password grants. This PR allows us to define what provider should be used at the oauth_client level. This way, we can have separate oauth_clients for different providers, which trickles down to different models.

This also fixes a potential security concern as Passport defaults to passing in the provider from the request's guard. This means the Authenticable returned by Passport would be whatever middleware guard you had in place on the route - allowing a `User` token to retrieve a `Customer` if their IDs match. With the PR it checks the guard's UserProvider with the oauth_client's UserProvider (if set) and does not return the Authenticatable if they are different.

billriess:
I've just pushed another update that fixes some of the tests. I'm a little stumped on the remaining failed tests as they are related to the way the PSR is mocked. If anyone has some ideas on how I can get those last tests to pass I would appreciate it.

billriess:
Fixed the validation on the TokenGuard to fall back properly if a provider is not set on the oauth_client.

src/Guards/TokenGuard.php
```diff
@@ -82,6 +82,24 @@ public function __construct(ResourceServer $server,
...
+    protected function validateProvider(Request $request)
+    {
+        $client = $this->client($request);
+
+        // If not client provider is defined, fallback to old behavior.
+        if ($client && empty($client->getProvider())) {
+            return true;
+        }
+
+        return $client && $this->provider == $client->getProvider();
```
driesvints, line 100: I still think we should a different comparison here instead of comparing two objects but I'm not sure what exact. Can we compare string names from the providers?

src/Guards/TokenGuard.php
```diff
@@ -82,6 +82,24 @@ public function __construct(ResourceServer $server,
         $this->encrypter = $encrypter;
     }
 
+    /**
+     * Determine if the requested provider matches the client's provider.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return bool
+     */
+    protected function validateProvider(Request $request)
```
driesvints, line 91:
```suggestion
    protected function hasValidProvider(Request $request)
```

src/Guards/TokenGuard.php
```diff
@@ -90,6 +108,10 @@ public function __construct(ResourceServer $server,
      */
     public function user(Request $request)
     {
+        if (! $this->validateProvider($request)) {
```
driesvints, line 111:
```suggestion
        if (! $this->hasValidProvider($request)) {
```

src/Client.php
```diff
@@ -49,7 +50,7 @@ class Client extends Model
     public function user()
     {
         return $this->belongsTo(
-            config('auth.providers.'.config('auth.guards.api.provider').'.model')
+            config('auth.providers.'.$this->provider ?: config('auth.guards.api.provider').'.model')
         );
```
driesvints, line 54:
```php
         $provider = $this->provider ?: config('auth.guards.api.provider');

         return $this->belongsTo(
             config("auth.providers.$provider.model")
         );
```

src/Guards/TokenGuard.php
billriess: I would like the do a more strict comparison here as well but I don't really have a good idea on how to get the name of the provider. The TokenGuard receives the requested UserProvider already and there isn't a string to pull from on the UserProvider class to my knowledge.

billriess: One solution I guess would be to rewrite the TokenGuard constructor to accept the string instead of the actual UserProvider then convert it to a UserProvider inside the TokenGuard based on the oauth_client provider. This would touch more code/tests though.

driesvints: Hmm, I think it's a simple as this?

```suggestion
        return $client && $this->provider === $client->provider;
```

billriess: I think this is what we discussed before, using `===` wouldn't work here because we would be comparing instances of the class. We could have different instances of the same class here so it would fail.

`$this->provider` is a `UserRepository` passed into the `TokenGuard` from the PassportServiceProvider based on the `RequestGuard`.
`$client->provider` is a string defined on the `oauth_client`.
`$client->getProvider()` could be a `UserRepository` but it would be its own instance.

driesvints: I see, thanks for clarifying.

I'm actually uncertain why this check is needed? Wouldn't this request prevent any authentication on clients that have a different provider than the default one set through the config? Maybe I'm not entirely following what's going on.

billriess: So let me try to clarify.

When the PassportServiceContainer loads the TokenGuard it is passing in the UserProvider from the RequestGuard. This is sort of the crux of the problem as this can allow a token created by a Customer to receive a User instance.

What I'm doing is allowing that UserProvider to still be passed in but instead of just trusting it and returning a user from it, I check if the oauth_client has a provider defined. If a provider is defined on the oauth_client then we use that one and compare it (loosely) to the one passed in from the constructor. If there is a mismatch then no user is returned and thus you are unauthenticated. If no provider is defined on the oauth_client then everything continues to work as it always has.

driesvints: But am I right that `$this->provider` only holds **the default provider set from the config**? This means that no client with a different custom provider will get through the check? Maybe I'm still misunderstanding something.

billriess: I did not change the way the default provider is passed into the TokenGuard. The UserProvider getting passed in has no context of the Laravel\Passport\Client. So it is passing in whatever provider the RequestGuard is using. Since this can be a mismatch to the one now defined on the Laravel\Passport\Client we need to check it inside the TokenGuard.

I can setup a demo to better showcase the problem if you want. I'm also available to discuss on discord if that works for you.

hamedov93: @billriess
I think there is another way to validate the provider by comparing the table of client provider against the table of user entity retrieved using oauth_user_id

$providers = config('auth.providers');
$clientProvider = $providers[$client->provider];
$table = $clientProvider['table'] ??
         (new $clientProvider['model'])->getTable();

And compare this table against that of the already retrieved user.

billriess: I think we're talking about different things...

You can look at the demo here:
https://github.com/billriess/passport-test

There is a sqlite db that you need to point to in your .env file.

Then you can make a REST call to /oauth/token using:
```js
{
    "grant_type": "password",
    "client_id": "1",
    "client_secret": "secret",
    "username": "user@test.test",
    "password": "password"
}
```
and
```js
{
    "grant_type": "password",
    "client_id": "2",
    "client_secret": "secret",
    "username": "customer@test.test",
    "password": "password"
}
```

Now use those tokens and hit `/api/user` or `/api/customer`. You'll see that `auth()->user()` will always return a `User` for the first endpoint and a `Customer` for the second endpoint even if you're using a token from the other model. The token represents who is authenticated so you shouldn't get an instance of `User` if you're token originated from `Customer`.

hamedov93: I know what you are talking about because I've addressed the same problem in my package.
See the auhtenticateViaBearerToken here:
https://github.com/hamedov93/passport-multiauth/blob/master/src/Guards/TokenGuard.php

billriess: Yeah, I guess it just seems odd to me to retrieve the user at all if it's not the right user.

hamedov93: The user will be retrieved any way for successful requests.

The cases when the provider doesn't match should be rare or suspecious, so another database query wouldn't be a big deal.

billriess: Yeah, I guess there are a bunch of different ways to tackle the problem. I think what I have is fairly clean though. We're still comparing the classes just not the instances.

driesvints: I think we should refactor the way the user is retrieved in general:

- Check if the client from the token has a provider
- If it has a provider, use that one to retrieve the user
- If not, use the default one which is passed in the token guard

That way you don't need to do any comparison.

driesvints: @billriess feel free to ping me on Discord in the passport channel

billriess: Ok, I'll look for you on discord.

I think always returning the user can be an issue though. If the two mismatch you want to not return any user because that means the token they are using is not the same as the one being protected by the route's guard.

billriess: @driesvints I think it's important to understand why different providers should not return any user. If you protected a route using the `auth:api:customers` middleware it would always return the provider for the customers. This, by default, would ensure that the user authenticated would be a customer. But, there is no check in place to make sure that the auth token actually belongs to the customer provider. So you end up in a scenario where a user auth token would return a customer instead of the expected user.

What I do is compare the 2 different providers. The provider on the RequestGuard is saying this must be a customer. Then inside the Passport TokenGuard I look at the the auth token's client and see if there is a provider defined. If there is, then we check if it matches the same class as the one from the RequestGuard. If they match, we can return the user because we know the user_id would come from the right provider. If they mismatch, then we have an authentication problem because you're trying to access a customer when you're token is not set up for that.

driesvints: I now fully understand what you mean. Thanks for your patience and explaining to well :-)

I think we should indeed pass the provider name to the token guard as well and do the comparison like that. That will be the only way to fully validate it. Comparing objects like this will lead to bugs I believe.

driesvints: We could add a `getName` method to the `UserProvider` contract but that would need a new major version of the framework. Passing the provider name to the TokenGuard is a good workaround.

billriess: Ok but this will end up with a fairly large amount of test rewrites because they are all expecting a provider to be passed so they currently anticipate the user.

A `getName` method would be great but yes that would require a framework modification to 8.x. I can push a PR for that as well if desired.

driesvints: @billriess I'll run this over with Taylor

billriess: Any update here? I don't mind making changes to the PR but it would be helpful to know what direction you or Taylor would like to take this in. This is one of the most requested/discussed features of Passport and as I see it, could be a potential security concern for some inexperienced developers.

@driesvints @taylorotwell

hamedov93: Actually this doesn't require a major release to the framework, as the provider can be reversely figured out using the model in EloquentUserProvider or the table in DatabaseUserProvider.
You already have a model instance or the table name, so you can fetch the provider from config that uses the model or the table name.

billriess: What if it uses something other than those two? I don't think it would be safe to assume those are the only two types of provider implementations.

hamedov93: There is a method that creates the provider class for each driver.
`createEloquentDriver` and `createDatabaseDriver`
If another driver is added there would be another method to be implemented and therefore passport will be updated.
I am not suggesting to use this approach forever, but to avoid the hassle of new major version for just a trivial need. But it should be added to the next release any way.

hamedov93: Looks like passport is using eloquent driver explicitly in the code so the is no need to check for database driver.
An easy solution for this would be like this:
```
$model = config('auth.providers.'.$client->provider.'.model');
if (class_exists($model))
{
    // Before calling retrieveById();
    $this->provider->setModel($model);
}
else
{
  // Other wise request should fail
}
```
I think this is enough :)

billriess:
```php
/**
     * Determine if the requested provider matches the client's provider.
     *
     * @return bool
     */
    protected function hasValidProvider(Request $request)
    {
        $client = $this->client($request);

        // If no client provider is defined, fallback to old behavior.
        if ($client && empty($client->getProvider())) {
            return true;
        }

        // Determine if the client's provider and the request's provider have matching models.
        return $client && $client->getProvider()->getModel() === $this->provider->getModel();
    }
```

How's this? Same outcome with strict comparison.

driesvints: @billriess can you give me push access to the PR? (see Discord)

billriess:
The PR has been updated to use decorator from @driesvints. I fixed the existing tests and pushed StyleCI suggestions.

taylorotwell:
It would be extremely helpful if someone could post a TLDR comment of why this is needed as well as a general explanation of how this implementation works, any breaking changes, any gotchas, etc.

billriess:
> It would be extremely helpful if someone could post a TLDR comment of why this is needed as well as a general explanation of how this implementation works, any breaking changes, any gotchas, etc.

@taylorotwell I updated the main post to better explain the issue and how this PR fixes it.

taylorotwell:
I'm confused. Where in this code does it actually use the Client model's provider to pull the model from the database. I see the PassportUserProvider provider name is compared to the clients. But PassportUserProvider's provider is always set to $config['provider']... which appears to be some static value and not controlled by the client being used at all. I guess I'm lost?

billriess:
> I'm confused. Where in this code does it actually use the Client model's provider to pull the model from the database. I see the PassportUserProvider provider name is compared to the clients. But PassportUserProvider's provider is always set to $config['provider']... which appears to be some static value and not controlled by the client being used at all. I guess I'm lost?

If the requested provider matches the one on the client then we know we can let them in and return the user. If there is a mismatch, we know that the token they are using is not setup for whatever provider you're trying to guard against.

`config/auth.php`
```php
...
    'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],
        'api' => [
            'driver' => 'passport',
            'provider' => 'users'
        ],
        'api:customers' => [
            'driver' => 'passport',
            'provider' => 'customers'
        ],
    ],

    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],
        'customers' => [
            'driver' => 'eloquent',
            'model' => App\Customer::class,
        ],
    ],
```
`routes/api.php`
```php
// This should only be accessible by Users...
Route::middleware('auth:api')->get('/user', function (Request $request) {
    return $request->user();
});

// This should only be accessible by Customers...
Route::middleware('auth:api:customers')->get('/customer', function (Request $request) {
    return $request->user();
});
```

taylorotwell:
Gotcha.

billriess:
#161 and #982 go more into detail to the problem and various ways its been discussed to resolve it.

taylorotwell:
Thanks. We'll want some documentation around this.

driesvints:
Thanks @billriess for all your work on this!

driesvints:
@ziming no, the column is nullable (see upgrade guide) and will take the default provider if it's set as null.

sajad2176:
how can we add this change to our project?

sajad2176:
I update the laravel/passport to v9.1 and add the customer provider to auth.php config and in the customer's routes if I run the 
```
dd($request->user());
```
it returns the customer's info and for user's routes it works fine and it return the user's info too but I think there is a bug about tokens
if I use user token for customer's  routes it doesn't return this response :
```
{
    "error": "Unauthenticated."
}
// 401 status 
```
It return customer's info again

billriess:
Did you set the provider on the client?

sajad2176:
actually my user types are : user , admin , courier and I set these config for auth.php :

```
    'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],
        'admin' => [
            'driver' => 'session',
            'provider' => 'admins',
        ],
        'api' => [
            'driver' => 'passport',
            'provider' => 'users',
            'hash' => false,
        ],
        'api:couriers' => [
            'driver' => 'passport',
            'provider' => 'couriers',
            'hash' => false,
        ],
    ],

    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],
        'admins' => [
            'driver' => 'eloquent',
            'model' => App\Admin::class,
        ],
        'couriers' => [
            'driver' => 'eloquent',
            'model' => App\Courier::class,
        ],
    ],
```

and for the user and courier's model I did like this :

```
namespace App;
use Laravel\Passport\HasApiTokens;
use Illuminate\Foundation\Auth\User as Authenticatable;

class Courier extends Authenticatable
{
    use  HasApiTokens;
}
```

and the api.php :
```
Route::prefix('courier')->name('courier.')->middleware('auth:api:couriers')->group(function () {

    Route::get('/' , function($request){
        dd($request->user());
    });

});

Route::prefix('user')->name('user.')->middleware('auth:api')->group(function () {

    Route::get('/' , function($request){
        dd($request->user());
    });

});
```

everything is fine but in postman i can use courier token for user's routes  and use user token for  courier's routes too

billriess:
In your database `oauth_clients` is the `provider` column set?

sajad2176:
no the provider column does not exist in the table and I ran the 
```
php artisan migrate:fresh --seed
```
and it didn't add to table by migration

driesvints:
@sajad2176 please read the upgrade guide: https://github.com/laravel/passport/blob/9.x/UPGRADE.md#support-for-multiple-guards

heyaj19:
@driesvints I still think  there is an issue - I have my providers set in auth.php, the provider column added. 

On LoginController with no middleware set, it uses users model. This is expected. 

When I change the guard to use      
```
            $this->middleware('guest:managersapi')->except('logout'); 
```

Its still using the users model not the managers model. 


In the auth.php file I have a guard 
```

    'managersapi' => [  //CURRENT GUARD ROUTE
                'driver' => 'passport',
                 'provider' => 'managers'
            ],
```

Am I missing something. I created a new client-id, client-secret with a provider column pointing to managers.

billriess:
@heyaj19 

Some things to note:

This PR covers password grants for bearer tokens. If you're not using that flow, it won't do anything new than it did previously.

You must have the provider defined to force that provider to be used. In your case, `managers` should be used in your oauth_clients table.

heyaj19:
@billriess I am using bearer tokens. This is my guzzle request after the client calls the login route 

`
$response = $client->post(config('app.PASSPORT_URL'), [
        'form_params' => [
            'client_id' => request('client_id'),
            'client_secret' => request('client_secret'),
            'scope' => '*',
            'username'=>request('email'),
            'password'=>request('password'),
            'grant_type' => request('grant_type')
        ],
    ]);`


And like I mentioned above, I do have a client_id, client_secret and provider for the managers table.

billriess:
@heyaj19 Glad to see you got it working.

gilles6:
This feature is working great. Thank you very much for making it.

However, from what I understand, it only works with password grant tokens access type (which is now [deprecated](https://oauth.net/2/grant-types/password/)).

Is there any chance that it is extended to the standard OAuth2 authorization code redirect flow ?

billriess:
@gilles6 Thanks, glad to see it's working well for you!

I don't see why this couldn't be extended to work with any other part of Passport. Given that the OAuth2 spec is moving away from Password grants we could see authorization code flow become more used. I don't have a lot of free time at the moment but I can keep this on my things to look at once I get some time.

gilles6:
@billriess That would be great :-) Let me know and I'll be one of your first testers!

guianzollin:
> ```
> api:customers
> ```

How can I do in my login method? Because `Auth::guard('api:customers')->attempt($validated)` and `Auth::guard('api:customers')->user()->createToken('authToken')` do not work.

"message": "Method Illuminate\\Auth\\RequestGuard::attempt does not exist.",

ktmsulaim:
@guianzollin `Auth::attempt` only works with driver set to session in `auth.php`

gilles6:
@billriess
Is there any update on implementing multiple providers with other part of Laravel Passport than Password Grant Token ?

tiggidoo:
I have same problem. multi-auth only work with session.If that the token does not work.

PR #1231, opened 2020-04-30
[8.x] Automatic configuration of client uuids

taylorotwell:
This PR adds a new `--uuids` flag that can be used when calling `passport:install`. This flag will automatically configure Passport for using UUIDs for client IDs instead of auto-incrementing IDs. This should not be a breaking change.

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -17,4 +17,17 @@ return [
 
     'public_key' => env('PASSPORT_PUBLIC_KEY'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Client UUIDs
+    |--------------------------------------------------------------------------
+    |
+    | By default, Passport uses auto-incrementing primary keys when assigning
+    | IDs to clients. However, if Passport is installed using the provided
+    | --uuids switch, this will be set to "true" and UUIDs will be used.
+    |
+    */
+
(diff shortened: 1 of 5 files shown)


billriess:
This would be a nice addition, my only concern is how we handle the UUID in the database. It would be most efficient to store the UUID as binary but I think that can get a little messy depending on database type.

If that adds too much complexity, then I would suggest using `Str::orderedUuid()` instead of `Str::uuid()`.

taylorotwell:
I thought about using `orderedUuid` but worried people would potentially have the same fears about those as auto-incrementing IDs.

billriess:
I think the main reason for making the `id` not incremental is that it's easily guessable. Even if the UUID is ordered, it is still mostly random - I could see how some people would want them to be strictly random and not ordered.

taylorotwell:
I don't think anyone really considers UUIDs to be a cryptographically secure random string. So, I don't think that this should be thought of as a "security" enhancement. The only real "benefit" I see is obscuring the amount of clients in the system.

billriess:
Yeah, it's more about obscurity.

Tom Scott has a good video somewhat covering this concept, https://www.youtube.com/watch?v=gocwRvLhDf8

Issue #1125, opened 2019-11-22
Client credentials middleware should allow any sort of valid/existing client


JuanDMeGon:
- Passport Version: 8.0.1
- Laravel Version: 6.5.2
- PHP Version: 7.3.11
- Database Driver & Version: Any driver and version

### Description:
Currently, it is not possible to get access to any action protected by the ClientCredentials middleware (CheckClientCredentials) if using a password or a personal client. BUT it should be possible, as those clients still being valid clients.

Now, even when it was discussed and seems to be a lot of confusion about this. The only job for the CheckClientCredentials middleware is to validate if the client is a valid client; does not matter the type of client it is.

Based on the theory and official standards of OAuth2: "The Client Credentials grant is used when applications request an access token to access their own resources, not on behalf of a user." ([REF1](https://oauth.net/2/grant-types/client-credentials/), [REF2](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)). So, basically you use the CheckClientCredentials to protect the route and validate if the request comes from a valid client (independently of the type of client). In fact, a token obtained using the client credentials grant does not have a user associated with it, so any user-related action is going to fail.

It was introduced in #1040 with the call to `$token->client->firstParty()` ([source](https://github.com/laravel/passport/pull/1040/files#diff-0c3d06260bd30e05008f5b46e3acf3e6R87)).

**In conclusion**: If ANY client obtains an access token using the client credentials grant it SHOULD be authorized when making a request to an endpoint protected with the CheckClientCredentials middleware and currently it is not working that way.

### Steps To Reproduce:
- Create a route protected with the CheckClientCredentials middleware
- Obtain a valid access token with a password or personal client (any of those)  & using the grant_type: client_credentials
- Send a request using that access token to the route created at the first step and it will fail with "unauthorized" response. BUT, again, it should be allowed as it is a valid client AND successfully obtained an access token.

Thanks in advance. I hope it helps to resolve the current issues with this.

alvirbismonte:
Just to add up to this issue, it should be either a Personal Access Client or Password Grant Client to replicate the Unauthorized response and not simply a Client Credentials Grant.

JuanDMeGon:
> Just to add up to this issue, it should be either a Personal Access Client or Password Grant Client to replicate the Unauthorized response and not simply a Client Credentials Grant.

Thanks. I fixed that now to get it more clear.

driesvints:
[...]

PR #1234, opened 2020-05-04
Patch 1

TiDJ:
Last december a change was made to allow any valid client :

Based on the theory and official standards of OAuth2: "The Client Credentials grant is used when applications request an access token to access their own resources, not on behalf of a user." (REF1, REF2).

Shouldn't this change be persistant ?

(taylorotwell merged commit on 5 Dec 2019)

Ref
https://github.com/laravel/passport/issues/1125
https://github.com/laravel/passport/pull/1132


driesvints:
This has been fixed already on master. Passport 9 will be released tomorrow.

PR #1255, opened 2020-05-08
[9.x] Allow to change Models database connection

antonkomarev:
This PR introduces configuration for model connection. It is working the same way as [Laravel Telescope Connection Config](https://github.com/laravel/telescope/blob/3.x/config/telescope.php#L47-L52).

This required when your database with OAuth server is working not on default connection.

Not a breaking change because models use `$connection` property as a fallback value.

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -46,4 +46,21 @@ return [
         'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET'),
     ],
 
+    /*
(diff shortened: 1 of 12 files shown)


driesvints:
@antonkomarev can you rebase? We did a change to the passport config.

antonkomarev:
Surе, will do it tomorrow

antonkomarev:
Surе, will do it tomorrow

antonkomarev:
@driesvints done

PR #1261, opened 2020-05-08
Match name of personal access token methods to config

aaronhuisinga:
The config values included in passport.php don't match those shown in the upgrade guide for personal access tokens in v9.1. This matches the values with those in the config file.


driesvints:
Thanks!

Issue #1270, opened 2020-05-13
passport:hash command double hashes secrets


corbosman:
- Passport Version: 9.2.0
- Laravel Version: 7.11.0
- PHP Version: 7.4.5
- Database Driver & Version: postgres 12.2

### Description:

Passport 9.0 introduced hashed client passwords. To help in migration an artisan command passport:hash was added.  Unfortunately this command does not work as expected.  Because 'Passport::$hashesClientSecrets' has to be on before you run it, both the command code, and Laravel's Client model secret setter hash the secret, resulting in a double hash. 

In the command class HashCommand.php:

```
$client->forceFill([
    'secret' => password_hash($client->secret, PASSWORD_BCRYPT),
])->save();
```

Then in the Client model:
```
    public function setSecretAttribute($value)
    {
        $this->plainSecret = $value;

        if (is_null($value) || ! Passport::$hashesClientSecrets) {
            $this->attributes['secret'] = $value;

            return;
        }

        $this->attributes['secret'] = password_hash($value, PASSWORD_BCRYPT);
    }
```

Probably the most logical fix would be for the command to not hash it and let the model hash it.

### Steps To Reproduce:

- start with unhashed client secrets
- run ```artisan passport:hash```
- secrets wont work because they're double hashed.

driesvints:
Heya, thank you for reporting! I just pushed a commit and new release that fixes this.

PR #1271, opened 2020-05-14
fix double hashing of secrets

corbosman:
The artisan command passport:hash does not need to hash the secret before updating the model, as the Client model itself already hashes the secret through a setter, resulting in a double hash and possibly loss of all secrets. This patch removes the hashing from the command.  More info in issue https://github.com/laravel/passport/issues/1270


driesvints:
Heya, I just pushed a commit and new release that fixes this. Thanks!

PR #1283, opened 2020-05-26
[9.x] Added Cookie Based Multi Auth

RahulDey12:
Added the multi auth support in the `CreateFreshApiToken` middleware.

> Currently `createToken` is not saving any provider information. But I've added a `provider` key to JWT then it is verifying in the cookie provider on `hasValidCookieProvider` method 

Just you need to add the `CreateFreshApiToken` middleware to the `$routeMiddleware` Like this:
```
protected $routeMiddleware = [
        . . .
        'api_token' => \Laravel\Passport\Http\Middleware\CreateFreshApiToken::class,
    ];
```

and add to the route middleware
```
Route::group(['middleware' => 'api_token:guard-name'], function() {
    // Your Routes Here
});
```

> It will also work with `web` middlewareGroups as well (This is gonna be web guard by default)

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -101,6 +101,21 @@ class TokenGuard
     }
 
     /**
+     * Determine if the request has valid provider in Cookie.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return bool
+     */
+    protected function hasValidCookieProvider(Request $request)
+    {
+        if($token = $this->getTokenViaCookie($request)) {
+            return $token['provider'] === $this->provider->getProviderName();
+        }
+
+        return false;
+    }
+
+    /**
      * Get the user for the incoming request.
      *
      * @param  \Illuminate\Http\Request  $request
@@ -222,6 +237,10 @@ class TokenGuard
     protected function authenticateViaCookie($request)
     {
         if (! $token = $this->getTokenViaCookie($request)) {
+            return;
+        }
+
+        if(! $this->hasValidCookieProvider($request)) {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


RahulDey12:
Are the controllers for test purposes only?

src/Guards/TokenGuard.php
```diff
@@ -100,6 +100,21 @@ protected function hasValidProvider(Request $request)
         return $client && $client->provider === $this->provider->getProviderName();
     }
 
+    /**
+     * Determine if the request has valid provider in Cookie.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return bool
+     */
+    protected function hasValidCookieProvider(Request $request)
```
RahulDey12, line 109: This Function is validating the provider information of the `Cookie`

driesvints:
We can't merge all this into the current stable release because of the breaking changes.

RahulDey12:
I will fix the tests

RahulDey12:
For this line
`$provider = config("auth.guards.{$guard}.provider") ?: config('auth.guards.web.provider');`
I am getting error on Test
```
Target class [config] does not exist.
```
Any Idea 🤔

taylorotwell:
Feel free to submit with passing tests and more explanation.

billriess:
I would not suggest modifying the cookie like this, there is no need to expose the provider in the cookie. It's also non-standard and adds another "verify provider" method which feels less DRY.

RahulDey12:
@billriess why is that ?

muarachmann:
@RahulDey12 , @billriess I was wondering if this was somehow implemented. Trying to consume my API with Javascript but I am using multi-auth and thus keeps returning me 401

RahulDey12:
@muarachmann I would suggest you to use [Sanctum](https://github.com/laravel/sanctum).

billriess:
Yeah sorry, I haven't used Passport much recently. We moved our project to use Sanctum instead.

Issue #1276, opened 2020-05-19
Passport's client secret hashing should use the main application's config for choosing hashing algorithm.


khan-zia:
Quoting from the upgrade guide of passport from v8.0 to v9.0 **"Client secrets may now be stored using a Bcrypt hash."**, It would seem that Passport's client secret hash uses Bcrypt instead of looking up what the main application's configured hashing algorithm is. If this is the case, then this must change as so many people out there including me would wanna use **argon2 and argon2id** instead of Bcrypt.

driesvints:
Sure. Feel free to pr.

khan-zia:
Sure. I will, over this weekend. Thanks!

PR #1284, opened 2020-05-26
[9.x] Use the Hash facade for hashing secrets to utilize main app's hashing…

khan-zia:
The client secrets are currently being hashed using *bcrypt* via PHP's *password_hash()*  function. Since there's already a nice *Hash* facade built into laravel, this should be used for the following end user benefit. It addresses #1276 
- It will allow the users to configure their hashing algorithm for client secrets via their main app's config.

### DIFF
--- a/src/Bridge/ClientRepository.php
+++ b/src/Bridge/ClientRepository.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Passport\Bridge;
 
+use Illuminate\Support\Facades\Hash;
 use Laravel\Passport\ClientRepository as ClientModelRepository;
 use Laravel\Passport\Passport;
 use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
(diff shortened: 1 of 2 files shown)


khan-zia:
@taylorotwell Correct! I believe you didn't take a look at the associated issue tagged in the description? I don't want my client secrets to be hashed by *bcrypt* but *argon2id*. How do I do that NOW (let's not wait for PHP password_hash) unless I completely implement my own way which will be impossible unless I extend many classes and interfaces and stuff from the core passport. Also Laravel can always make necessary updates in the *Hash* wrapper anyway in case PHP comes up with something cool with its core *password_hash*

khan-zia:
@driesvints It would be nice to have your input as well on this?

driesvints:
@khan-zia I'm guessing this is a no-go, sorry.

khan-zia:
Alright! However, I am convinced that it's worth it for the following reason that I would really like you to review @taylorotwell @driesvints . Also, I am leaving these comments here for future reference and others that might stumble upon it.
- Let's say in the future PHP comes up with something like `password_hash($val, BEST_ALGO)` or removing second param altogether, then you would need to update this at the `Hash` facade as well as at passport or any other Laravel packages for that matter?
- Since `Hash` is using the exact same thing under the hood, then why introduce even the slightest inconsistency into such a huge ecosystem?
- By using `Hash` , there's no new dependency being introduced to the passport package. `Hash` facade exists anyway?

PR #1285, opened 2020-05-26
Added Cookie Based Multi auth

RahulDey12:
It is regarding PR #1283. 
I am encoding the `provider` key to the JWT and matching the provider it the time cookie verification on `TokenGuard` File


billriess:
I feel like adding the provider to the JWT cookie is non-standard. I'm not sure why we need to store it there when we can just look up the OAuth client and verify it that way as we do for the Bearer token.

driesvints:
I'm gonna close this PR because I agree with @billriess. As soon as I find some time I'll try to look into all of this.

RahulDey12:
> I feel like adding the provider to the JWT cookie is non-standard. I'm not sure why we need to store it there when we can just look up the OAuth client and verify it that way as we do for the Bearer token.

@billriess some time we just use laravel session-based authentication by default and use the API for some small tasks like the auto-saving post. In that case, we use the cookie-based authentication on API

PR #1294, opened 2020-06-11
[9.x] Improve `passport:install` command

baijunyao:
I open this PR for two reasons:
1. Here is the hope to change `client_uuids` to `true`, instead of change all `false` to `true`.
2. The user cannot know which configuration item was changed by this code.  

Thanks.


driesvints:
There's no other option with a boolean value: https://github.com/laravel/passport/blob/9.x/config/passport.php

baijunyao:
Yes, but in the future, we may add other options with a boolean value, and it is easy to miss this code.

driesvints:
@baijunyao that's true.

PR #1307, opened 2020-06-25
Fix provider migration

bhosie:
A new column was added to an existing table by editing a migration file from 2016. This makes it impossible to run a migration after upgrading and receive the new column.

We've been using Passport since version 1. After upgrading 7.x to 9.x, I was no longer able to create passport clients via artisan because this column is missing from the table. Running `php artisan migrate` yielded `nothing to migrate`. In the future please refrain from editing old migration files.


bhosie:
pr sent to wrong branch. closing and resubbing

taylorotwell:
Has conflicts.

PR #1308, opened 2020-06-25
check if column is missing and add it. Fixes issue with legacy instal…

bhosie:
A new column was added to an existing table by editing a migration file from 2016. This makes it impossible to run a migration after upgrading from an older version of Passport and receive the new column.

We've been using Passport since version 1. After upgrading 7.x to 9.x, I was no longer able to create passport clients via artisan because this column is missing from the table. Running `php artisan migrate` yielded nothing to migrate.

### DIFF
--- a/database/migrations/2020_06_25_000001_provider_column_when_table_exists.php
+++ b/database/migrations/2020_06_25_000001_provider_column_when_table_exists.php
@@ -46,7 +46,7 @@ class ProviderColumnWhenTableExists extends Migration
      */
     public function up()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Please read the UPGRADE.md file in the repository.

bhosie:
Wouldn't it be a nicer experience for users to just be able to run an existing migration provided by the package over having everyone downstream add one?

Not going to press it beyond here. Just thought it'd be helpful to others.

bolechen:
useful for me,thx @bhosie

PR #1311, opened 2020-06-30
Allow guzzle v7

ankurk91:
Hi,

Guzzle v7 stable is out now.

https://github.com/guzzle/guzzle/blob/master/UPGRADING.md


driesvints:
Thanks!

PR #1327, opened 2020-08-16
[9.x] Remove Guzzle dependency

X-Coder264:
For some reason Passport required Guzzle as a dependency even though the only Guzzle namespace mention in the entire codebase was in just one test (the Guzzle PSR 7 response class). So I've changed that to use the Nyholm PSR 7 response which is also required by Passport (and recommended by Symfony PSR HTTP bridge) and in doing so I've removed the entire Guzzle dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,6 @@
         "php": "^7.2",
         "ext-json": "*",
         "firebase/php-jwt": "^5.0",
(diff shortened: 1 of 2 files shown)


taylorotwell:
Ping @driesvints

driesvints (approved):
This seems to be okay. I have no idea why Guzzle is required either since there have been no usages of it in Passport (not even in the very first release).

PR #1330, opened 2020-08-20
[10.x] Use only one PSR 7 implementation

X-Coder264:
Passport currently requires two PSR 7 implementations which is weird. Laravel already started the migration to Nyholm PSR 7 (https://github.com/laravel/framework/pull/31028) so with this PR I'm dropping the dependency on the Laminas Diactoros PSR 7 implementation in favor of Nyholm PSR 7. The few docblocks that were explicitly stating that a Diactoros implementation would be returned have been changed to specify that a PSR 7 compliant equivalent is going to be returned which will allow us to change PSR 7 implementations if there's gonna be a need for it without causing breaking changes for the users of this library.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,9 +27,8 @@
         "illuminate/encryption": "^6.18.31|^7.22.4",
         "illuminate/http": "^6.18.31|^7.22.4",
         "illuminate/support": "^6.18.31|^7.22.4",
-        "laminas/laminas-diactoros": "^2.2",
         "league/oauth2-server": "^8.1",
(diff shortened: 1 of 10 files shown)


taylorotwell:
What are the breaking change implications?

X-Coder264:
The only breaking change here is for people who overrode the `protected function dispatchRequestToAuthorizationServer` method because before the Diactoros response was typehinted and now the PSR 7 interface is typehinted so they'd get an incompatible method signature error. However I consider that to be a very low impact change as I doubt that that particular method is frequently overridden in projects (if at all).

PR #1347, opened 2020-09-12
[10.x] ClientFactory Error

victorbuild:
UnitTest create Client use ClientFactory Error

```
 Passport::actingAsClient(
        Client::factory()->create()
    ); 
```

### DIFF
--- a/database/factories/ClientFactory.php
+++ b/database/factories/ClientFactory.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Laravel\Passport\Database\Factories;
+namespace Database\Factories;
 
 use Illuminate\Database\Eloquent\Factories\Factory;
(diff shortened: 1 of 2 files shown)


victorbuild:
no modify ClientFactory namespace copy to poject

`php artisan vendor:publish --tag=passport-factories`

 Error 
image

I don’t know how to modify it, but modifying the namespace of the ClientFactory file in the project can be executed correctly

taylorotwell:
@driesvints I think we need to use a separate client factory for our own tests and then have a stub factory with the proper namespace. I have also added a new method to the "HasFactory" trait to allow the model to define what factory it uses (newFactory method):

image

taylorotwell:
We will fix this with a different approach.

PR #1351, opened 2020-09-18
[11.x] Better observability on Token-cookie decoding problems

HenkPoley:
Fixes silently ignoring Token-cookie problems, such as:

https://github.com/laravel/framework/issues/34227

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -245,6 +245,10 @@ class TokenGuard
         try {
             $token = $this->decodeJwtTokenCookie($request);
         } catch (Exception $e) {
+            if (config('app.debug')) {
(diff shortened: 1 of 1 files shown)


HenkPoley:
Ah, you probably can't access .env / config() from Passport?

Anyways, the intent is clear and would have helped a lot.

HenkPoley:
With JetBrain's IDEA editors (e.g. PhpStorm) you can quickly find similar _silent try-catch_ problem spots with Edit > Find > 'Search Structurally...'

And this query:

`try { $a$ } catch (Exception $b$) { return; }`

Which will sadly find all `{ return .*; }` as well, which may include some proper handling.

In Passport this appears to be the only place.

driesvints:
You can't use facades there like that. Feel free to attempt again with a passing test suite. Thanks

PR #1354, opened 2020-09-24
actingAs should as well cover request

juslintek:
When $request->user() is used in code and Passport::actingAs to mock corrently logged in user. Tests fail, because $request->user() returns null, while Auth::user() returns user.


driesvints:
Please send this to 10.x

juslintek:
But I need this in 8.x as well.

driesvints:
@juslintek 8.x isn't maintained anymore. Only the latest version.

PR #1355, opened 2020-09-24
[10.x] actingAs should as well cover request

juslintek:
When $request->user() is used in code and Passport::actingAs to mock currently logged in user. Tests fail, because $request->user() returns null, while Auth::user() returns user.


taylorotwell:
I wouldn't really expect this to be necessary because the `AuthManager` set a user resolver to retrieve the current user from the guard, so I would expect setting the user on the guard to be sufficient?

image

taylorotwell:
Seems to work already for me as I spoke about above:

image

image

image

PR #1358, opened 2020-09-26
Token cookie should set in /login POST

gaodeng:


driesvints:
This breaks the test suite.

PR #1366, opened 2020-10-09
[8.x] Avoid decrypt exception if Mac is invalid

leandrogrando:
Avoid decrypt exception if Mac is invalid


driesvints:
This seems unwanted. You definitely want to be informed if the decryption fails I believe.

driesvints:
Also, this will catch any exception not just an invalid Mac.

leandrogrando:
@driesvints  This solved an issue I am experiencing in an application that is in production. I reported the problem here if you can take a look https://laracasts.com/discuss/channels/laravel/the-mac-is-invalid-only-in-some-cases-even-without-changing-app-keys. Do you perhaps have any other solution to help me?

PR #1368, opened 2020-10-13
[10.x] Cache tokens to prevent unnecessary queries

### LINKED ISSUE #382: TokenGuard query token twice
The following sql statement is queried twice during a call to `TokenGuard::authenticateViaBearerToken()` function:

```sql
select * from `oauth_access_tokens` where `oauth_access_tokens`.`id` = ? limit 1
```

Jefemy:
Fixes #382 by storing the tokens in memory after first retrieval. This is the only direct passport fix as the other queries come from league/oauth2-server and can't be stopped without passing the tokens around like with this.
It also removes the check for revoked client since the server already checks for it.

- [ ] Improve the way the service provider serves the same TokenRepository
- [ ] Add cacheing to the rest of the functions in the TokenRepository

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -172,13 +172,6 @@ class TokenGuard
 
         $clientId = $psr->getAttribute('oauth_client_id');
 
-        // Finally, we will verify if the client that issued this token is still valid and
-        // its tokens may still be used. If not, we will bail out since we don't want a
-        // user to be able to send access tokens for deleted or revoked applications.
-        if ($this->clients->revoked($clientId)) {
-            return;
-        }
-
         return $token ? $user->withAccessToken($token) : null;
     }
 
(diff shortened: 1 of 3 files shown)


src/PassportServiceProvider.php
```diff
@@ -284,10 +284,15 @@ protected function registerGuard()
     protected function makeGuard(array $config)
     {
         return new RequestGuard(function ($request) use ($config) {
+            $tokenRepo = $this->app->make(TokenRepository::class);
+           
             return (new TokenGuard(
-                $this->app->make(ResourceServer::class),
+                new ResourceServer(
```
driesvints, line 290: This will break apps that are overriding this class.

src/Guards/TokenGuard.php
```diff
@@ -172,13 +172,6 @@ protected function authenticateViaBearerToken($request)
 
         $clientId = $psr->getAttribute('oauth_client_id');
 
-        // Finally, we will verify if the client that issued this token is still valid and
-        // its tokens may still be used. If not, we will bail out since we don't want a
-        // user to be able to send access tokens for deleted or revoked applications.
-        if ($this->clients->revoked($clientId)) {
-            return;
-        }
-
```
driesvints, line 181: This is a breaking change.

driesvints:
Seems like there are still some fundamental breakages in this PR. Feel free to reattempt it when you've solved them. Thanks.

Issue #1376, opened 2020-11-19
Return sane error for bad credentials


jdavidbakr:
With the recent change that oauth2-server now throws an invalid_grant instead of invalid_credentials, the error message returned to the user is cryptic and misleading.  I propose that we modify `AccessTokenController::issueToken()` to a try-catch similar to https://github.com/laravel/passport/issues/1113#issuecomment-698813835

driesvints:
I'm not sure this is needed. I'll ping the OAuth2 server maintainer on your PR.

jdavidbakr:
oauth2-server has stated that they are not going to change it, that the message is intentionally obfuscated.

driesvints:
@jdavidbakr we in general follow the path that oauth2-server chooses which is why I want to verify it first with the maintainer. Let's see what he says.

Issue #1113, opened 2019-11-10
oauth2-server now throws an uncaught exception on incorrect credentials


patrickomeara:
- Passport Version: 8.0.0
- Laravel Version: 6.5.0
- PHP Version: 7.3.11
- Database Driver & Version: Mysql 8.0.16

### Description:
The league/oauth2-server 8.0 upgrade now throws 400 Bad Request instead of the previous 401 Unauthorized.

This is currently not handled gracefully in Passport and bubbles up to bug reporting.

@driesvints is this by design, and if so how/where should we catch the error?

### Steps To Reproduce:
* try to get a bearer token with incorrect login details

### Stacktrace

```
Caused by: League\OAuth2\Server\Exception\OAuthServerException The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client. 
    /src/vendor/league/oauth2-server/src/Exception/OAuthServerException.php:263 League\OAuth2\Server\Exception\OAuthServerException::invalidGrant
    /src/vendor/league/oauth2-server/src/Grant/PasswordGrant.php:107 League\OAuth2\Server\Grant\PasswordGrant::validateUser
    /src/vendor/league/oauth2-server/src/Grant/PasswordGrant.php:54 League\OAuth2\Server\Grant\PasswordGrant::respondToAccessTokenRequest
    /src/vendor/league/oauth2-server/src/AuthorizationServer.php:198 League\OAuth2\Server\AuthorizationServer::respondToAccessTokenRequest
    /src/vendor/laravel/passport/src/Http/Controllers/AccessTokenController.php:63 Laravel\Passport\Http\Controllers\AccessTokenController::Laravel\Passport\Http\Controllers\{closure}
    /src/vendor/laravel/passport/src/Http/Controllers/HandlesOAuthErrors.php:24 Laravel\Passport\Http\Controllers\AccessTokenController::withErrorHandling
    /src/vendor/laravel/passport/src/Http/Controllers/AccessTokenController.php:65 Laravel\Passport\Http\Controllers\AccessTokenController::issueToken
    /src/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php:48 Illuminate\Routing\ControllerDispatcher::dispatch
    /src/vendor/laravel/framework/src/Illuminate/Routing/Route.php:219 Illuminate\Routing\Route::runController
[...]

PR #1377, opened 2020-11-19
[10.x] Return 'auth.failed' message for bad password credentials

jdavidbakr:
Recently the oauth2-server package changed the failed credentials response to be a generic `invalid_grant` error to line up with OAuth2 spec.  However, the message sent by this package is intentionally cryptic.  Because Passport is simply passing this message back to the response, API clients no longer have a default message that makes sense to the end user.

This pull request catches this exception and returns the 'auth.failed' message in the `error_description` and `message` fields of the response.  It retains the `invalid_grant` and 400 status that is correct for the OAuth2 spec.

Resolves https://github.com/laravel/passport/issues/1376

### DIFF
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -7,6 +7,7 @@ use Lcobucci\JWT\Parser as JwtParser;
 use League\OAuth2\Server\AuthorizationServer;
 use Nyholm\Psr7\Response as Psr7Response;
 use Psr\Http\Message\ServerRequestInterface;
+use Laravel\Passport\Exceptions\OAuthServerException;
 
 class AccessTokenController
 {
@@ -41,10 +42,11 @@ class AccessTokenController
      * @param  \Lcobucci\JWT\Parser  $jwt
      * @return void
      */
-    public function __construct(AuthorizationServer $server,
-                                TokenRepository $tokens,
-                                JwtParser $jwt)
-    {
+    public function __construct(
+        AuthorizationServer $server,
+        TokenRepository $tokens,
+        JwtParser $jwt
+    ) {
         $this->jwt = $jwt;
         $this->server = $server;
         $this->tokens = $tokens;
@@ -58,10 +60,22 @@ class AccessTokenController
      */
     public function issueToken(ServerRequestInterface $request)
     {
-        return $this->withErrorHandling(function () use ($request) {
-            return $this->convertResponse(
-                $this->server->respondToAccessTokenRequest($request, new Psr7Response)
-            );
-        });
+        try {
+            return $this->withErrorHandling(function () use ($request) {
+                return $this->convertResponse(
+                    $this->server->respondToAccessTokenRequest($request, new Psr7Response)
+                );
+            });
+        } catch (OAuthServerException $e) {
+            if ($e->getCode() === 10) {
+                return response()->json([
+                    'error'=>'invalid_grant',
+                    'error_description'=>__('auth.failed'),
+                    'hint'=>'',
+                    'message'=>__('auth.failed')
+                ], $e->statusCode());
+            }
+            throw $e;
+        }
     }
 }
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -2,12 +2,12 @@
 
 namespace Laravel\Passport\Http\Controllers;
 
+use Laravel\Passport\Exceptions\OAuthServerException;
 use Laravel\Passport\TokenRepository;
 use Lcobucci\JWT\Parser as JwtParser;
 use League\OAuth2\Server\AuthorizationServer;
 use Nyholm\Psr7\Response as Psr7Response;
 use Psr\Http\Message\ServerRequestInterface;
-use Laravel\Passport\Exceptions\OAuthServerException;
 
 class AccessTokenController
 {
@@ -72,7 +72,7 @@ class AccessTokenController
                     'error'=>'invalid_grant',
                     'error_description'=>__('auth.failed'),
                     'hint'=>'',
-                    'message'=>__('auth.failed')
+                    'message'=>__('auth.failed'),
                 ], $e->statusCode());
             }
             throw $e;
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -44,7 +44,8 @@ class AccessTokenController
      */
     public function __construct(AuthorizationServer $server,
                                 TokenRepository $tokens,
-                                JwtParser $jwt) {
+                                JwtParser $jwt)
+    {
         $this->jwt = $jwt;
         $this->server = $server;
         $this->tokens = $tokens;
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -42,11 +42,9 @@ class AccessTokenController
      * @param  \Lcobucci\JWT\Parser  $jwt
      * @return void
      */
-    public function __construct(
-        AuthorizationServer $server,
-        TokenRepository $tokens,
-        JwtParser $jwt
-    ) {
+    public function __construct(AuthorizationServer $server,
+                                TokenRepository $tokens,
+                                JwtParser $jwt) {
         $this->jwt = $jwt;
         $this->server = $server;
         $this->tokens = $tokens;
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/AccessTokenController.php
```diff
@@ -41,10 +42,11 @@ class AccessTokenController
...
      */
-    public function __construct(AuthorizationServer $server,
-                                TokenRepository $tokens,
-                                JwtParser $jwt)
-    {
+    public function __construct(
+        AuthorizationServer $server,
+        TokenRepository $tokens,
+        JwtParser $jwt
+    ) {
```
driesvints, line 49: Please revert this change.

driesvints:
I'm not sure if this is needed. @sephster does this look okay to you? I don't see a specific reason why we'd need to format this OAuth2 Server exception?

jdavidbakr:
@driesvints - maybe I'm not handling bad password through passport correctly, but I have been sending back the message from the server to the client user.  I cannot find a way to catch this and translate this message server-side without doing some significant gymnastics, which means that by default this message must be interpreted by the client (whereas all other error messages can just be sent directly).  It took me quite a while to figure out what was going on when people started asking about the default cryptic message.

Sephster:
The error_description isn't meant to be shown to the end user. It is meant for the client developer only but often does get surfaced to the end user. For this reason. I've exercised caution on what is in the error description to avoid the aforementioned scenario and indeed, what appears to be happening here.

In the error description, we do mention that the _resource owner credentials_ could be the issue. If you are receiving this error when using the password grant, it is reasonable to assume this will be the problem.

Ideally what should happen is the client captures the exception and shows a user friendly message, not the error_description, to the end user. The client should know what it has just done so it should be possible to narrow down what the problem might be e.g.:

* used the authorization code grant - likely to be an invalid authorization code
* used the password grant - will be a resource owner credentials issue as this is the only scenario we use this error I believe
* use a refresh token - could have expired or been invalidated

At present, I don't think this change is needed. I think the concern is with the client rather than the server. I am looking into this though and if it turns out best practice is to be more specific, I will update the oauth2-server to reflect this.

jdavidbakr:
Thanks for the explanation.  My argument against that is that it is less plug-and-play on the client end, i.e. I have in most of my projects a class that handles all communication with the server and while I could catch this one error in every client application, every other error response is coming from the server and it seems that for consistency it would be cleaner to return an error message from the server about invalid credentials in the same way.

patrickomeara:
I would also like to see this change implemented. I originally authored #1113 along with https://github.com/patrickomeara/passport_error to demonstrate the issue with a vanilla install. 

It still seems intuitive that the recommendation is to catch this in the application layer, when it was previous handled by passport. I understand that the underlying oauth-server had to fall in line with the spec, but handling one specific error differently to how the others are handled is odd.

Passport at least needs to document as to where the exception should be caught, handled and translated if it is indeed to be in the application layer.

driesvints:
I'm sorry all, but I agree with @Sephster that this is something that should be resolved in the client. Just like exception messages in Laravel aren't meant to be displayed to end users (we often get PRs for those as well) this one isn't meant to be displayed to end users either. Please catch this exception in your own app to provide the message you'd like to display to your  users.

Sephster:
To be clear and avoid confusion, all exceptions issued from Passport which originate from oauth2-server are intended for the client developer, not just this one discussed. Apologies if I caused any confusion there.

patrickomeara:
I can see your point @driesvints, where do you recommend we catch it? That logic is pretty well tucked behind the `/token` route that passport provides.

The only application code in my stacktrace is in middleware. Is that the correct spot?

```
Caused by: League\OAuth2\Server\Exception\OAuthServerException The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client. 
    /src/vendor/league/oauth2-server/src/Exception/OAuthServerException.php:263 League\OAuth2\Server\Exception\OAuthServerException::invalidGrant
    /src/vendor/league/oauth2-server/src/Grant/PasswordGrant.php:107 League\OAuth2\Server\Grant\PasswordGrant::validateUser
    /src/vendor/league/oauth2-server/src/Grant/PasswordGrant.php:54 League\OAuth2\Server\Grant\PasswordGrant::respondToAccessTokenRequest
    /src/vendor/league/oauth2-server/src/AuthorizationServer.php:198 League\OAuth2\Server\AuthorizationServer::respondToAccessTokenRequest
    /src/vendor/laravel/passport/src/Http/Controllers/AccessTokenController.php:63 Laravel\Passport\Http\Controllers\AccessTokenController::Laravel\Passport\Http\Controllers\{closure}
    /src/vendor/laravel/passport/src/Http/Controllers/HandlesOAuthErrors.php:24 Laravel\Passport\Http\Controllers\AccessTokenController::withErrorHandling
    /src/vendor/laravel/passport/src/Http/Controllers/AccessTokenController.php:65 Laravel\Passport\Http\Controllers\AccessTokenController::issueToken
    /src/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php:48 Illuminate\Routing\ControllerDispatcher::dispatch
    /src/vendor/laravel/framework/src/Illuminate/Routing/Route.php:219 Illuminate\Routing\Route::runController
    /src/vendor/laravel/framework/src/Illuminate/Routing/Route.php:176 Illuminate\Routing\Route::run
    /src/vendor/laravel/framework/src/Illuminate/Routing/Router.php:680 Illuminate\Routing\Router::Illuminate\Routing\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:130 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    Http/Middleware/ThrottleRequests.php:23 App\Http\Middleware\ThrottleRequests::handle
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:171 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:105 Illuminate\Pipeline\Pipeline::then
    /src/vendor/laravel/framework/src/Illuminate/Routing/Router.php:682 Illuminate\Routing\Router::runRouteWithinStack
    /src/vendor/laravel/framework/src/Illuminate/Routing/Router.php:657 Illuminate\Routing\Router::runRoute
    /src/vendor/laravel/framework/src/Illuminate/Routing/Router.php:623 Illuminate\Routing\Router::dispatchToRoute
    /src/vendor/laravel/framework/src/Illuminate/Routing/Router.php:612 Illuminate\Routing\Router::dispatch
    /src/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php:176 Illuminate\Foundation\Http\Kernel::Illuminate\Foundation\Http\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:130 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/TransformsRequest.php:21 Illuminate\Foundation\Http\Middleware\TransformsRequest::handle
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:171 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/ValidatePostSize.php:27 Illuminate\Foundation\Http\Middleware\ValidatePostSize::handle
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:171 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/CheckForMaintenanceMode.php:62 Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::handle
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:171 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    /src/vendor/fideloper/proxy/src/TrustProxies.php:57 Fideloper\Proxy\TrustProxies::handle
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:171 Illuminate\Pipeline\Pipeline::Illuminate\Pipeline\{closure}
    /src/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php:105 Illuminate\Pipeline\Pipeline::then
    /src/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php:151 Illuminate\Foundation\Http\Kernel::sendRequestThroughRouter
    /src/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php:116 Illuminate\Foundation\Http\Kernel::handle
    /src/public/index.php:55 [main]
```

driesvints:
@patrickomeara a middleware is a good solution for this yeah 👍

PR #1387, opened 2020-11-26
[9.x] Support PHP 8.0

driesvints:
This provides support for PHP 8 so people on Laravel 6 LTS can still install Passport. ~Unfortunately this requires the dropping of PHP 7.2 support because OAuth2 Server v8.2 dropped it as well.~

~An alternative solution to the issues for people on Laravel 6 LTS and keep PHP 7.2 support is to constrain `lcobucci/jwt` to `<=3.3.` but that then we won't be able to offer PHP 8 support on Passport 9.~

~@taylorotwell I'll leave it up to you to decide on this.~

Thanks to @Sephster PHP 7.2 was re-added to OAuth2 Server 8 🎉

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.3, 7.4, 8.0]
+        php: [7.2, 7.3, 7.4, 8.0]
         laravel: [^6.0, ^7.0]
 
     name: P${{ matrix.php }} - L${{ matrix.laravel }}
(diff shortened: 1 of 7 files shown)


Sephster:
I dropped PHP 7.2 support because it is almost out of security release support. I think it only has a couple of days left before it is fully unsupported. However, there is no technical reason for this. If it is helpful, I can change the release to re-add PHP 7.2 for ease if it is going to cause you issues so there is another option for you if needed :+1:

driesvints:
@Sephster I think your choice for dropping PHP 7.2 is a sensible one and one we can also make here. Let's see what @taylorotwell says 🙂

velkymx:
This bug is cause lots of headaches going back to Laravel 6 LTS and is needed.

driesvints:
@velkymx you're not really contributing anything productive to this discussion. Please be patient until we're ready with this PR. We're close to merging and releasing this.

PR #1392, opened 2020-12-10
The `aud` is array in lcobucci/jwt v4

baijunyao:
lcobucci/jwt: https://github.com/lcobucci/jwt/blob/9c64c1739d9b8ceb1b919fbae3368ab6afef1d39/src/Token/Parser.php#L100
```
        if (array_key_exists(RegisteredClaims::AUDIENCE, $claims)) {
            $claims[RegisteredClaims::AUDIENCE] = (array) $claims[RegisteredClaims::AUDIENCE];
        }
```

RFC: https://tools.ietf.org/html/rfc7519#section-4.1.3


baijunyao:
Sorry, I was wrong, This `aud` of token is from cookie, but this token does not seem to contain `aud`, This may be another issue, I will close this PR first.

PR #1396, opened 2021-01-14
[10.x] Add purging of invalid refresh tokens to command

carestad:
The `passport:purge` command does a good job of removing revoked and expired tokens in the different tables, but we end up with lots of invalid refresh tokens pointing to non-existing access tokens. This PR adds purging of invalid refresh tokens to the command as well. Also added a test class for this command.

### DIFF
--- a/src/Console/PurgeCommand.php
+++ b/src/Console/PurgeCommand.php
@@ -22,7 +22,7 @@ class PurgeCommand extends Command
      *
      * @var string
      */
-    protected $description = 'Purge revoked and / or expired tokens and authentication codes';
+    protected $description = 'Purge revoked and / or expired tokens and authentication codes, and invalid refresh tokens.';
 
     /**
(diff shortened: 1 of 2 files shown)


Sephster:
@driesvints I just want to check if this is correct. In the oauth2-server it is fine to have a refresh token that isn't tied to an access token as it _could_ be used later.

It might be better to check the refresh token's expiry to clean up old refresh tokens if this is possible.

driesvints:
@Sephster yeah you're correct. I'm going to revert this PR as the query right above the one that was added should cover that use case already.

@carestad it seems purging these tokens based on this condition isn't valid (see Andrew's answer above).

carestad:
> @carestad it seems purging these tokens based on this condition isn't valid (see Andrew's answer above).

Oh wow, sorry about that! I should have done a better research job before assuming this

Issue #1409, opened 2021-02-17
Laravel Nova and Laravel Passport cannot install together, if done in the wrong oder.


addgod:
- Passport Version: 10.1.0
- Laravel Version: 8.28.1
- PHP Version: 7.4.15

### Description:
Cannot install Laravel Passport along side Laravel Nova, if done in the wrong order. So if nova is installed first then passport, it will not work. But if passport is installed first, and then nova. Then it will work.

Laravel Nova and Laravel Passport both uses phpseclib/phpseclib. Passport uses it directly, where nova it is quite deeply nested in its dependencies. But when installing both in a fresh install of laravel, composer complains because of the version mismash shown below.

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - laravel/passport[v10.1.0, ..., 10.x-dev] require phpseclib/phpseclib ^2.0 -> found phpseclib/phpseclib[2.0.0, ..., 2.0.x-dev] but the package is fixed to 3.0.5 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - Root composer.json requires laravel/passport ^10.1 -> satisfiable by laravel/passport[v10.1.0, 10.x-dev].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.

```

### Steps To Reproduce:
Install Laravel.
Install Laravel Nova
Install Laravel Passport

driesvints:
What version of Nova are you using? I'll see if I can get passport to update to phpseclib/phpseclib v3.0

driesvints:
We'll do a new release next Tuesday for phpseclib v3 support

PR #1410, opened 2021-02-18
[10.x] Update to phpseclib v3

driesvints:
There is technically a breaking change here for people people overwriting the `handle` method of the `KeysCommand`. I had to remove it because `RSA` is an abstract class in v3. I believe in reality there aren't many people who would be overwriting this command. phpseclib was first introduced in https://github.com/laravel/passport/pull/72 which adds the `RSA` class to the handle method but it's unclear why that was done. I cannot imagine anyone would want to swap it out.

Closes https://github.com/laravel/passport/issues/1409

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "league/oauth2-server": "^8.2",
         "lcobucci/jwt": "^3.4|^4.0",
         "nyholm/psr7": "^1.3",
-        "phpseclib/phpseclib": "^2.0",
+        "phpseclib/phpseclib": "^3.0",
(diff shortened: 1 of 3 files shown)


smujaddid:
I suggest to revert this for 10.x because if there are any future updates for 10.x then anyone, who are using league/flysystem-sftp 1.x and Laravel 8.x, will not be able to update their laravel/passport package because the flysytem and flysystem-sftp still using phpseclib v2 in their libraries (both 1.x and 2.x versions).

Unless there is a way to support both phpseclib v2 and v3, then this could be a problem.

PR #1411, opened 2021-02-20
[10.x] Add created_at and updated_at to date-mutator

MrKrisKrisu:
Unfortunately, the two variables `created_at` and `updated_at` are not automatically mutated. I could address the `expires_at` variable normally with Carbon, the other two not, because they were only a string.

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -50,7 +50,7 @@ class Token extends Model
      * @var array
      */
     protected $dates = [
-        'expires_at',
+        'created_at', 'updated_at', 'expires_at',
     ];
 
     /**


taylorotwell:
This shouldn't be needed? The Eloquent base model merges created_at and updated_at into the list of date castable attributes:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1091

taylorotwell:
Hmm - I guess it is because this model has `timestamps = false`. Changing this now is a breaking change unfortunately since applications may already be working around this behavior 😢

driesvints:
This has been a long outstanding issue which was rejected a couple of times in the past:

https://github.com/laravel/passport/issues/786
https://github.com/laravel/passport/pull/1075

You can still adjust this by overriding the models as I explained here: https://github.com/laravel/passport/issues/786#issuecomment-557521784

Issue #1370, opened 2020-10-22
Passport migrations ignore --database option


magoun:
- Passport Version: 9.3.2
- Laravel Version: 7.28.3
- PHP Version: 7.4.11
- Database Driver & Version: MySQL 5.7.31

### Description:
The included passport migrations no longer consider the --database flag passed to migrate on the command line, instead reading the default database from the connection specified in config/passport.

### Steps To Reproduce:
On a new Laravel project with at two database connections in config/database (mysql and testing), run `php artisan migrate:fresh` then run `php artisan migrate:fresh --database testing`.

The second migration will fail, reporting that the oauth_auth_codes_table already exists, as it is accessing the `mysql` connection instead of the `testing` connection.

I believe this behavior was introduced in 555e02c on May9th.

driesvints:
I'm investigating this one with the team. Thanks for reporting 👍

driesvints:
We're gonna look at how we're going let the `--database` flag take over precedence here.

tecbeast42:
I can reproduce this problem in laravel 8.

driesvints:
Yeah still haven't gotten to this I'm afraid. Have been knee deep into PHP 8 updates. Appreciating help with it.

CaptainBart:
Also running into this issue. Is there a workaround (except running the tests on the default database) available to avoid this issue?

driesvints:
Not atm.

magoun:
One workaround is to manually change the database config item to your testing database before running tests. If you have a database connection set up in Laravel called `'testing'`, edit the below section in `config/passport.php`:

After running tests, be sure to uncomment the production database setting and comment out the 'testing' line.

```
/*
    |--------------------------------------------------------------------------
    | Passport Storage Driver
    |--------------------------------------------------------------------------
    |
    | This configuration value allows you to customize the storage options
    | for Passport, such as the database connection that should be used
    | by Passport's internal database models which store tokens, etc.
    |
    */

    'storage' => [
        'database' => [
            'connection' => 'testing',
            // 'connection' => env('DB_CONNECTION', 'mysql'),
        ],
    ],
```

CaptainBart:
I do something similar, but I prefer to change my .env file, since it won't be comitted to git. Locally I change the .env DB_DATABASE setting to the test database when I want to create/run unit tests and change it back when I'm done testing. For the build pipeline I've created a dedicated test .env file that is copied/renamed before the tests are run.

micc83:
[...]

PR #1412, opened 2021-02-22
[11.x] Revert model DB connection customization

driesvints:
This PR reverts https://github.com/laravel/passport/pull/1255. In hindsight that probably wasn't the best choice as it's [breaking some default Laravel behavior like the `--database` flag on the migration command](https://github.com/laravel/passport/issues/1370). This also causes Passport to differ from any other Laravel first party package (except Telescope). 

The problem that the PR wanted to solve can better be solved imo by overwriting the default models as explained here: https://laravel.com/docs/8.x/passport#overriding-default-models. In combination with [customizing the migrations](https://laravel.com/docs/8.x/passport#migration-customization) this already allows people to do whatever they want. I believe this is a more sensible approach to also keep the package straightforward and the code clean. I believe that not much people actually need https://github.com/laravel/passport/pull/1255. https://github.com/laravel/passport/pull/1255 wasn't documented anyway atm.

Targeting the next major Passport release because of the breaking change.

Ping @antonkomarev 

Fixes https://github.com/laravel/passport/issues/1370

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -46,21 +46,4 @@ return [
         'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET'),
     ],
 
-    /*
-    |--------------------------------------------------------------------------
-    | Passport Storage Driver
-    |--------------------------------------------------------------------------
-    |
-    | This configuration value allows you to customize the storage options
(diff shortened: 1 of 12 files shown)


jonnycbw:
I see, thanks for clarifying!

PR #1429, opened 2021-03-23
Auth Query optimization for api calls

AgonAziri:
While logging database queries I saw that there were 5 queries such as:

- 2 queries to oauth_clients table
- 3 queries to oauth_access_tokens table (2 of which were made by league/oauth2-server package)

In order to reduce number of queries some changes in code where made:
- revoked function in ClientRepository now will accept client as parameter instead of ID (this removes 1 query to oauth_clients table). Client object that is passed to this function is returned from findActive function in ClientRepository.
-  getPsrRequestViaBearerToken function in TokenGuard that returns some result from league/oauth2-server now will be passed as parameter to hasValidProvider and from there to client function in TokenGuard (this process removes 1 query from 3 total queries made to oauth_access_tokens table)

Changes that were made in this pull request will remove 2 queries out of 5, bringing down number of total queries on each request from 5 to 3.

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -224,10 +224,8 @@ class ClientRepository
      * @param  int  $id
      * @return bool
      */
-    public function revoked($id)
+    public function revoked($client)
     {
-        $client = $this->find($id);
-
         return is_null($client) || $client->revoked;
     }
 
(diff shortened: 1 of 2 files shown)


driesvints:
As you can see this breaks the test suite. Please resubmit this with passing tests.

PR #1443, opened 2021-05-03
Add prompt=login parameter option

mikeschoneveld:
Add possibility to work with the prompt=login parameter so the user has to login again based on the OpenID specs:

https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -58,6 +58,10 @@ class AuthorizationController
                               ClientRepository $clients,
                               TokenRepository $tokens)
     {
+        if ($request->get('prompt') === 'login') {
+            $request->session()->flush();
+        }
+
         $authRequest = $this->withErrorHandling(function () use ($psrRequest) {
             return $this->server->validateAuthorizationRequest($psrRequest);
         });


taylorotwell:
Are we sure that flushing the entire session is what we want to do here? Is that all that needs to be done?

mikeschoneveld:
> Are we sure that flushing the entire session is what we want to do here? Is that all that needs to be done?

Maybe the entire session would be unnecessary, if there is a better way to clear the users loggin session so they can login again or login with a different account that would be great.

The problem right now is that if we implement the OAuth 2.0 login in our mobile apps and the user logs out and tries to login again he automatically logs in (because the session is still active in de native browser) and he is not able to switch accounts or something.

driesvints:
@mikeschoneveld hmm I also think flushing the entire session probably isn't what you want to do.

Hey @sephster. Do you believe this should be part of OAuth2-Server instead?

Sephster:
Hey @driesvints, we don't have support for openID connect yet so probably wouldn't take this in at this time.

OpenID connect deals with authentication but the oauth2 server only handles authorisation at the moment.

driesvints:
Thanks @Sephster, that makes sense.

@mikeschoneveld I don't think we should support this at the time. Passport is really meant to be compatible with the OAuth2 protocol and not with OpenID.

franck-grenier:
Too bad... 
I also need this feature in Passport.

PR #1446, opened 2021-05-26
Improve authenticateViaBearerToken() performance

alecpl:
One select from oauth_access_tokens and one from oauth_clients less.


driesvints:
We can't remove a method on a stable release. Please attempt this PR to the master branch. Also try to give a much more thorough explanation why this is needed. Thanks.

PR #1447, opened 2021-05-26
[11.x] Improve authenticateViaBearerToken() performance

alecpl:
One select from oauth_access_tokens and one from oauth_clients less. The issues mentioned in #382.

Before:
```
select * from `oauth_access_tokens` where `id` = ? limit 1
select * from `oauth_access_tokens` where `id` = ? limit 1
select * from `oauth_clients` where `id` = ? limit 1
select * from `oauth_access_tokens` where `id` = ? limit 1
select * from `oauth_clients` where `id` = ? limit 1
```
After:
```
select * from `oauth_access_tokens` where `id` = ? limit 1
select * from `oauth_clients` where `id` = ? limit 1
select * from `oauth_access_tokens` where `id` = ? limit 1
```
as you see there's still one redundant query, but it might be harder to get rid of it.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -82,23 +82,6 @@ class TokenGuard
     }
 
     /**
-     * Determine if the requested provider matches the client's provider.
-     *
-     * @param  \Illuminate\Http\Request  $request
-     * @return bool
-     */
-    protected function hasValidProvider(Request $request)
-    {
-        $client = $this->client($request);
-
-        if ($client && ! $client->provider) {
-            return true;
-        }
-
-        return $client && $client->provider === $this->provider->getProviderName();
-    }
-
-    /**
      * Get the user for the incoming request.
      *
      * @param  \Illuminate\Http\Request  $request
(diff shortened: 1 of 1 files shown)


driesvints:
@billriess I think it's best if you go over these changes if you can spare the time?

src/Guards/TokenGuard.php
```diff
@@ -170,15 +163,6 @@ protected function authenticateViaBearerToken($request)
             $psr->getAttribute('oauth_access_token_id')
         );
 
-        $clientId = $psr->getAttribute('oauth_client_id');
-
-        // Finally, we will verify if the client that issued this token is still valid and
-        // its tokens may still be used. If not, we will bail out since we don't want a
-        // user to be able to send access tokens for deleted or revoked applications.
-        if ($this->clients->revoked($clientId)) {
```
billriess, line 178: Is this check still happening somewhere else?

alecpl: $this->clients->findActive() above makes sure the $client is not revoked.

billriess:
I can take a look at this in the next day or two but glancing over it seems good.

taylorotwell:
I see this was sent to `master`. Is there a breaking change? Should this go to `10.x`?

alecpl:
@driesvints didn't like the protected method removal.

driesvints:
@alecpl @taylorotwell technically we could do the changes on 10.x if we left the method in there. 

@billriess thanks for reviewing 👍

PR #1448, opened 2021-05-29
Update TokenRepository.php

realanmup:
Removing the Type of the Parameter here will allow us to use custom Token model class from auth service provider. Currently, I am having this issue.

`TypeError: Laravel\Passport\TokenRepository::save(): Argument #1 ($token) must be of type Laravel\Passport\Token, App\Models\Passport\Token given, called in /var/www/vendor/laravel/passport/src/PersonalAccessTokenFactory.php on line 80`

### DIFF
--- a/src/TokenRepository.php
+++ b/src/TokenRepository.php
@@ -73,7 +73,7 @@ class TokenRepository
      * @param  \Laravel\Passport\Token  $token
      * @return void
      */
-    public function save(Token $token)
+    public function save($token)
     {
         $token->save();
     }


driesvints:
You need to extend your custom model with the passport token class. This is also detailed in the docs: https://laravel.com/docs/8.x/passport#overriding-default-models

realanmup:
I got that error when i tried extending the Token model using Passport::useTokenModel(Token::class); But it throws the error
`TypeError: Laravel\Passport\TokenRepository::save(): Argument #1 ($token) must be of type Laravel\Passport\Token, App\Models\Passport\Token given, called in /var/www/vendor/laravel/passport/src/PersonalAccessTokenFactory.php on line 80`

driesvints:
Your Token class is not extending the passport token model as detailed in the docs.

driesvints:
```php
use Laravel\Passport\Token as Passport Token;

class Token extends PassportToken
{
    // ...
}
```

PR #1453, opened 2021-06-08
Implement new oauth2-server revoke refresh tokens config option.

janhopman-nhb:
This pr implements a newly added config option for oauth2-server package. The config options adds the ability to prevent refresh tokens from being revoked.

Hope someone can take a look at this.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -160,6 +160,13 @@ class Passport
     public static $withInheritedScopes = false;
 
     /**
+     * Indicates if refresh tokens should be revoked.
+     *
+     * @var bool
+     */
+    public static $revokeRefreshTokens = true;
+
+    /**
      * Enable the implicit grant type.
      *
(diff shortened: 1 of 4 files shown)


src/Passport.php
```diff
@@ -691,4 +698,25 @@ public static function withoutCookieSerialization()
...
+
+    /**
+     * Determine if refresh tokens must be revoked.
+     *
+     * @return bool
+     */
+    public static function revokeRefreshTokens()
+    {
+        return static::$revokeRefreshTokens;
+    }
```
driesvints, line 710: This method isn't needed. Just check the property directly. I realize this is inconsistent with the other methods in this class but they're also unneeded actually. We can remove those in a future release.

driesvints:
Just realized you sent this to v9 which isn't supported anymore. Please send this to 10.x, thanks.

PR #1454, opened 2021-06-08
[10.x] Implement new OAuth2 Server revoke refresh tokens config option

janhopman-nhb:
This pr implements a newly added config option for oauth2-server package. The config options adds the ability to prevent refresh tokens from being revoked.

Hope someone can take a look at this.

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -63,4 +63,16 @@ return [
         ],
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Refresh token revocation
+    |--------------------------------------------------------------------------
+    |
+    | This configuration value allows you to switch revocation of oauth refresh
+    | tokens on or off. By default refresh tokens will always be revoked when
+    | used.
+    |
+    */
+
+    'revoke_refresh_tokens' => env('PASSPORT_REVOKE_REFRESH_TOKENS', true),
 ];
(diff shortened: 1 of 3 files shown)


src/Passport.php
```diff
@@ -651,4 +658,25 @@ public static function withoutCookieSerialization()
...
+
+    /**
+     * Determine if refresh tokens must be revoked.
+     *
+     * @return bool
+     */
+    public static function revokeRefreshTokens()
+    {
+        return static::$revokeRefreshTokens;
+    }
```
driesvints, line 670: Like I said in the previous PR, you can remove this method. The property can be called directly.

config/passport.php
```diff
@@ -63,4 +63,16 @@
...
+    | Refresh token revocation
+    |--------------------------------------------------------------------------
+    |
+    | This configuration value allows you to switch revocation of oauth refresh
+    | tokens on or off. By default refresh tokens will always be revoked when
+    | used.
+    |
+    */
+
+    'revoke_refresh_tokens' => env('PASSPORT_REVOKE_REFRESH_TOKENS', true),
```
driesvints, line 77: Is `true` the current default behavior?

janhopman-nhb: Yes it is!

driesvints:
@janhopman-nhb seems like some tests are broken.

Issue #1452, opened 2021-06-08
Remove Firebase JWT dependencies in favour of Lcobucci JWT


peterfox:
To put it simply, the `firebase/php-jwt` package is a dependency of Passport. Passport has a dependency of `league/oauth2-server` which depends upon `lcobucci/jwt` which has the functionality to encode and decode JWT tokens the same as the Firebase package.

I'm happy to do the work on the changes so `firebase/php-jwt` can be removed so the package only relies on the `lcobucci/jwt` package but would prefer a maintainer to sign off on it before putting the work in if there's a reason why the Firebase package is preferred over using Lcobucci.

driesvints:
I've actually been thinking of that as well. We could do that for the next major release maybe. Although I don't know the exact implications atm. If you're willing, feel free to work on a PR for master to check if we can drop it. Thanks!

peterfox:
In theory it might not even need a major release as from a quick glance the code using Firebase isn't used in any parameters etc and it's just array/objects being returned so as long as the code for Lcobucci is doing the same it could be a swap out. Doesn't seem to reference the exceptions thrown by encode/decode. The package mostly references Lcobucci classes.

The problem I guess would be those using the firebase JWT classes in their apps but haven't added it to their own composer dependencies though.

driesvints:
@peterfox technically that indeed shouldn't be an issue. They should have added firebase to their own dependencies in that case. Maybe try a pr to current stable branch and if we feel it's too much of a breaking change we could still swap to master.

driesvints:
I'm gonna close this one for now. If you're willing, feel free to send in a PR.

PR #1458, opened 2021-06-19
Refactors/Removes Firebase from Package

peterfox:
Satisfies enhancement issue #1452 in removing all the code relying on Firebase's JWT package to instead rely on Lcobucci's package.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -3,7 +3,6 @@
 namespace Laravel\Passport\Guards;
 
 use Exception;
-use Firebase\JWT\JWT;
 use Illuminate\Container\Container;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Contracts\Encryption\Encrypter;
@@ -11,6 +10,7 @@ use Illuminate\Cookie\CookieValuePrefix;
 use Illuminate\Cookie\Middleware\EncryptCookies;
 use Illuminate\Http\Request;
 use Laravel\Passport\ClientRepository;
+use Laravel\Passport\Jwt\Decoder;
 use Laravel\Passport\Passport;
 use Laravel\Passport\PassportUserProvider;
 use Laravel\Passport\TokenRepository;
@@ -131,7 +131,7 @@ class TokenGuard
             );
         } elseif ($request->cookie(Passport::cookie())) {
             if ($token = $this->getTokenViaCookie($request)) {
-                return $this->clients->findActive($token['aud']);
+                return $this->clients->findActive($token['aud'][0]);
             }
         }
     }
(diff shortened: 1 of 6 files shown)


src/Guards/TokenGuard.php
```diff
@@ -131,7 +131,7 @@ public function client(Request $request)
             );
         } elseif ($request->cookie(Passport::cookie())) {
             if ($token = $this->getTokenViaCookie($request)) {
-                return $this->clients->findActive($token['aud']);
+                return $this->clients->findActive($token['aud'][0]);
```
taylorotwell, line 134: Why was this change needed?

peterfox: The Lcobucci library breaks up audiences so you can have multiple values instead of a single string. When it parses the token it always produces an array. This is the default behaviour anyways. I don't know if there is potentially a way to stop that. It could in the Encoder class the audience should just be imploded into a string to replicate the way Firebase presents that value. I can do a bit more research into that.

I'll admit it as it stands it for people extending the method of the TokenGuard, then this change could be problematic for just a minor release.

driesvints:
Hey @peterfox, appreciate the PR. I didn't realize we needed this much code to remove Firebase. Seems a bit problematic to maintain all of that code ourselves?

Also: I indeed think it's best that we do this in the next major release. Just to prevent breakges.

peterfox:
@driesvints fair comments, it was more involved than I initially expected as well. I guess I could consider wrapping it up in a simple package as a drop in replacement for Firebase to reduce the code being added to passport itself.

Would you like me to look at opening up a PR for this against the master branch or would you rather we leave it for now?

driesvints:
@peterfox. Thanks, I appreciate the offer but wouldn't it then be better if we kept Firebase? Swapping one package out for the other doesn't seems like the right choice. What was your original reason to remove Firebase? Also: is the other way around (swapping Lcobucci's JWT library out for Firebase) a better option?

peterfox:
@driesvints the reason for removal was less redundancy. Laravel passport relies on [League's OAuth2 Server](https://github.com/thephpleague/oauth2-server/blob/master/composer.json) which in turn relies on Lcobucci's JWT package. So either way Lcobucci will be installed with an installation of Laravel passport. It can't be swapped out for Firebase because Passport is built around that OAuth2 Server package so that just isn't an option.

In the realms of Passport itself, keeping the Firebase package is fine and doesn't affect anything. In the larger scale of things, the Firebase package is a bit outdated and isn't actively maintained as well as the Lcobucci one, but as you've seen from the PR it's also a more complicated package with more configuration required.

This PR can probably be dropped for now, but I can see a future where the Firebase package ends up getting abandoned for more people picking up Lcobucci as it supports more in terms of security and functionality.

driesvints:
Thanks @peterfox for that explanation. I do see @bshaffer is still maintaining the Firebase package actively so maybe this isn't needed in the end. I think we can leave things be for now. This was a good experiment though so thanks for that.

peterfox:
@driesvints no worries. Was worthwhile looking into for myself also. Feel free to drop me a line about this if the situation changes.

bshaffer:
It's true!! I'm maintaining it! We use it for all Google's client libraries.

PR #1461, opened 2021-07-01
Custom Token Fails

blpraveen:
Please make the changes necessary
When I use Custom token
attempt to login failed
Passport::useTokenModel(Token::class);


driesvints:
Heya. This is a breaking change. Most likely, you haven't extended the base Token model as documented: https://laravel.com/docs/8.x/passport#overriding-default-models

Issue #1414, opened 2021-02-23
Move routes to dedicates routes file


driesvints:
Passport is probably one of the oldest first party package we have. It still uses the old way of setting up routes through a dedicated routes method on the Passport class. It's best that we move it to a new `routes/web.php` file in similar fashion as all the other first party packages.

HassanYA:
I second this. I have installed laravel passport very recently and have no idea what the endpoints to begin with and cannot write my api documentation

driesvints:
Will land in Passport v11

PR #1464, opened 2021-07-16
[11.x] Refactor routes to dedicated file

driesvints:
This PR refactors Passport's routes to a dedicated file at `routes/web.php`. Passport is one of the oldest first party packages that we maintain and it was still using an outdated way to handle routes. This PR will bring Passport's routes on-par with the rest of our first party packages and to what our users are used to.

The upgrade path for this is easy as it'll only require the removal of `Passport::routes()` in their service provider.

Additionally, I refactored the service provider a bit to be more similar to our other package's service providers.

Closes https://github.com/laravel/passport/issues/1414


taylorotwell:
Does this remove any customization people use to have the ability to do?

driesvints:
You mean the params from `routes($callback = null, array $options = [])`? Yes that wouldn't be possible anymore unless you overwrite the service provider and the `registerRoutes` method. But we didn't document that ability to customize. 

The `$callback` seems to be useful for when you only want to load certain parts of routes. But you can still adjust the prefix (which I guess is what people mostly want to do) through `config('passport.path', 'oauth')`.

sebastiaanluca:
#1564 does help a lot here. Currently made the change to disable the routes and copied the token route manually.

A system like Fortify would be very welcome, you don't always want to expose all Passport endpoints (actually only ever needed to provide a few, never all).

driesvints:
I'll check into adding that except array to the ignoreRoutes method when I find some time. Otherwise we'd appreciate a PR. Thanks

siarheipashkevich:
@driesvints thanks.

Do you plan to add an ability to ignore specific routes?

driesvints:
Whenever I find the time. Otherwise a PR is appreciated.

SturmB:
Our company's Laravel API app uses this specifically:

```php
// app/Providers/AuthServiceProvider.php

    public function boot()
    {
        $this->registerPolicies();

        //Below we intercept the generation of the /oauth/token route from passport and set the middleware on it to Throttle
        //This limits the user to hit the route 10 times per minute (10,1).
        //The solution for this was found at the link here: https://stackoverflow.com/a/59985823/1621272
        Passport::routes(
            function($routeRegistrar){
                $routeRegistrar->all();
                Route::post('/token', [
                    'uses' => 'AccessTokenController@issueToken',
                    'middleware' => 'throttle:' . env('THROTTLE_LIMIT_AUTHSERVICE', self::DEFAULT_THROTTLE) . ',1',
                    'as' => 'passport.token'
                ]);
            });
    }
```

Is there a way to perform the same throttling now that `routes()` is gone?

driesvints:
@SturmB please see https://laravel.com/docs/9.x/passport#overriding-routes on how to override routes and do whatever you want 👍

SturmB:
Thank you for responding, @driesvints. I'm still learning about routing and middleware, so I'm having difficulty understanding how to adapt this to our use case. For context, I'm attempting to upgrade our Laravel API app from 8 to 9 and bring all of its dependencies up to date in the process.

Please correct me if I am wrong, but from the documentation you linked as well as Passport's `web.php` file, it appears that I would want to write something like this:

```php
Route::post('/token', [
    'uses' => 'AccessTokenController@issueToken',
    'as' => 'token',
    'middleware' => 'throttle:' . env('THROTTLE_LIMIT_AUTHSERVICE', 750) . ',1',
]);
```

and place it in our `routes/apiV2.php` file (since we use Laravel only as an API server).

Would this intercept the generation of the `/oauth/token` route from Passport and adjust its throttle amount? And I'm guessing it'll still work for any of our endpoints that use Passport (of which almost all of them do)?

driesvints:
I think so yes. In general you can overwrite routes in your own routes files if you want. I *think* that'll work but you should definitely try it out to be sure (as well as try route caching).

vstyler96:
Some tip for retrieving personal access tokens from: `/oauth/personal-access-tokens` from axios? This is now resulting in a "Unauthenticated" issue whehn calling from JS client

src/PassportServiceProvider.php
```diff
@@ -34,13 +35,61 @@ class PassportServiceProvider extends ServiceProvider
...
-        if ($this->app->runningInConsole()) {
-            $this->registerMigrations();
+    /**
+     * Register the Passport routes.
+     *
+     * @return void
+     */
+    protected function registerRoutes()
+    {
+        Route::group([
```
siarheipashkevich, line 54: @driesvints hello, why did you not move the group declaration to `web.php`?

In that case you can avoid useless `Route::group` declaration if routes are cached (condition inside `$this->loadRoutesFrom()`).

And this method will look like:
```
protected function registerRoutes()
{
    $this->loadRoutesFrom(__DIR__.'/../routes/web.php');
}
```
or we can remove this method wrapper and call directly `$this->loadRoutesFrom(__DIR__.'/../routes/web.php');`

What do you think?

driesvints: We don't do this mostly in our packages. It's fine I think.

siarheipashkevich: Yes, I've checked all laravel packages with route and think that the core team avoid using `config` inside `routes` files, but I saw that you it. And also we have duplicated conditions in that way.

For example horizon package:
https://github.com/laravel/horizon/blob/5.x/src/HorizonServiceProvider.php#L54

we have the condition which already exists inside `loadRoutesFrom` method

driesvints: @siarheipashkevich this condition was added in https://github.com/laravel/horizon/pull/1367 only recently.

siarheipashkevich: @driesvints thanks, but I think we should optimize this for avoiding extra time of code execution.

driesvints: It's fine really.

PR #1468, opened 2021-07-27
chore: update PHPDoc of issueToken

DeepDiver1975:
withErrorHandling() can throw an OAuthServerException - issueToken should declare this via PHPDoc

### DIFF
--- a/src/Http/Controllers/AccessTokenController.php
+++ b/src/Http/Controllers/AccessTokenController.php
@@ -57,6 +57,7 @@ class AccessTokenController
      *
      * @param  \Psr\Http\Message\ServerRequestInterface  $request
      * @return \Illuminate\Http\Response
+     * @throws \Laravel\Passport\Exceptions\OAuthServerException
      */
     public function issueToken(ServerRequestInterface $request)
     {


DeepDiver1975:
seriously? changing PHPDoc causes tests to fail ...... I'll have a look ....

DeepDiver1975:
> seriously? changing PHPDoc causes tests to fail ...... I'll have a look ....

master fails as well ....

Sephster:
This is likely due to a change in OAuth 2 server in the way we handle invalid passwords. Bit busy just now but should be able to take a look in an hour or so

DeepDiver1975:
> This is likely due to a change in OAuth 2 server in the way we handle invalid passwords.

The last successful ci run in here used 8.2.4 - this run used 8.3.2 ...
here is the diff - https://github.com/thephpleague/oauth2-server/compare/8.2.4...8.3.2

Sephster:
PR #1469 fixes the issues you are seeing with tests

taylorotwell:
I don't see any exceptions thrown in this method.

DeepDiver1975:
In trait HandlesOAuthErrors the method withErrorHandling() is for sure throwing this exception. Please check it's implementation and phpdoc. THX

DeepDiver1975:
> In trait HandlesOAuthErrors the method withErrorHandling() is for sure throwing this exception. Please check it's implementation and phpdoc. THX

@taylorotwell

DeepDiver1975:
:wave: @taylorotwell

PR #1501, opened 2021-10-28
[10.x] Add custom encryption key for JWT tokens

thijskok:
This PR adds the ability to use a custom encryption key for encrypting tokens.

## Background
Passport provides a convenient way for authenticating front-end calls to the backend using a **XSRF token**: 

- The token is generated using the Encrypter class with the APP_KEY as the encryption key. 
- Authentication is successful when the submitted token can be decrypted in the backend using the same APP_KEY. 

The user id is retrieved from the payload and you're good to go.

## Problem

In some cases, the ability to decrypt the token is insufficient proof for establishing the user's identity. This happens when multi-tenancy is used (or multiple instances of the same application) while sharing the same APP_KEY among the tenants. 

For example, when using Archtechx's [tenancy](https://github.com/archtechx/tenancy) or Spatie's [laravel-multitenancy](https://spatie.be/docs/laravel-multitenancy/) together with separate databases per tenant, tokens that are encrypted by the first tenant will decrypt without a hassle by a second tenant. This causes a nasty security breach.

## Solution

Our initial thought was to rotate APP_KEY's for each environment, but that's quite complicated as the Encrypter is instantiated in quite an early stage in the Laravel lifecycle - and many developers recommend against changing the APP_KEY along the way.

This solution adds a callback method that provides a way of changing the encryption key without compromising the existing implementation:

```php

// Example code fragment. Tenant is represented by $instance.

Passport::encryptTokenUsing(function(Encrypter $encrypter) { 
    return $encrypter->getKey() . $instance->getKey();
});

```

Default behaviour is to remain using the APP_KEY by calling the `$encrypter->getKey()` method, which avoids breaking any existing implementations.

## Alternatives

- Rotating APP_KEY's per tenant: is discouraged by many (but perhaps unjustly?).
- Override the Encrypter's behaviour using the IoT Container: doesn't seem to work, the Encrypter is instantiated in an early stage, causing different APP_KEY's to be used throughout the request lifecycle.
- Create your own PassportServiceProvider that extends the existing one and override the token behaviour, which requires quite a bit of rewriting. It's an alternative, but many with the same setup would benefit from a built-in solution, mitigating potential security issues.

### DIFF
--- a/src/ApiTokenCookieFactory.php
+++ b/src/ApiTokenCookieFactory.php
@@ -77,6 +77,6 @@ class ApiTokenCookieFactory
             'sub' => $userId,
             'csrf' => $csrfToken,
             'expiry' => $expiration->getTimestamp(),
-        ], Passport::encryptionKey($this->encrypter));
+        ], Passport::tokenEncryptionKey($this->encrypter));
     }
 }
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -269,7 +269,7 @@ class TokenGuard
     {
         return (array) JWT::decode(
             CookieValuePrefix::remove($this->encrypter->decrypt($request->cookie(Passport::cookie()), Passport::$unserializesCookies)),
-            Passport::encryptionKey($this->encrypter),
+            Passport::tokenEncryptionKey($this->encrypter),
             ['HS256']
         );
     }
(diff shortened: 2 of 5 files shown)


thijskok:
Thanks, @taylorotwell 👍

PR #1508, opened 2021-11-23
[11.x] Allow authenticated client to be retrieved from the guard

### LINKED ISSUE #882: Allow to retrieve the current client through the RequestGuard
We should enhance the `RequestGuard` so it's able to retrieve the current client that's being used as an addition to https://github.com/laravel/passport/pull/854.

This probably requires some breaking changes and should target the next major release.

axlon:
Fixes #882 

This PR allows developers to retrieve the currently authenticated client from the guard, this functionality was already present but inaccessible due to the guard being wrapped in another guard.
Tests all still work (I had to move some stuff around because request is now getting passed through the constructor, but I didn't functionally change any tests).

What this PR changes:
- Turns `TokenGuard` into an actual guard, implementing the `Guard` interface
- Removes the wrapped `RequestGuard`, instead `TokenGuard` is now used as is

A lot of the non-public methods on `TokenGuard` still take `$request` as an argument, which could be removed in favor of `$this->request`. I opted not to do this to keep the PR as small as possible

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -12,6 +12,7 @@ use Illuminate\Contracts\Encryption\Encrypter;
 use Illuminate\Cookie\CookieValuePrefix;
 use Illuminate\Cookie\Middleware\EncryptCookies;
 use Illuminate\Http\Request;
+use Illuminate\Support\Traits\Macroable;
 use Laravel\Passport\ClientRepository;
 use Laravel\Passport\Passport;
 use Laravel\Passport\PassportUserProvider;
@@ -24,7 +25,7 @@ use Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory;
 
 class TokenGuard implements Guard
 {
-    use GuardHelpers;
+    use GuardHelpers, Macroable;
 
     /**
      * The resource server instance.
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -146,6 +146,10 @@ class TokenGuard implements Guard
      */
     public function client()
     {
+        if (! is_null($this->client)) {
+            return $this->client;
+        }
+
         if ($this->request->bearerToken()) {
             if (! $psr = $this->getPsrRequestViaBearerToken($this->request)) {
                 return;
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -4,7 +4,9 @@ namespace Laravel\Passport\Guards;
 
 use Exception;
 use Firebase\JWT\JWT;
+use Illuminate\Auth\GuardHelpers;
 use Illuminate\Container\Container;
+use Illuminate\Contracts\Auth\Guard;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Contracts\Encryption\Encrypter;
 use Illuminate\Cookie\CookieValuePrefix;
@@ -20,8 +22,10 @@ use League\OAuth2\Server\ResourceServer;
 use Nyholm\Psr7\Factory\Psr17Factory;
 use Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory;
 
-class TokenGuard
+class TokenGuard implements Guard
 {
+    use GuardHelpers;
+
     /**
      * The resource server instance.
      *
@@ -58,6 +62,20 @@ class TokenGuard
     protected $encrypter;
 
     /**
+     * The request instance.
+     *
+     * @var \Illuminate\Http\Request
+     */
+    protected $request;
+
+    /**
+     * The currently authenticated client.
+     *
+     * @var \Laravel\Passport\Client|null
+     */
+    protected $client;
+
+    /**
      * Create a new token guard instance.
      *
      * @param  \League\OAuth2\Server\ResourceServer  $server
@@ -65,6 +83,7 @@ class TokenGuard
      * @param  \Laravel\Passport\TokenRepository  $tokens
      * @param  \Laravel\Passport\ClientRepository  $clients
      * @param  \Illuminate\Contracts\Encryption\Encrypter  $encrypter
+     * @param  \Illuminate\Http\Request  $request
      * @return void
      */
     public function __construct(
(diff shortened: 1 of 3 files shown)


driesvints:
I'll try to review this soon. Thanks!

src/Guards/TokenGuard.php
```diff
@@ -65,56 +83,79 @@ class TokenGuard
...
      *
-     * @param  \Illuminate\Http\Request  $request
      * @return mixed
      */
-    public function client(Request $request)
+    public function client()
     {
-        if ($request->bearerToken()) {
-            if (! $psr = $this->getPsrRequestViaBearerToken($request)) {
+        if ($this->request->bearerToken()) {
```
driesvints, line 148: You probably want to return the already resolved client here early like the user method does.

src/Guards/TokenGuard.php
```diff
@@ -20,8 +22,10 @@
 use Nyholm\Psr7\Factory\Psr17Factory;
 use Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory;
 
-class TokenGuard
+class TokenGuard implements Guard
 {
+    use GuardHelpers;
```
driesvints, line 27: Maybe make this macroable as well?

driesvints:
@axlon this is a great PR, thank you. If you could answer or resolve my two remarks above then I think we can mark this as ready. Please request a new review from me before doing that.

src/Guards/TokenGuard.php
axlon: Oops, fixing that 👍

driesvints (approved):
@axlon awesome, thanks!

taylorotwell:
I have some fears this will break usage of Passport in Octane with the way it is caching the user and client on the class / guard instance.

driesvints:
@axlon are you willing to check into that? Otherwise I can pick this up once I find some time on hand. That might still be a while though.

axlon:
@driesvints I already briefly checked this before I made the PR initially, if I'm not mistaken Octane flushes all guards from the AuthManager on every request (I think the framework's own guards also cache users this way).

That being said, I'll do an actual test on Octane to see if I'm not missing anything. I'll report back when I have.

axlon:
@driesvints @taylorotwell I did some tests, both against current Passport and this PR

TL;DR: This PR should not break Octane-driven Passport applications

## Test 1: Octane flushes guards on every request

My initial assumption was based on [this Octane code](https://github.com/laravel/octane/blob/1.x/src/Listeners/FlushAuthenticationState.php). To verify, I added the following middleware to an existing Passport + Octane app:

```php
public function handle(Request $request, Closure $next)
{
    $response = $next($request);

    logger(spl_object_hash(Auth::guard('api')));

    return $response;
}
```

Then I refreshed the application homepage a couple of times and got this log file:

```txt
[2021-11-25 19:07:04] local.DEBUG: 000000007750030e000000004c1943a6  
[2021-11-25 19:07:07] local.DEBUG: 000000007750030f000000004c1943a6  
[2021-11-25 19:07:11] local.DEBUG: 00000000775000a1000000004c1943a6  
[2021-11-25 19:07:11] local.DEBUG: 00000000775000c9000000004c1943a6  
[2021-11-25 19:07:11] local.DEBUG: 00000000775000de000000004c1943a6  
[2021-11-25 19:07:11] local.DEBUG: 00000000775000cf000000004c1943a6  
[2021-11-25 19:07:16] local.DEBUG: 00000000775000dc000000004c1943a6  
[2021-11-25 19:07:16] local.DEBUG: 00000000775000d4000000004c1943a6  
```

This verifies that the guard is flushed on every request.

## Test 2: PR does not break this behaviour

I also ran a test on this PR branch, with a slightly modified middleware:

(I used a Postman Collection to test against various routes)

```php
public function handle(Request $request, Closure $next)
{
    $response = $next($request);

    logger(vsprintf('%s-%s-%s', [
        spl_object_hash(Auth::guard('api')),
        Auth::guard('api')->client() ? spl_object_hash(Auth::guard('api')->client()) : 'null',
        Auth::guard('api')->user() ? spl_object_hash(Auth::guard('api')->user()) : 'null',
    ]));

    return $response;
}
```

Which resulted in this log file:

(The two log entries with nulls are HTTP 401 responses)

```txt
[2021-11-25 19:28:35] local.DEBUG: 0000000049ace958000000007329001e-0000000049ace899000000007329001e-0000000049ace8c3000000007329001e  
[2021-11-25 19:28:36] local.DEBUG: 000000005541b041000000004555dd68-000000005541b003000000004555dd68-000000005541b014000000004555dd68  
[2021-11-25 19:28:36] local.DEBUG: 0000000072ade5a2000000004a198ad1-0000000072ade41a000000004a198ad1-0000000072ade406000000004a198ad1  
[2021-11-25 19:28:36] local.DEBUG: 0000000044677116000000005ec83017-0000000044677155000000005ec83017-0000000044677159000000005ec83017  
[2021-11-25 19:28:36] local.DEBUG: 0000000058221f04000000004a1bf6d8-0000000058221f47000000004a1bf6d8-0000000058221f53000000004a1bf6d8  
[2021-11-25 19:28:36] local.DEBUG: 00000000216f03b600000000050d1a85-00000000216f03e500000000050d1a85-00000000216f03d300000000050d1a85  
[2021-11-25 19:28:36] local.DEBUG: 000000005cf4349200000000164782d9-000000005cf434d400000000164782d9-000000005cf434cf00000000164782d9  
[2021-11-25 19:28:36] local.DEBUG: 0000000035fb85150000000010211077-0000000035fb85770000000010211077-0000000035fb85610000000010211077  
[2021-11-25 19:28:37] local.DEBUG: 0000000025f433db000000005f94cdab-0000000025f43399000000005f94cdab-0000000025f43386000000005f94cdab  
[2021-11-25 19:28:37] local.DEBUG: 000000006b8e3b00000000006494e8eb-000000006b8e3b4e000000006494e8eb-000000006b8e3b55000000006494e8eb  
[2021-11-25 19:28:37] local.DEBUG: 000000000174137f000000000a48e52c-000000000174133d000000000a48e52c-0000000001741328000000000a48e52c  
[2021-11-25 19:28:37] local.DEBUG: 000000000ca7f6820000000012b86bc1-null-null  
[2021-11-25 19:28:37] local.DEBUG: 000000002688bebb0000000024295dfb-null-null  
[2021-11-25 19:28:37] local.DEBUG: 000000007c1d7831000000007b972b76-000000007c1d7859000000007b972b76-000000007c1d784a000000007b972b76  
[2021-11-25 19:28:37] local.DEBUG: 000000003ab0afad0000000004dc10e6-000000003ab0afc50000000004dc10e6-000000003ab0afd60000000004dc10e6  
```

driesvints:
Thanks a lot @axlon 👍

taylorotwell:
Thanks for contributing to Laravel! ❤️

PR #1517, opened 2022-01-12
add optional clientId param to createToken

saeedvaziry:
Currently, The `createToken` method uses the first personal access client from the database if the`passport.personal_access_client` config wasn't set at all.
I added the `clientId` param to the `createToken` method so you can create tokens for any clients by their id.

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -106,7 +106,7 @@ class ClientRepository
     /**
      * Get the personal access token client for the application.
      *
-     * @param  int|null $clientId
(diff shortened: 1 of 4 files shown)


taylorotwell:
Breaking change I don't want to make - set the personal_access_client in the configuration.

saeedvaziry:
@taylorotwell Correct. But what if there is multiple personal access clients? How we can switch dynamically between them?

saeedvaziry:
What if we add a new method to create tokens for a specific client?

waney:
me too.

PR #1519, opened 2022-01-18
[11.x] Stub client on guard when calling Passport::actingAsClient()

axlon:
This PR changes the `Passport::actingAsClient()` method. Currently when stubbing the client via Passport, the guard doesn't actually reflect the stub:

```php
>>> $client = Laravel\Passport\Database\Factories\ClientFactory::new()->createOne()
=> Laravel\Passport\Client {#4807}
>>> Laravel\Passport\Passport::actingAsClient($client)
=> Laravel\Passport\Client {#4807}
>>> auth()->guard('api')->client()
=> null
```

After this PR the following happens:

```php
>>> $client = Laravel\Passport\Database\Factories\ClientFactory::new()->createOne()
=> Laravel\Passport\Client {#4807}
>>> Laravel\Passport\Passport::actingAsClient($client)
=> Laravel\Passport\Client {#4807}
>>> auth()->guard('api')->client()
=> Laravel\Passport\Client {#4807}
```

This makes `Passport::actingAsClient()` behave similar to `Passport::actingAs()` which sets the user on the api guard.

Why is this necessary:
- Code that uses Passport's middleware and also uses the TokenGuard to retrieve the client (e.g. for tracking which client created a model) will be hard to test because `Passport::actingAsClient()` does trick the middleware but not the guard. Note that this isn't an issue in the current version of Passport because `TokenGuard::client()` isn't currently accessible as a way to resolve the client making a request.

Breaking changes:
- This PR makes `Passport::actingAsClient()` assume that a `setClient()` method exists on the guard. This should almost never be a problem because you're testing Passport-related code specifically. But a `instanceof` check may be necessary to make it fully backwards compatible

Are there alternative ways of fixing this:
- Yes, you could mock the request instance as well, and make sure all the things that the guard checks return correctly. Though this has impact on the application state as whole (after all a lot of things rely on the request instance) and probably would introduce a lot of issues into tests calling `Passport::actingAsClient()`

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -13,6 +13,7 @@ use Illuminate\Cookie\CookieValuePrefix;
 use Illuminate\Cookie\Middleware\EncryptCookies;
 use Illuminate\Http\Request;
 use Illuminate\Support\Traits\Macroable;
+use Laravel\Passport\Client;
 use Laravel\Passport\ClientRepository;
 use Laravel\Passport\Passport;
 use Laravel\Passport\PassportUserProvider;
@@ -352,5 +353,18 @@ class TokenGuard implements Guard
     public static function serialized()
     {
         return EncryptCookies::serialized('XSRF-TOKEN');
+    }
+
+    /**
+     * Set the client for the current request.
+     *
+     * @param  \Laravel\Passport\Client  $client
+     * @return $this
+     */
+    public function setClient(Client $client)
+    {
+        $this->client = $client;
+
+        return $this;
     }
 }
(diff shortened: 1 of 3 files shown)


driesvints:
How are you testing this? What's holding you back? Why doesn't `actingAsClient` work for you? Can you share some code?

axlon:
@driesvints I updated the PR, I hope its more clear now (its honestly kind of hard to explain haha). If you have any questions don't hesitate to ask me

driesvints (approved):
This seems okay to me. Thanks @axlon 👍

PR #1520, opened 2022-01-19
Allow to use custom authorization server response

yaroslawww:
*league/oauth2-server* supports custom response type [https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L103](https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L103).

This pull request add possibility to override default "BearerTokenResponse" to custom schema (for example to add new extra parameters to support OpenID Connect flow...).

To override response need add to service provider:
```php
Passport::$authorizationServerResponseType = new CustomResponse();
```

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -7,6 +7,7 @@ use DateInterval;
 use DateTimeInterface;
 use Illuminate\Contracts\Encryption\Encrypter;
 use League\OAuth2\Server\ResourceServer;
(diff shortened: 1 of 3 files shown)


taylorotwell:

driesvints:
@yaroslawww I talked to Taylor about this and we can make this work. This can even go to the current stable branch. Can you send this to 10.x?

PR #1521, opened 2022-01-21
[10.x] Allow to use custom authorization server response

yaroslawww:
**league/oauth2-server** supports custom response type [https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L103](https://github.com/thephpleague/oauth2-server/blob/master/src/AuthorizationServer.php#L103).

This pull request add possibility to override default "BearerTokenResponse" to custom schema (for example to add new extra parameters to support OpenID Connect flow...).

To override response need add to service provider:

```php
Passport::$authorizationServerResponseType = new CustomResponse();
```

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -37,6 +37,13 @@ class Passport
     ];
 
     /**
+     * The authorization server response formatter.
+     *
+     * @var \League\OAuth2\Server\ResponseTypes\ResponseTypeInterface|null
+     */
+    public static $authorizationServerResponseType = null;
+
+    /**
      * The date when access tokens expire.
      *
      * @var \DateTimeInterface|null
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -37,13 +37,6 @@ class Passport
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


src/Passport.php
```diff
@@ -36,6 +36,13 @@ class Passport
         //
     ];
 
+    /**
+     * The authorization server response formatter.
+     *
+     * @var \League\OAuth2\Server\ResponseTypes\ResponseTypeInterface|null
+     */
+    public static $authorizationServerResponseType = null;
```
driesvints, line 44: Think you can leave off the `null` here?
```suggestion
    public static $authorizationServerResponseType;
```

driesvints (approved):

src/Passport.php
yaroslawww: Sure :100:

driesvints:
Thanks @yaroslawww

yaroslawww:
@driesvints Thank you for your approval. This really helps me not to replace the service provider or "binding" to set a custom response class.

PR #1530, opened 2022-02-15
[10.x] Fix Faker deprecations

X-Coder264:
When using this factory in feature tests with `$this->withoutDeprecationHandling();` call the test fails as the deprecation gets converted to an exception:

```
Since fakerphp/faker 1.14: Accessing property "company" is deprecated, use "company()" instead.
```

https://github.com/FakerPHP/Faker/pull/164


driesvints (approved):

driesvints:
Thanks @X-Coder264

Issue #1532, opened 2022-02-26
SQLSTATE[01000]: Warning: 1265 Data truncated for column 'id'


sinarajabpour1998:
- Passport Version: 10.3.1
- Laravel Version: 8.81.0
- PHP Version: 7.3
- Database Driver & Version: MySQL 5.7.33

### Description:
in `php artisan passport:client` cli command with entering this id `1` to answer the `Which user ID should the client be assigned to?`, getting this error :

`
SQLSTATE[01000]: Warning: 1265 Data truncated for column 'id' at row 1 (SQL: insert into 'oauth_clients' ('user_id', 'name', 'secret', 'provi
der', 'redirect', 'personal_access_client', 'password_client', 'revoked', 'id', 'updated_at', 'created_at') values (1, crafter, hashed, ?, http://localhost:8000/auth/callback, 0, 0, 0, hashed, 2022-02-2
6 11:03:19, 2022-02-26 11:03:19))
`

driesvints:
Works fine for me:

<img width="991" alt="Screenshot 2022-02-28 at 09 14 41" src="https://user-images.githubusercontent.com/594614/155947937-6a6cbcbf-95c1-4be4-b3a9-df38a7d979d1.png">

Please try a support channel:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

PR #1533, opened 2022-02-26
Changed id column type to string in create_oauth_clients_table migrat…

sinarajabpour1998:
…ion to fix data truncated bug in creating new client via cli. Ref #1532

Fixed `SQLSTATE[01000]: Warning: 1265 Data truncated for column 'id' at row 1` error for `oauth_clients` table


taylorotwell:
Seems like a breaking change?

PR #1548, opened 2022-06-16
[10.x] Fix scope inheritance when using Passport::actingAs()

### LINKED ISSUE #1138: Inherited scopes ignored when using actingAs
- Passport Version: 7.5.1
- Laravel Version: 6.6.2
- PHP Version: 7.2.24
- Database Driver & Version: n/a 

### Description:

When using inherited scopes and you attempt to set scope inside `Passport::actingAs($user, ['api'])` the can method returns false when checking scope `api:users:index`

### Steps To Reproduce:

```
    public function test_acting_as_ignores_with_inherited_scopes()
    {
        Passport::$withInheritedScopes = true;

        $user = Passport::actingAs(factory(User::class)->create(), ['api']);

        $this->assertTrue($user->tokenCan('api')); // Works as expected
        $this->assertTrue($user->tokenCan('api:users')); // Fails
        $this->assertTrue($user->tokenCan('api:users:index')); // Fails
    }

    public function test_acting_as_ingores_wildcard()
    {
        Passport::$withInheritedScopes = true;

        $user = Passport::actingAs(factory(User::class)->create(), ['*']);

        $this->assertTrue($user->tokenCan('api')); // Fails
    }
``` 

The issue related to `Passport.php` acting as method:

```
    public static function actingAs($user, $scopes = [], $guard = 'api')
    {
        $token = Mockery::mock(self::tokenModel())->shouldIgnoreMissing(false);

        // This check is different to the logic performed inside the standard token model
        foreach ($scopes as $scope) {
            $token->shouldReceive('can')->with($scope)->andReturn(true);
        }

        $user->withAccessToken($token);

        if (isset($user->wasRecentlyCreated) && $user->wasRecentlyCreated) {
            $user->wasRecentlyCreated = false;
        }

        app('auth')->guard($guard)->setUser($user);

        app('auth')->shouldUse($guard);

        return $user;
    }
```
 
`Token.php` here logic to handle wildcard and inherited scopes is checked

```
    public function can($scope)
    {
        if (in_array('*', $this->scopes)) {
            return true;
        }

        $scopes = Passport::$withInheritedScopes
            ? $this->resolveInheritedScopes($scope)
            : [$scope];

        foreach ($scopes as $scope) {
            if (array_key_exists($scope, array_flip($this->scopes))) {
                return true;
            }
        }

        return false;
    }
```

axlon:
Fixes #1138 

This fixes the issue of inherited scopes not working with `Passport::actingAs()`, by not mocking the Token model and just passing a real one with the requested scopes. I couldn't find any reason as to why the model is currently mocked, and all the tests still seem to pass fine.

(PS. When will Passport 11 be released? Passport did not get a new major when Laravel 9 came out, is there some stuff that still needs to be done?)

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -390,11 +390,8 @@ class Passport
      */
     public static function actingAs($user, $scopes = [], $guard = 'api')
     {
-        $token = Mockery::mock(self::tokenModel())->shouldIgnoreMissing(false);
-
-        foreach ($scopes as $scope) {
-            $token->shouldReceive('can')->with($scope)->andReturn(true);
-        }
+        $token = app(self::tokenModel());
+        $token->scopes = $scopes;
 
         $user->withAccessToken($token);
 
--- a/tests/Feature/ActingAsTest.php
+++ b/tests/Feature/ActingAsTest.php
@@ -4,6 +4,7 @@ namespace Laravel\Passport\Tests\Feature;
 
 use Illuminate\Contracts\Routing\Registrar;
 use Illuminate\Foundation\Auth\User;
+use Illuminate\Support\Facades\Route;
 use Laravel\Passport\HasApiTokens;
 use Laravel\Passport\Http\Middleware\CheckForAnyScope;
 use Laravel\Passport\Http\Middleware\CheckScopes;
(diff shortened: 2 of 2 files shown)


axlon:
Should this go to master? I suppose changing the fake token's type could be a breaking change albeit only in tests

taylorotwell:
Yeah I'm not sure we could do this on a patch release?

Issue #1466, opened 2021-07-22
Revoke token with a POST request


hsul4n:
I't will be good feature if we can revoke a token directly using a built in `POST` request, for example:

```curl
curl -X POST https://www.example.com/oauth/revoke \
     -d token=$TOKEN \
     -H "Authorization: Bearer $TOKEN"
```

I think this will be used also as a `logout` option.

driesvints:
Feel free to attempt a PR.

Issue #1389, opened 2020-12-02
Logout and PKCE


pktharindu:
- Passport Version: 10.1.0
- Laravel Version: 8.16.1
- PHP Version: 7.4.8
- Database Driver & Version: MySQL 5.7

### Description:
I'm using [Authorization Code Grant with PKCE](https://laravel.com/docs/8.x/passport#code-grant-pkce) for a browser extension we are working on, and it works as expected. The only inconvenience that I have is when I log out (by revoking the token), and try to re-login, it skips the login form entirely and jumps straight into the callback page. In other words, users won't get to enter their credentials the second time around.

This happens because even though the access token is revoked, the server session is still active.

Although this may be the default behavior, I personally think this breaks users' expectations and possibly introduces a security concern (i.e. after I log out, someone else can access my account just by clicking "log in", without having to enter my credentials).

After a bit of Googling, I found that the `prompt` parameter can be used to change this behavior.

- `prompt=none`: When prompt is set to none, Hosted Login first checks to see if the client has a valid session. If a valid session is found the user doesn't need to authenticate; instead, he or she is automatically logged in using the existing session. If a valid session can't be found a "No authenticated session found" error is generated and the user is not given the option of logging in.
- `prompt=login`: The sign-in screen is always displayed first, even if a valid session is found. This ensures that users log in each time they access the site.
- `prompt=create`: The traditional registration screen (used for creating new account) is always displayed first. Note, however, that the Sign In link isn’t found on the traditional registration screen. That means that setting the prompt to create represents a dead-end for existing users: they don’t need to create account, but they can’t log on using their existing account.

Source: https://janrain-education-center.knowledgeowl.com/home/authorization-code-pkce-for-mobile-apps. (There is a nice little video down below explaining how each of it works.)

But setting this parameter has no effect on Laravel Passport.

**How can I achieve this `prompt=login` behavior?**

BTW, my API logout function looks like this:

```php
public function logout()
{
    Auth::user()->token()->revoke();
    $tokenId = Auth::user()->token()->id;

    $refreshTokenRepository = app('Laravel\Passport\RefreshTokenRepository');
    $refreshTokenRepository->revokeRefreshTokensByAccessTokenId($tokenId);

    return response(null, Response::HTTP_NO_CONTENT);
}
```

jasperjorna:
[...]

PR #1557, opened 2022-08-12
[10.x] Add a route and a controller to revoke current token

hafezdivandari:
Related to #1466 and #1389

This PR adds a new `DELETE /oauth/token` route to revoke the current access token being used by the user. This will be used for logout. I think it's good to have a built-in method to be used in logout process.


taylorotwell:

PR #1564, opened 2022-08-29
[11.x] Allow for bootstrapping without loading routes

axlon:
Passport's bootstrapping of routes was changed recently (https://github.com/laravel/passport/pull/1464), this PR aims to make it a little easier for developers to disable the routes that ship with Passport (which currently requires disabling its service provider and replacing it with a custom one).

PS. I lifted this code from Fortify

### DIFF
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -50,13 +50,15 @@ class PassportServiceProvider extends ServiceProvider
      */
     protected function registerRoutes()
     {
-        Route::group([
-            'as' => 'passport.',
-            'prefix' => config('passport.path', 'oauth'),
-            'namespace' => 'Laravel\Passport\Http\Controllers',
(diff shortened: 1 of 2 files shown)


src/PassportServiceProvider.php
```diff
@@ -50,6 +50,10 @@ public function boot()
      */
     protected function registerRoutes()
     {
+        if (! Passport::$registersRoutes) {
+            return;
+        }
+
```
driesvints, line 56: Can you move the Route group inside the if statement here and negate the check?

driesvints (changes requested):

src/PassportServiceProvider.php
axlon: Done! @driesvints

driesvints (approved):
Thanks!

PR #1571, opened 2022-09-06
[11.x] Support prompting login when redirecting for authorization

hafezdivandari:
Related to #1389

As explained on #1567 and in addition to #1569, this PR adds support for `prompt=login` that causes the authorization server to display the login page first.

This PR also causes `prompt=none` to return an error if the user is not already authenticated instead of prompting the login page, which is the expected behavior (to prompt nothing).

PS: I'll send a PR to laravel/docs to explain `prompt` after this.

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -188,7 +188,9 @@ class AuthorizationController
                     ? $authRequest->getClient()->getRedirectUri()[0]
                     : $authRequest->getClient()->getRedirectUri());
 
-            $uri = $uri.(str_contains($uri, '?') ? '&' : '?').'state='.$authRequest->getState();
+            $separator = $authRequest->getGrantTypeId() === 'implicit' ? '#' : '?';
+
+            $uri = $uri.(str_contains($uri, $separator) ? '&' : $separator).'state='.$authRequest->getState();
 
             return $this->withErrorHandling(function () use ($uri) {
                 throw OAuthServerException::accessDenied('Unauthenticated', $uri);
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/AuthorizationController.php
```diff
@@ -31,17 +34,28 @@ class AuthorizationController
...
      *
      * @param  \League\OAuth2\Server\AuthorizationServer  $server
      * @param  \Illuminate\Contracts\Routing\ResponseFactory  $response
+     * @param  \Illuminate\Contracts\Auth\StatefulGuard  $guard
      * @return void
      */
-    public function __construct(AuthorizationServer $server, ResponseFactory $response)
+    public function __construct(AuthorizationServer $server,
+                                ResponseFactory $response,
+                                StatefulGuard $guard)
```
taylorotwell, line 54: Do we resolve StatefulGuard like this anywhere else in Passport? How do we know this is the right guard or is even valid?

hafezdivandari: No we don't, what if I add the following to `register` method of the `PassportServiceProvider`?
```php
$this->app->bind(StatefulGuard::class, function () {
    return Auth::guard();
});
```

taylorotwell:
I honestly just don't have the bandwidth to take this on right now and review it. I don't know the implications of the StatefulGuard stuff, etc.

src/Http/Controllers/AuthorizationController.php
hafezdivandari: @driesvints Can you assist on this, please?

Issue #1568, opened 2022-09-02
Allow clients to generate a new client secret.


khan-zia:
Alright, so I looked this up in the repository before opening this issue and I think this hasn't been proposed and closed before so here I am proposing it.

I understand that most of the time clients are supposed to protect `secrets` and never lose them. However, I think there will be genuine use cases when a client wants to simply generate a fresh `secret` value without having to create an entirely new OAuth client. Examples include but are not limited to, parting ways with a company's employee that had access to the `secret` value, accidentally exposing the value, having a company-wide secrets rotation policy, etc.

It would be even better if client accounts can use more than one `secret` value to ensure zero downtime for their applications. This is until they replace the new `secret`, their app will be using the old one which they can safely discard later. However, this would be a big change and would require the database schema to change. Simply allowing a new `secret` to be generated would be a great start.

If you guys like the idea, I can put together a PR in the coming days. Cheers!

@taylorotwell @driesvints

driesvints:
I'm not sure what the difference would be with generation a new OAuth client? The tokens related to the client would be invalid anyway so the old client can simply be closed off.

khan-zia:
AFAIK, the tokens will not be invalidated if you simply update the `secret` value for a client. In fact, I just confirmed this before writing this comment. With this feature, I see the following benefits:

- It's just cumbersome to create a new client every time. You would be entering the OAuth values i.e. `redirect_uri` every time.
- This is especially cumbersome with heavily customized implementations. For e.g. we have overridden the `oauth_clients` model and require a dozen other inputs when creating a client.
- It saves the users of the client application from re-consent. If a new client is created, all its users will be forced to authorize it again.
- As I stated above, all existing tokens will continue to work.

PR #1576, opened 2022-09-15
[11.x] Add the ability to regenerate client secret

khan-zia:
As discussed several days ago at #1568 , this PR allows users to generate a new client secret. These changes are fully backward compatible and the tests are included.

[This comment](https://github.com/laravel/passport/issues/1568#issuecomment-1237116832) explains the benefits of having this feature.

### Considerations
- Updating a client secret will require an authenticated user to provide the previous client secret. This stays consistent with the behavior of getting a new auth token which requires a refresh token.
- I have covered 3 different scenarios of the feature that I added in one test while ensuring everything is still neat, readable, and comprehensive. This allowed me to avoid repeating a dozen lines of test boilerplate code. It may be split to multiple tests if preferred.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -54,6 +54,11 @@ Route::middleware(['web', 'auth'])->group(function () {
         'as' => 'clients.update',
     ]);
 
+    Route::put('/clients/{client_id}/generate-secret', [
+        'uses' => 'ClientController@generateSecret',
+        'as' => 'clients.update.secret',
+    ]);
+
     Route::delete('/clients/{client_id}', [
         'uses' => 'ClientController@destroy',
         'as' => 'clients.destroy',
--- a/src/Bridge/ClientRepository.php
+++ b/src/Bridge/ClientRepository.php
@@ -97,7 +97,7 @@ class ClientRepository implements ClientRepositoryInterface
      * @param  string  $storedHash
      * @return bool
      */
-    protected function verifySecret($clientSecret, $storedHash)
+    public function verifySecret($clientSecret, $storedHash)
     {
         return Passport::$hashesClientSecrets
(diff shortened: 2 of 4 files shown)


taylorotwell:
Add this to your own application if you need it please.

khan-zia:
@taylorotwell a `regenerateSecret` method in the `clientRepository` has been hanging around for over 6 years that was added by you. I simply made use of that otherwise completely unused piece of code. Anyway, this was my second attempt to contribute to the framework I have used for years and both the times, hours of my efforts have ended by, I would say a not so satisfactory reason. It's certainly disappointing.

driesvints:
@khan-zia I'm sorry we couldn't accept your contribution. Taylor just wants to be mindful about what gets into our codebases because in the end we're the ones who have to maintain everything that gets accepted. Please remember that it's your own choice to contribute, we don't force you to do that. Likewise it's our own choice to decline code if we want. That might sound harsh but in the end we need to be wary of what we accept or not. You can still use the code in your own codebase so no time is wasted really.

khan-zia:
@driesvints I completely agree. It's not a problem if a contribution is rejected, in fact, it's absolutely necessary to keep the codebase clean and secure. It felt odd and a bit disappointing in my case because first, I hoped your thought process would resonate with Taylor's and you and I agreed on the issue for this thing as a feature. I did not just through a PR. Secondly, when you see a genuine effort by someone, writing a couple of more lines with a nice reason for rejection if possible, would always keep everyone happy. The same happened to me a year ago when I proposed that Laravel passport should utilize hashing algo from the main app instead of hard coding `bcrypt` which is being phased out anyway. I am still beyond thankful for this wonderful framework 👍

PR #1581, opened 2022-10-08
[11.x] Removed deprecated `dates` property from `Token` model

siarheipashkevich:


taylorotwell:
Is there any reason to remove this on a patch release. I assume nothing is broken?

PR #1584, opened 2022-10-13
Added passport:ready console command

nhaynes:
Added a new console command, `passport:ready` which can be used by automated deployments to detect if `passport:install` has already been run in an environment or not.

This solves an issue reported here: https://laracasts.com/discuss/channels/laravel/running-laravel-passport-install-on-production-server

---

The issue is primarily with the `oauth_clients`. If you run `passport:install` each time you deploy, it adds new clients and the tokens created with the older clients are no longer valid. This just gives a simple way to check if the environment has already been setup.

```bash
    php artisan passport:ready

    if [[ $? -eq 1 ]] ; then
        echo "installing passport"
        php artisan passport:install
    fi
```

---

Additionally, I refactored the test `PassportTest::test_missing_personal_access_client_is_reported` to restore the **PersonalAccessClientModel** after its assertion because the static reference to **PersonalAccessClientStub** was being carried over to subsequent tests.

### DIFF
--- a/src/Console/ReadyCommand.php
+++ b/src/Console/ReadyCommand.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Laravel\Passport\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Database\QueryException;
+use Illuminate\Support\Facades\DB;
+use Laravel\Passport\Passport;
+
+class ReadyCommand extends Command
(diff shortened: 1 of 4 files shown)


taylorotwell:
Hi! I suggest just writing this command as you need for your own application via `make:command`. Thanks!

PR #1591, opened 2022-10-31
update group middleware

Ashwanimgs:
Please update group middleware as it is giving errors in lumen 9.x version.


driesvints:
We don't consider Passport to be compatible with Lumen. Lumen is also not developed anymore. Please switch to Laravel.

Issue #1598, opened 2022-11-19
Redirecting issue after adding "Support prompting login when redirecting for authorization"


adrorocker:
- Passport Version: 11.3.0
- Laravel Version: 9.39.0
- PHP Version: 8.1.12
- Database Driver & Version: N/A

### Description:

[This commit](https://github.com/laravel/passport/commit/741595738b80f64ec35ad82de477adee925fff7e) introduced a change in behavior. When the Laravel Application does not use the default `login` named route (or even Laravel to log the user in) the login flow gets broken, now by default passport will redirect you  to that `login` route.  If the applications does not have the login route it will throw an error complaining that the login route is not defined.

Prior this commit the redirection was handle by the [Authenticate](https://github.com/laravel/laravel/blob/040d548810dd2d517bbd5cc792818b5d0429a0fc/app/Http/Middleware/Authenticate.php#L18) middleware and not the [AuthorizationController](https://github.com/laravel/passport/blob/741595738b80f64ec35ad82de477adee925fff7e/src/Http/Controllers/AuthorizationController.php#L218) 

I'm assuming that this is not an intended behavior since now the request gets passed the `Authenticate` middleware. This is not letting us customize the redirection url since the controller lives in the vendor dir. 

I found 3 solutions to this:

1. Create a redirection route

```php
Route::redirect('/login', 'https://another-auth-system.com/auth/login')->name('login');
```

2. Move the [/autorize](https://github.com/laravel/passport/commit/741595738b80f64ec35ad82de477adee925fff7e#diff-193e04aa1705f6f3e484d7efca8f45fe4d19d0ece3b6e6880d3ea070938a1fad) route back to group that uses the `['web', 'auth']` middlewares (Not sure what the implications of doing this will cause on what [this](https://github.com/laravel/passport/pull/1577) PR was suppose to addressed)

3. Add a config in the passport config file that let you customize where the exception should redirect . Something like 

[src/Http/Controllers/AuthorizationController.php](https://github.com/laravel/passport/pull/1577/files#diff-9fc1a4ecd0d38e60fac1081388d782df71dd7a5eb72d1e1cd9586df4dfd889ac) 

```php
    /**
     * Prompt the user to login by throwing an AuthenticationException.
     *
     * @param  \Illuminate\Http\Request  $request
     *
     * @throws \Illuminate\Auth\AuthenticationException
     */
    protected function promptForLogin($request)
    {
        $request->session()->put('promptedForLogin', true);

[...]

PR #1599, opened 2022-11-22
[11.x] Add a way to customize the `redirectTo` when prompt for login

adrorocker:
This PR is to address #1598.

Now you can customize the redirection setting up the `login_url` config on the `passport.php` config file. Example:

```php
<?php

return [
   ...

    'login_url' => '/auth/login',

];
```

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -46,4 +46,17 @@ return [
         'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET'),
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Authorize Login Redirection
+    |--------------------------------------------------------------------------
+    |
+    | By default, Passport will redirect to the `login` named route when the
+    | user is not authenticated. However, you can configure Passport to
+    | redirect to a specific URL when prompt for logging in.
+    |
+    */
+
+    'login_url' => null,
+
 ];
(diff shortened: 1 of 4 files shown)


config/passport.php
```diff
@@ -46,4 +46,17 @@
...
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Authorize Login Redirection
+    |--------------------------------------------------------------------------
+    |
+    | By default, Passport will redirect to the `login` named route when the
+    | user is not authenticated. However, you can configure Passport to
+    | redirect to a specific URL when prompt for logging in.
```
jayan-blutui, line 56:
```suggestion
    | redirect to a specific URL when prompted for logging in.
```

jayan-blutui (commented):
Just a small grammar fix 😃❤️

taylorotwell:
Is this really better than just defining a route named `login` that does whatever you want?

adrorocker:
I think in practice is the same, I just wanted to keep the logic in the "same place". Since Passport is the one bringing the change in behavior I wanted a way for Passport to override that it if necessary. But feel free to close this if you think it doesn’t bring any value. 

Thanks for the awesome work.

taylorotwell:
I think I'll just stick with the convention of defining a named `login` route to control login behavior. Thank you!

robertmarney:
#1704 would be resolved by a reboot of this PR.

For additional context, I feel larger/complex applications are the most likely audience of Passport (versus Sanctum), and are more likely to have multiple guards / manners of authentication. I think this would be in line with my understanding of the use cases of Passport vs Sanctum.

Issue #1602, opened 2022-12-02
AuthorizationController gets TokenGuard instead of StatefulGuard


pontoffeltier:
- Passport Version: 11.3.0
- Laravel Version: 9.42.2
- PHP Version: 8.1.13
- Database Driver & Version: mysql 8.0.30

### Description:
I followed the documentation for Passport to implement "Authorization Code Grant with PKCE". When I am trying to send a GET request to /oauth/authorize, I get the following error message:
```
Laravel\\Passport\\Http\\Controllers\\AuthorizationController::__construct(): Argument #3 ($guard) must be of type Illuminate\\Contracts\\Auth\\StatefulGuard, Laravel\\Passport\\Guards\\TokenGuard given
```
What I found in the source code was, that with the [PR 1577](https://github.com/laravel/passport/pull/1577) the `Laravel\\Passport\\Http\\Controllers\\AuthorizationController` now has a third parameter with the type of `Illuminate\\Contracts\\Auth\\StatefulGuard`. The problem is, the only guard in this package (which is bound in the `Laravel\Passport\PassportServiceProvider`) is the TokenGuard - which does not implement the StatefulGuard interface. 

**I totally expect that I am missing something important here, but for now I think this is a bug.**

### Steps To Reproduce:

As stated in the documentation:
- Install laravel/passport
- In `config/app.php` set `guards.api.driver` to `passport`
``` 
   'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],
        'api' => [
            'driver' => 'passport',
            'provider' => 'users',
        ],
    ],
```
- Create a client: `php artisan passport:client --public` with ID set to `1` and default values for name and redirect
- Make a GET request (I used postman) to `/oauth/authorize` - I tested with and without the query parameters, both give the same error response

driesvints:
@hafezdivandari can you weigh in here?

hafezdivandari:
@pontoffeltier I guess your default guard (the `auth.defaults.guard` config value) is not `web` right? 
Would you please add `'guard' => 'web'` to `config/passport.php` configuration file and try again?

hafezdivandari:
> I followed the documentation for Passport to implement "Authorization Code Grant with PKCE". When I am trying to send a GET request to /oauth/authorize, I get the following error message...

[...]

PR #1603, opened 2022-12-02
[11.x] Add auth guard to routes

hafezdivandari:
Related to #1602, We have already defined and used `passport.guard` config here: https://github.com/laravel/passport/blob/662325a90b8c3184c548eaf66ce971925f2b90dc/src/PassportServiceProvider.php#L140-L141

This PR adds a default value for `guard` to `config/passport.php` and also uses this guard on predefined Passport routes. This change is totally backward compatible and if `passport.guard` is not defined or `null`, the Passport will use `auth.defaults.guard` config value by default.

PS: The same approach is used on Fortify:
- https://github.com/laravel/fortify/blob/eb97f088c2a3e010e76cb58e9406427d1cee8de3/routes/routes.php
- https://github.com/laravel/fortify/blob/eb97f088c2a3e010e76cb58e9406427d1cee8de3/config/fortify.php#L6

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -4,6 +4,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Passport Guard
+    |--------------------------------------------------------------------------
+    |
+    | Here you may specify which authentication guard Passport will use when
+    | authenticating users. This value should correspond with one of your
+    | guards that is already present in your "auth" configuration file.
+    |
+    */
+
+    'guard' => 'web',
+
+    /*
+    |--------------------------------------------------------------------------
     | Encryption Keys
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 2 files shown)


driesvints:
This shouldn't be necessary? Passport has worked fine ever since without this config value.

hafezdivandari:
@driesvints This is necessary when user sets `auth.defaults.guard` to a guard which is not an instance of `StatefulGuard`, there should be a way to customize the guard on Passport just like Fortify.
Meanwhile, the `passport.guard` config is introduced by @taylorotwell on this commit: https://github.com/laravel/passport/commit/066eafe680cd0d8513ea921fdd24ae84b65aa3f6#diff-b9d9183efdbe0ebe4b06bc23eac77b61825ceba1144aadcd40a2071ac0eebd37

Issue #1604, opened 2022-12-09
Issue when use multi auth


michaelnabil230:
- Passport Version: 11.3
- Laravel Version: 9.19
- PHP Version: 8.1.13

### Description:
I have a problem when I create multiple authentications.

When I visit one of the links that have the user's authority and I make sure if this user is an admin or user using `auth()->check()` gives the result that the admin is true, although the token used in the request is a user's token, not an admin

I followed the instructions in the package's docs but to no avail.

I was done creating a unit test to try this point only.

### Steps To Reproduce:
https://github.com/michaelnabil230/laravel-passport-issue

michaelnabil230:
Can be seen in the unit test with the link:
https://github.com/michaelnabil230/laravel-passport-issue/blob/master/tests/Feature/HomeTest.php

Can also see the result of the unit test with the link:
https://github.com/michaelnabil230/laravel-passport-issue/actions/runs/3660881096

driesvints:
Multi auth isn't meant to be used with personal access tokens. It's meant to be used with the OAuth grants like the passport grant. Also, you're using the `passport:install` command at the beginning of your tests but that won't install the password grant for the admins provider. Please see the docs about that here: https://laravel.com/docs/9.x/passport#customizing-the-user-provider

PR #1606, opened 2022-12-10
Fix issue multi auth

michaelnabil230:
This PR is to fix the issue of multi-auth

See: https://github.com/laravel/passport/issues/1604

And I need help for fix the unit test and creating a new test for a multi-auth


driesvints:
As I said in https://github.com/laravel/passport/issues/1604 it's not meant to use multi-auth with personal access tokens. 

I also very much recommend to use Sanctum instead of Passport if you need personal access tokens: https://laravel.com/docs/9.x/sanctum

PR #1613, opened 2023-01-09
Refactor guards

m-bosch:
Hi,

We are currently implementing Passport in one of our projects. We had two guards in place `backend` and `frontend`, the `backend` guard is set as the default guard. We want to use Passport for the frontend users. 

So in the config we set the `guard` variable to `frontend`. Out of the box the AuthorizationController is now using `backend` guard, and the oauth routes are using the `frontend` guard. What we want is the AuthorizationController using the `guard` variable from the config and the oauth routes using the `api` guard. The docs mention creating this `api` guard in the installation steps so I guess `api` is a sensible default?

To achieve backwards compatibility maybe we can set `web` as default in the packages' routes file?


taylorotwell:

m-bosch:
Hi @taylorotwell,

In my opinion there has to be a solution for cases where the default guard isn't the guard you want to use for Passport. In this PR this bug.

To be clear this isn't a solution for multi-guards. Just for a non-default guard.

Maybe we can implement this otherwise, please let me know.

PR #1622, opened 2023-01-26
feat: allow XSRF token read from cookies

13DaGGeR:
Feature: try to acquire xsrf-token from cookies, in the case when it is absent from the headers.

Use case: [navigator.sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) does not support headers. When the system needs to gather statistical data, passport authentication currently does not work with the sendBeacon method.


taylorotwell:
This explicitly enables a CSRF vulnerability though I believe? The point of checking the header is that a foreign website is unable to insert this header. However, the browser will send all cookies with a request even if the request is initiated from a foreign website.

Issue #1624, opened 2023-01-27
Error creating a user's personal access token


lucaspanik:
- Passport Version: v11.3.0
- Laravel Version: v9.10.1
- PHP Version: 8.0.21
- Database Driver & Version:  mysqlnd 7.4.9

### Description:

Good morning, good afternoon, good evening dear.

#### [TL;DR]

Disparity between the sent id and the column used to search the DB for the client.

In my oauth_clients table I have the ID column auto incremented by the database and another column called client_id which is a UUID generated by config `config('passport.client_uuids')`.

PersonalAccessTokenFactory@createRequest uses the `->id` column forcibly instead of using the `->getKey()` method  (line 101)
https://github.com/laravel/passport/blob/b7bc60c9df4b7ba42bc193e36f392880865250ee/src/PersonalAccessTokenFactory.php#L87-L106

ClientRepository@find uses the `->getKeyName()` method to get the primary column defined in the client model.
https://github.com/laravel/passport/blob/b7bc60c9df4b7ba42bc193e36f392880865250ee/src/ClientRepository.php#L37-L48

---

Today I had an issue with manual token generation for a user.

https://laravel.com/docs/9.x/passport#managing-personal-access-tokens
```php
$user->createToken('Token Name')->accessToken
```

In my oauth_clients table I have the ID column auto incremented by the database and another column called client_id which is a UUID generated by config `config('passport.client_uuids')`.

My problem is in the PersonalAccessTokenFactory#101 file where it obtains the client model (\Laravel\Passport\Client) and defines that it will use the value of the id column `->id` instead of obtaining the value of the primary column with the method `->getKey()`
https://github.com/laravel/passport/blob/11.x/src/PersonalAccessTokenFactory.php#L101

Then there is client validation in PersonalAccessGrant.php
https://github.com/laravel/passport/blob/11.x/src/Bridge/PersonalAccessGrant.php#L21

Which in turn uses the `find()` method of the ClientRepository.php class; in this method, `->getKeyName()` from the client model is used instead of forcing the ID column as mentioned above, causing a disparity between the id sent and the column used to search for the client in the DB.
https://github.com/laravel/passport/blob/11.x/src/ClientRepository.php#L47

I believe that changing the way to get the primary_key value of the client model from `$client->id` to `$client->getKey()` in PersonalAccessTokenFactory@createRequest the problem will be fixed.

tested on passport version 10 also on:
https://github.com/laravel/passport/issues/1623

driesvints:
[...]

PR #1626, opened 2023-01-30
[11.x] Get model PK instead of forcibly id column

lucaspanik:
PR of the change presented here:
https://github.com/laravel/passport/issues/1624

This PR has the purpose of modifying the way that the oauth Client model is obtaining its primary key.

For most people it won't have any impact, for some people who use another column as the primary key (instead of ID) it will be fixed in this flow.

### DIFF
--- a/resources/views/authorize.blade.php
+++ b/resources/views/authorize.blade.php
@@ -68,7 +68,7 @@
                                 @csrf
 
                                 <input type="hidden" name="state" value="{{ $request->state }}">
-                                <input type="hidden" name="client_id" value="{{ $client->id }}">
+                                <input type="hidden" name="client_id" value="{{ $client->getKey() }}">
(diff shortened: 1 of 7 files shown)


taylorotwell:
Looks like tests are failing.

lucaspanik:
@taylorotwell 

I was wondering whether the change from `$client->id` to `$client->getKey()` should also be applied to the `authorize.blade.php` file, could you help me with that?
https://github.com/laravel/passport/blob/b7bc60c9df4b7ba42bc193e36f392880865250ee/resources/views/authorize.blade.php#L65-L86

driesvints:
@lucaspanik yes feel free to do so 👍

lucaspanik:
> @lucaspanik yes feel free to do so 👍

Thank you, I believe the PR is now ok for review. 👍

driesvints (approved):

PR #1628, opened 2023-02-07
[11.x] Add support for `EncryptCookies` middleware

axlon:
This PR adds a new option to Passport to not decrypt the JWT cookie when authenticating.

Currently when dealing with the JWT cookie Passport's behaviour is this:
- When creating the JWT cookie, Passport expects the `EncryptCookies` middleware to be active, and relies on it for encryption
- When reading the JWT cookie, Passport expects the `EncryptCookies` middleware to be inactive, it decrypts the cookie on its own

Because of this, implementing a middleware similar to Sanctum's [EnsureFrontendRequestsAreStateful](https://github.com/laravel/sanctum/blob/3.x/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php) is currently not possible, as it will cause the JWT token to be decrypted twice.

Note that this PR does not contain any breaking changes, Passport's current behaviour stays the default behaviour. This PR only adds the option to opt-out of this behaviour.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -295,8 +295,12 @@ class TokenGuard implements Guard
      */
     protected function decodeJwtTokenCookie($request)
     {
+        $jwt = $request->cookie(Passport::cookie());
+
         return (array) JWT::decode(
-            CookieValuePrefix::remove($this->encrypter->decrypt($request->cookie(Passport::cookie()), Passport::$unserializesCookies)),
+            Passport::$decryptsCookies
+                ? CookieValuePrefix::remove($this->encrypter->decrypt($jwt, Passport::$unserializesCookies))
+                : $jwt,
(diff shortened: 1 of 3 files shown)


taylorotwell:
@axlon I'm curious - how does Passport expect the EncryptCookies middleware to be inactive when reading the JWT cookie. How does it actually enforce that in the code?

axlon:
@taylorotwell it reads the cookie when trying to authenticate through the token guard, when it does this it decrypts the cookie similarly to how `EncryptCookies` does (this implicitly relies on `EncryptCookies` not being active, because if it was the cookie would have already been decrypted and trying to decrypt it again would cause an error):

https://github.com/laravel/passport/blob/31ed56954f044ffd6dcef237d5087955376693ff/src/Guards/TokenGuard.php#L296-L302

PR #1607, opened 2022-12-13
New authorization view response for custom views

JonErickson:
The following PR allows customization of the view used for authorization. While you can publish Passport's views and edit them to your needs, unfortunately, this limits you to only the blade. This PR follows Fortify's precedence and allows you to pass a callback or string to a new Passport::authorizationView() function. The following example is how one would return and render an Inertia component for use with authorization.

Example:
```
AppServiceProvider.php

public function register() {

    // Return and render our custom Inertia component for authorization
    Passport::authorizationView(function ($parameters) {
        return Inertia::render('Passport/Authorize', [
            'client' => $parameters['client']->id,
            'name' => $parameters['client']->name,
            'scopes' => $parameters['scopes'],
            'state' => $parameters['request']->state,
            'authToken' => $parameters['authToken'],
            'csrfToken' => csrf_token(),
        ])->toResponse($parameters['request']);
    });
}
```

PR #1629, opened 2023-02-08
[11.x] Custom authorization view response

JonErickson:
This is an updated PR to [#1607](https://github.com/laravel/passport/pull/1607)

The following PR allows customization of the view used for authorization. While you can publish Passport's views and edit them to your needs, unfortunately, this limits you to only the blade. This PR follows Fortify's precedence and allows you to pass a callback or string to a new Passport::authorizationView() function. The following example is how one would return and render an Inertia component for use with authorization.

``` php
Passport::authorizationView(function ($client, $user, $scopes, $request, $authToken) {
    return Inertia::render('passport/Authorize', [
        'client' => $client->id,
        'name' => $client->name,
        'scopes' => $scopes,
        'state' => $request->state,
        'authToken' => $authToken,
        'csrfToken' => csrf_token()
    ]);
});
```


src/Http/Controllers/AuthorizationController.php
```diff
@@ -45,12 +45,12 @@ class AuthorizationController
...
      *
      * @param  \League\OAuth2\Server\AuthorizationServer  $server
-     * @param  \Illuminate\Contracts\Routing\ResponseFactory  $response
+     * @param  \Laravel\Passport\Contracts\AuthorizationViewResponse  $response
      * @param  \Illuminate\Contracts\Auth\StatefulGuard  $guard
      * @return void
      */
     public function __construct(AuthorizationServer $server,
-                                ResponseFactory $response,
+                                AuthorizationViewResponse $response,
```
driesvints, line 53: You're tightening the contract here which is a breaking change.

JonErickson: Would you like me to change to the master branch?

taylorotwell: @driesvints do we allow or document people to modify this controller?

driesvints: I guess not 👍

driesvints (approved):

src/Contracts/AuthorizationViewResponse.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Laravel\Passport\Contracts;
+
+use Illuminate\Contracts\Support\Responsable;
+
+interface AuthorizationViewResponse extends Responsable
+{
+    /**
+     * @param $parameters
```
axlon, line 10:
```suggestion
     * @param array $parameters
```

src/Http/Controllers/AuthorizationController.php
```diff
@@ -65,7 +65,7 @@ public function __construct(AuthorizationServer $server,
      * @param  \Illuminate\Http\Request  $request
      * @param  \Laravel\Passport\ClientRepository  $clients
      * @param  \Laravel\Passport\TokenRepository  $tokens
-     * @return \Illuminate\Http\Response
+     * @return \Laravel\Passport\Contracts\AuthorizationViewResponse|mixed|void
```
axlon, line 68: According to the interface the `withParameters` method returns `mixed`

```suggestion
     * @return mixed
```

src/Http/Responses/AuthorizationViewResponse.php
```diff
@@ -0,0 +1,68 @@
...
+    public function __construct($view)
+    {
+        $this->view = $view;
+    }
+
+    /**
+     * Add parameters to response.
+     *
+     * @param $parameters
+     * @return $this|mixed
```
axlon, line 39:
```suggestion
     * @return $this
```

src/Http/Controllers/AuthorizationController.php
JonErickson: Should return an instance of AuthorizationViewResponse. Will removed mixed/void.

axlon: Not necessarily though, right? Because according to the interface the return type of `AuthorizationViewResponse` can be anything.

E.g. this would be valid:

```php
public function withParameters(...)
{
    return view('my-authorization-view');
}
```

JonErickson: That would technically work as the controller would be able to return a response from the view however it would fail the return type definition of AuthorizationViewResponse. Following Fortify's precedence, the return type should be an implementation of the contract.

```
/**
 * Show the login view.
 *
 * @return \Laravel\Fortify\Contracts\LoginViewResponse
 */
public function create(Request $request): LoginViewResponse
{
  return app(LoginViewResponse::class);
}
```

[     * @return \Laravel\Fortify\Contracts\LoginViewResponse](https://github.com/laravel/fortify/blob/9ba71f3e448ae44370bdfe72f19952e23b4d6191/src/Http/Controllers/AuthenticatedSessionController.php#L44)

axlon: If that's the case then `withParameters()` should return `$this` in the interface rather than `mixed`, but IMHO the way you have it set up now makes more sense as it allows people to return whatever they want (and Laravel supports)

src/Passport.php
```diff
@@ -715,4 +717,17 @@ public static function withoutCookieEncryption()
...
+
+    /**
+     * Specify which view should be used as the authorization view.
+     *
+     * @param  callable|string  $view
+     * @return void
+     */
+    public static function authorizationView($view)
+    {
+        app()->singleton(AuthorizationViewResponseContract::class, function ($app) use ($view) {
```
axlon, line 729: Passport doesn't require laravel/framework, so the `app()` helper might not be present

```suggestion
        \Illuminate\Container\Container::getInstance()->singleton(AuthorizationViewResponseContract::class, function ($app) use ($view) {
```

src/Http/Controllers/AuthorizationController.php
driesvints: `mixed` can never be combined with something else. I'd follow the contract.

src/Passport.php
driesvints: Every first party package is only compatible with the framework so this is fine.

JonErickson:
Updated doc blocks. Should be ready for merge pending any more requested changes.

driesvints:
@JonErickson seems this one has conflicts. Can you rebase?

JonErickson:
Can you reopen the PR? I need to redo the rebase and I reset the HEAD causing the PR to close.

JonErickson:
@driesvints done.

driesvints:
Thanks @JonErickson

PR #1630, opened 2023-02-10
[11.x] Update PurgeCommand.php

Safemood:
avoid repeating the word "items"


driesvints:
I feel like the current sentence reads a bit better as it's two separate things the user is informed about.

PR #1633, opened 2023-02-14
[11.x] Move AuthenticationException into the scope of Laravel Passport

chrispage1:
## The problem

After looking for an easy way to tap into Passport's behaviour to manipulate the destination path when unauthenticated, I found there wasn't a particularly clear route to do so.

When an authentication error occurs, the script simply throws an `Illuminate\Auth\AuthenticationException`, the same as any other authentication section within a Laravel project. When this generic is thrown, it's caught by the exception handler and simply redirected to `route('login')`.

I felt it'd be useful to have a simpler way to tap into this functionality and customise the redirect behaviour.

## The solution

To achieve this, I've simply created Passport's own AuthenticationException class that anyone can then catch further up the chain and then perform the logic required for their specific use case.

In my use case, I'll be returning them to the typical login route but with some additional query parameters such as `?type=oauth` for example, so that I can remove registration links from the form.

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -2,13 +2,13 @@
 
 namespace Laravel\Passport\Http\Controllers;
 
-use Illuminate\Auth\AuthenticationException;
 use Illuminate\Contracts\Auth\StatefulGuard;
 use Illuminate\Contracts\Routing\ResponseFactory;
 use Illuminate\Http\Request;
 use Illuminate\Support\Str;
 use Laravel\Passport\Bridge\User;
 use Laravel\Passport\ClientRepository;
+use Laravel\Passport\Exceptions\AuthenticationException;
 use Laravel\Passport\Passport;
 use Laravel\Passport\TokenRepository;
 use League\OAuth2\Server\AuthorizationServer;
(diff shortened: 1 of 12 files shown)


src/Http/Controllers/AuthorizationController.php
```diff
@@ -209,7 +209,7 @@ protected function denyRequest($authRequest, $user = null)
      *
      * @param  \Illuminate\Http\Request  $request
      *
-     * @throws \Illuminate\Auth\AuthenticationException
+     * @throws \Laravel\Passport\Exceptions\\AuthenticationException;
```
driesvints, line 212:
```suggestion
     * @throws \Laravel\Passport\Exceptions\AuthenticationException
```

chrispage1:
@driesvints thanks for the updates. I've also just applied the Style-CI patch.

siarheipashkevich:
@chrispage1 @driesvints these changes are breaking changes for minor version. My code is handled `Illuminate\Auth\AuthenticationException` exception and after the PR it's broken.

driesvints:
Thanks, I reverted this PR and tagged a patch release. @chrispage1 it's best that this is sent to master instead.

taylorotwell:
@siarheipashkevich can you describe how this broke your application? Wouldn't `catch` still work since this new exception extends `AuthenticationException`?

siarheipashkevich:
Sorry guys, I checked my code again and everything works fine with 11.8.0 version.

I thought that you changed the namespace and after that my parts of code will be broken. Below my places where I use checking on the exception:

First:
image

Second:
image
image

Sorry for the misunderstanding

driesvints:
Thanks for confirming that. I've re-released this PR.

PR #1642, opened 2023-02-28
Make `$userId` nullable in `ClientRepository->createPersonalAccessClient`

bram-pkg:
Since `createPersonalAccessClient` and `createPasswordGrantClient` are both called with `null` as the `$userId` and the migrations support this, it would be good to have the docblock in-line with the actual usage of the code.

I'm trying to set up a personal access client in my tests to issue API tokens, and this docblock is incorrect.

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -158,7 +158,7 @@ class ClientRepository
     /**
      * Store a new personal access token client.
      *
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks!

driesvints:
Thanks!

PR #1636, opened 2023-02-17
laravel10 The Eloquent model's deprecated $dates property has been removed.

TonyWong9527:
- Passport Version: 11.7.0
- Laravel Version: 10.0.3

### Description:
laravel 10 update:
image

passport token model problem:
image

PR #1644, opened 2023-03-15
[11.x] Removed deprecated `dates` property from `AuthCode` model

siarheipashkevich:

### DIFF
--- a/src/AuthCode.php
+++ b/src/AuthCode.php
@@ -34,15 +34,7 @@ class AuthCode extends Model
      */
     protected $casts = [
         'revoked' => 'bool',
-    ];
-
-    /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure we would do this on a patch release? /cc @driesvints

siarheipashkevich:
https://github.com/laravel/passport/pull/1636

On the Laravel 10 it will break application if I use expires_at and expect that this field is carbon instance

siarheipashkevich:
@driesvints could you check this PR please?

driesvints:
I think this is fine tbh. I really doubt people are overwriting this directly.

driesvints (approved):

PR #1645, opened 2023-03-15
[11.x] Removed deprecated `dates` property from `RefreshToken` model

siarheipashkevich:


taylorotwell:
Not sure we would do this on a patch release? /cc @driesvints

driesvints:
I think this is fine tbh. I really doubt people are overwriting this directly.

driesvints (approved):

PR #1651, opened 2023-04-05
O auth guest sessions

JamieSTV:
Give passport the ability to authenticate the client without having an associated user.


driesvints:
This PR seems to contain quite a few conflicts. Please re-attempt with a clean commit history.

PR #1656, opened 2023-04-25
[11.x] Fix the multiple authorization providers in Laravel Passport

michaelnabil230:
This PR adds support for multiple authorization providers in Laravel Passport. Currently, when a user logs in with the `admin` provider, the token is created with the `admin` provider as well, even if the user has access to other providers.

With this PR, users can now log in with any provider they have access to, and the token will be created with the corresponding provider. For example, if a user login with the `admin` provider but has access to the `user` provider as well, the token will be created with the `user` provider.

This is achieved by adding a new `provider` field to the `createToken` method in the `TokenFactory` class, which allows specifying the provider for the token. Additionally, the `AuthCode` and `RefreshToken` models now have a `provider` field to store the provider associated with the token.

Overall, this PR improves the flexibility and usability of Laravel Passport for applications with multiple authorization providers.

It complements the idea of (https://github.com/laravel/passport/pull/1655)

***I think we need to display an error exception if the auth provider is not available in the config file.***

### DIFF
--- a/src/ClientRepository.php
+++ b/src/ClientRepository.php
@@ -98,19 +98,20 @@ class ClientRepository
      */
     public function activeForUser($userId)
     {
-        return $this->forUser($userId)->reject(function ($client) {
-            return $client->revoked;
-        })->values();
+        return $this->forUser($userId)
+            ->reject(fn ($client) => $client->revoked)
+            ->values();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


taylorotwell:
I honestly can't handle anymore Passport customization PRs right now. I need a break from it. Feel free to fork Passport to meet your needs.

PR #1667, opened 2023-05-25
supoort php 8.1

djoudi:


yankewei:
The `^8.1` is not required, The `^8.0` is okay when you install it on the PHP 8.1 version.

PR #1679, opened 2023-07-24
Add default scopes to issued token

xdiegom:
Hi there,

This PR is related to passport default scopes. 

I was working on a project using Laravel Passport and after I register the default scopes in `AuthServiceProvider.php` with `Passport::setDefaultScope()` the returned token didn't have the scopes defined in the payload. Thus, I decided to make some changes which are:

1. Default scopes are now part of the token if they are defined with Passport::setDefaultScope()`.
2. Given the premise of the RFC about [Access Token Scope](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3), specially in this part
> "The authorization server MAY fully or partially ignore the scope requested by the client, based on the authorization server policy or the resource owner's instructions.  If the issued access token scope is different from the one requested by the client, the authorization server MUST include the "scope" response parameter to inform the client of the actual scope granted"

when validating scopes, Passport should add the `defaultScope` as part of the validation `hasScope()` and methods like `scopeIds()` and `scopes()`. For that matter, I added a new Passport static method called `useDefaultScopes()` that sets a boolean static attribute to true in case default scopes MAY want to be included in issued token.

For this, I am not quite sure if this method is useful and always add default scopes if `defaultScope` is not empty 🤔 trying to adhere to the RFC fragment I addressed in this numeral.

3. I believe that `Passport::$defaultScope` should follow the same data type as `Passport::$scopes` regarding of one of the reasons of what scopes are for: ["displaying the description on the authorization approval screen"](https://laravel.com/docs/10.x/passport#defining-scopes)

Hope this helps 😃, if this PR is approved, I think that the only part where Laravel Passport docs may change is here [Default Scope](https://laravel.com/docs/10.x/passport#default-scope) and also within an upgrade docs.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -22,11 +22,13 @@ class Passport
     public static $implicitGrantEnabled = false;
 
     /**
-     * The default scope.
+     * The default(s) scope(s) defined for the application.
      *
-     * @var string
+     * @var array
      */
-    public static $defaultScope;
+    public static $defaultScope = [
+        //
+    ];
 
     /**
      * All of the scopes defined for the application.
@@ -199,12 +201,12 @@ class Passport
     /**
      * Set the default scope(s). Multiple scopes may be an array or specified delimited by spaces.
      *
-     * @param  array|string  $scope
+     * @param  array $scopes
      * @return void
      */
-    public static function setDefaultScope($scope)
+    public static function setDefaultScope($scopes)
     {
-        static::$defaultScope = is_array($scope) ? implode(' ', $scope) : $scope;
+        static::$defaultScope = $scopes;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


taylorotwell:

xdiegom:
Hi @taylorotwell, 

Thanks for your feedback. 

Just realized that I didn't understand the documentation about Default Scopes until now and either `Passport::tokensCan()` and `Passport::defaultScope()` must co-exists in the AuthServiceProvider in order to work, in other words `Passport::setDefaultScope()` can't be defined in `AuthServiceProvider` by itself unless `Passport::tokensCan()` is declared/defined before, which is something that in the Default Scopes [documentation section](https://laravel.com/docs/10.x/passport#default-scope) does not say directly, it has to be inferred in the example and with knowledge of OAuth2, I guess.

I wanted to make it clear in order to definitely close this issue in case someone find this useful 🙌🏼

PR #1686, opened 2023-09-01
[11.x] Allow scope repository to be constructed without parameters

axlon:
This fixes a breaking change introduced by #1682


driesvints:
Thanks @axlon

Smoggert:
Thanks for the quick response to fix the BC. 👍🏼

PR #1693, opened 2023-10-15
Add return to revokeRefreshTokensByAccessTokenId method

aminkhoshzahmat:


aminkhoshzahmat (commented):
Add return to revokeRefreshTokensByAccessTokenId method

PR #1698, opened 2023-11-16
[11.x] Allow unsetting a user's access token

axlon:
Small PHPDoc fix to allow the following

```php
$user->withAccessToken(null);
```

https://github.com/laravel/passport/blob/6dab7d087ef933d6ca4cc0038d10571d8a886b8a/src/HasApiTokens.php#L9-L14


driesvints:
Thanks!

PR #1702, opened 2023-12-13
[12.x] Adds Laravel 11 support

nunomaduro:
This pull request adds Laravel 11 support to Passport.


nanaya:
it sure would be nice to have the removal of [`ignoreMigrations`](https://laravel.com/docs/10.x/passport#migration-customization) documented somewhere instead of having to see generic error after updating

PR #1705, opened 2024-01-10
[11.x] Add getGrantTypesAttribute method to fix Eloquent strict mode error

gdebrauwer:
Since Laravel v10.40, Eloquent strict mode throws an error because "grant_types" does not exist on the Client model. This is caused by https://github.com/laravel/framework/pull/49480. By adding a getter-method for the attribute, the error disappears

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -106,6 +106,16 @@ class Client extends Model
     }
 
     /**
+     * Get the grant types the client can use.
+     *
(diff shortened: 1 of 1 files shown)


taylorotwell:
Eloquent strict mode is the worst feature in Laravel.

mtawil:
@taylorotwell Please release a new version for this fix.

PR #1708, opened 2024-01-16
[Feature] Allow configuration of login redirect route via config

### LINKED ISSUE #1704: Allow for configuration of which login route to use
I am happy to do the leg work as required, but wanted to get a feeling for whether a PR would be acceptable before I began.

We have multiple authentication in our Laravel app, and for passport we require the users authenticate via a secondary login route (for clarity our primary login is `/login` and the desired passport login is `/sso/login`.

I believe if we were to extend those places where an `Laravel\Passport\Exceptions\AuthenticationException` is thrown to utilize the existing `redirectTo` property on the base `Illuminate\Auth\AuthenticationException` to direct the user to be logged in via the correct route. 

e.g.

```
class CheckClientCredentialsForAnyScope extends CheckCredentials
{
    /**
     * Validate token credentials.
     *
     * @param  \Laravel\Passport\Token  $token
     * @return void
     *
     * @throws \Laravel\Passport\Exceptions\AuthenticationException
     */
    protected function validateCredentials($token)
    {
        if (! $token) {
            throw new AuthenticationException(redirectTo: config('passport.login_route'));
        }
    }
```

I envisage the route name being an optional config value that can be configured in `passport.php`

robertmarney:
Closes #1704 

### Description:

Allow developers to override the opinion of where unauthenticated users are redirected to for login.

This is particularly useful in cases where the application has multiple guards in place.

### Implementation:

1. Introduce new config property to store the custom login route.
2. Introduce new `make` function on the existing  `Laravel\Passport\Exceptions\AuthenticationException` and use the existing `redirectTo` argument.
3. Replace all existing `throw new AuthenticationException` with `throw AuthenticationException::make()`
4. Update existing tests with the requisite mocks
5. Introduce new test for this explicit functionality


taylorotwell:
Catch the exception in your application's exception handler's `render` method and return the redirect from there.

PR #1709, opened 2024-01-17
Add getScopesAttribute method

uintaam:
Since Laravel v10.40, Eloquent strict mode throws an error because "scopes" does not exist on the Client model. This is caused by https://github.com/laravel/framework/pull/49480. By adding a getter-method for the attribute, the error disappears

Based of changes in https://github.com/laravel/passport/pull/1705 for "grant_types"


uintaam:
Hmm tests not passing do to some scopes not being excluded now. Looking.

uintaam:
Tests fixed

PR #1714, opened 2024-02-06
Add `path` option to publishable config file

selfsimilar:
All `config('passport.*')` values should be represented in the template config file. `path` was the only one missing.

In context, I have spent a good deal of time today trying to determine how to modify Passport's route prefix, and most web articles are before July 2021 when the PR to [refactor routes to dedicated file](https://github.com/laravel/passport/pull/1464) was accepted. It's not clear to me when the static `routes()` function was removed, but most online instructions reference using that in the `boot()` function of `AuthServiceProvider.php` to modify the route prefix (e.g. https://pownall.dev/post/prefix-laravel-passport-routes).

I am also planning to open a PR to update the [official Laravel Passport docs](https://laravel.com/docs/10.x/passport#overriding-routes), to clarify this there as well. But having this represented in the publishable config file is, imo, a necessary regardless so all calls to `config('passport.*')` in Passport's code are visible in the template file.

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -59,4 +59,15 @@ return [
         'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET'),
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Route prefix
+    |--------------------------------------------------------------------------
+    |
+    | All routes start with this. Default is `oauth`.
+    |
+    */
+
+    'path' => 'oauth',
+
 ];


driesvints:
I personally don't feel we should add this to the config. `/oauth` at the root of the path is a good default and general goto when using oauth authentication. Customizing this might indeed be needed in certain situations but those are edge cases at best.

selfsimilar:
I agree 100% that `/oauth` is the correct default and 99% of the time doesn't need to be changed. However, I would note the following:

- This PR is pretty benign in that it simply exposes what was hidden, but doesn't change any behavior. Perhaps adding to the docblock above the value to say something like "Usually does not need to be adjusted" would be helpful?
- There is currently nothing in the documentation to indicate that changing the prefix is possible. I've [opened a PR](https://github.com/laravel/docs/pull/9308) to fix that, but it took me some time wading through out-of-date articles before I started combing through the Passport code to determine the current way.
- Most people will not publish the config file unless they need to customize something, at which point you should trust that they will not simply change values on a whim. To hide controls after publishing the config file feels mistrustful of other developers, or malicious.
- Finally, a fairly popular package by [Joel Butcher](https://github.com/joelbutcher) called [Socialtream](https://github.com/joelbutcher/socialstream) (>390 stars) conflicts with the `/oauth` route. I have [updated the Socialstream doc](https://github.com/joelbutcher/socialstream-docs/pull/6)s to reflect the new way of updating Passport's prefix route, but suffice it to say that while changing the route is an edge case, it is certainly still encountered.

I hope you will still consider adding this to help save even a handful of developers time while working with this wonderful project.

Cheers

driesvints:
Thanks for that @selfsimilar. I guess what we can maybe do is add the config option but not document it. Let's see what @taylorotwell says.

PR #1721, opened 2024-02-15
[12.x] Use more secure key permissions

axlon:
Passport currently does not check filesystem permissions of the keys it uses, per https://github.com/laravel/passport/pull/454. Passport also doesn't set the correct permissions on the keys it creates.

This PR aims to have Passport use more secure permissions by default, whilst still allowing developers to opt out.

Passport will now:
- Automatically use stricter permissions when creating keys through the `passport:keys` command
- Check key permissions before using them
  - This may be toggled off by setting `Passport::$checkKeyPermissions` to `false`
  - This check is disabled on Windows, because it does not support this

### DIFF
--- a/src/Console/KeysCommand.php
+++ b/src/Console/KeysCommand.php
@@ -55,6 +55,11 @@ class KeysCommand extends Command
                 file_put_contents($privateKey, (string) $key);
             }
 
+            if (! windows_os()) {
+                chmod($publicKey, 0660);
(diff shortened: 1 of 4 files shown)


hafezdivandari:
The default behavior has been changed on this commit https://github.com/laravel/passport/commit/e063df36bef0f184f0f5ab5af8be8dd9d88b0651

Shouldn't we document this?

PR #1722, opened 2024-02-29
Check that properties `grant_types` and  `scopes` exist

uintaam:
Calling `isset` on a property that doesn't exist triggers strict mode violations.  We should check that the attribute isset exists instead.


crynobone:
Test are failing, marking this as draft.

uintaam:
Thanks just fixed the tests.

PR #1736, opened 2024-04-06
Don't ignore `force` option in InstallCommand

MrKrisKrisu:
The new Interactions from #1724 break some console interactions by scripts - this change should fix this while reusing the `--force` option.

It may be discussed whether to include `--no-interaction` here too, as this is mentioned in the `--help` section of this command.

```
Usage:
  passport:install [options]

Options:
      ...
      --force            Overwrite keys they already exist
      ...
      -n, --no-interaction   Do not ask any interactive question
      ...
```

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -31,7 +31,7 @@ class InstallCommand extends Command
      *
      * @return void
      */
-    public function handle()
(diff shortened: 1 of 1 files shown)


taylorotwell:
Doesn't `--no-interaction` do exactly what you want?

PR #1738, opened 2024-04-15
[12.x] Make Passport's database connection configurable

axlon:
This PR allows developers to configure Passport's database connection without having to extend all of Passport's models.

This change should be fully backwards compatible because:
- Omitting this configuration will have Laravel use the default connection of the application (which is the current behavior)
- If someone has extended one of (or multiple of) Passport's models and has overwritten the `$connection` property, that will take precedence over the config
- If someone has extended one of (or multiple of) Passport's models and has overwritten the `getConnectionName()` method this new logic won't be executed at all

### DIFF
--- a/config/passport.php
+++ b/config/passport.php
@@ -32,6 +32,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Passport Database Connection
+    |--------------------------------------------------------------------------
+    |
+    | By default, Passport's models will use your application's default
+    | database connection. If you wish to use a different connection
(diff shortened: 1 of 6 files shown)


axlon:
I just realized Passport's `ClientFactory` always points to Passport's version of the client model. If someone were to extend the Client model to adjust the connection (which is what prompted me to create this PR) it makes the client factory unusable.

This PR fixes that issue as well, but long term it probably makes more sense to make `ClientFactory` use `Passport::clientModel()` to always use the correct model

PR #1739, opened 2024-04-17
[12.x] Add `refreshToken` relation to `Token` model

gdebrauwer:
In my project, when I user logs out, I need to revoke the access token and the refresh token linked to that access token. Currently I have to this in the following way:
```php
auth()->user()->token()->revoke();
RefreshToken::where('access_token_id', auth()->user()->token()->id)->first()->revoke();
```
This PR makes that a bit easier by adding a 'refreshToken' relation to the access token model:
```php
auth()->user()->token()->revoke();
auth()->user()->token()->refreshToken->revoke();
```

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -58,7 +58,7 @@ class Token extends Model
     }
 
     /**
-     * Get the refresh token that the token belongs to.
+     * Get the refresh token associated with the token.
      *
      * @return \Illuminate\Database\Eloquent\Relations\HasOne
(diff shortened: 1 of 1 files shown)


alecpl:
A better idea would be `refreshTokens` method, as `revokeRefreshTokensByAccessTokenId` is plural.

And am I missing something? If an access token can have multiple refresh tokens, how `hasOne` can work properly?

PR #1740, opened 2024-04-17
[12.x] Create access and refresh token in `Passport::actingAs()`

gdebrauwer:
If you use `Passport::actingAs()` in a test while in the tested API route needs to revoke the authenticated user's current access token, then that causes an error in the test. That is because Eloquent can not revoke a token that does not exist in database.
```php
Passport::actingAs($user);

auth()->user()->token()->revoke(); // This currently causes an error
```
This PR fixes that by creating and saving an access and refresh token in the database.

I also create a refresh token because of #1739 so the following works as well:

```php
Passport::actingAs($user);

auth()->user()->token()->refreshToken->revoke();
```

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -6,6 +6,7 @@ use Carbon\Carbon;
 use DateInterval;
 use DateTimeInterface;
 use Illuminate\Contracts\Encryption\Encrypter;
+use Illuminate\Support\Str;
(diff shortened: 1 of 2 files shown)


taylorotwell:
This might be a breaking change to existing applications. I would recommend maybe creating the tokens yourself in the few tests that need to actually revoke. 👍

PR #1743, opened 2024-05-10
[12.x] Support rehash on login

gdebrauwer:
In Laravel 11, "automatic rehashing of user passwords when validating credentials" was added as a feature (https://github.com/laravel/framework/pull/48665). This PR ensures that this feature also works when a user provides its credentials to passport (this is the case when using the password oauth grant type)


taylorotwell:
I don't think we're recommending the password grant at this time. I'll try to revisit this at some point.

PR #1745, opened 2024-05-21
[13.x] Always hash client secret

hafezdivandari:
Clients' secret are now always hashed.

## Changes
* `Hash` facade has been used to hash and check the values.
* `Passport::$hashesClientSecrets` property has been removed.
* `Passport::hashClientSecrets()` method has been removed.


driesvints (approved):
Think this will be a good one to have as a default and the upgrade path isn't too hard.

src/Bridge/ClientRepository.php
```diff
@@ -15,12 +15,18 @@ class ClientRepository implements ClientRepositoryInterface
...
+    /**
+     * The hasher implementation.
+     */
+    protected Hasher $hasher;
+
     /**
      * Create a new repository instance.
      */
-    public function __construct(ClientModelRepository $clients)
+    public function __construct(ClientModelRepository $clients, Hasher $hasher)
```
siarheipashkevich, line 26: what do you think about https://stitcher.io/blog/constructor-promotion-in-php-8 ?

src/Client.php
```diff
@@ -129,13 +130,7 @@ public function setSecretAttribute($value)
...
         $this->plainSecret = $value;
 
-        if (is_null($value) || ! Passport::$hashesClientSecrets) {
-            $this->attributes['secret'] = $value;
-
-            return;
-        }
-
-        $this->attributes['secret'] = password_hash($value, PASSWORD_BCRYPT);
+        $this->attributes['secret'] = is_null($value) ? $value : Hash::make($value);
```
siarheipashkevich, line 133: `$this->attributes['secret'] = $value ?: Hash::make($value);` ?

src/Bridge/ClientRepository.php
hafezdivandari: I prefer to follow the same style of the file I'm changing.

PR #1747, opened 2024-05-27
[13.x] Fix issue with ClientRepository::handlesGrant method

AsmaShayea:
This pull request resolves an issue where calling createToken('admin_token')->accessToken resulted in a **"Call to undefined method" error due to the hasGrantType() method being incorrectly referenced in the ClientRepository**. The proposed fix updates the handlesGrant method in the ClientRepository to address this error by replacing the erroneous method call with the correct logic to handle grant types.

### **Changes:**
- Updated the handlesGrant method in the ClientRepository to resolve the "Call to undefined method" error.
- Replaced the incorrect hasGrantType() method call with logic to check grant types using is_array() and in_array().

**Replaced** 

> if (! $record->hasGrantType($grantType)) {
>             return false;
>         }

**with**
> if (is_array($record->grant_types) && ! in_array($grantType, $record->grant_types)) {
>             return false;
>         }

### **Testing:**
- Tested and verified that calling createToken('admin_token')->accessToken produces the expected result without errors.

### DIFF
--- a/src/Bridge/ClientRepository.php
+++ b/src/Bridge/ClientRepository.php
@@ -72,7 +72,7 @@ class ClientRepository implements ClientRepositoryInterface
      */
     protected function handlesGrant($record, $grantType)
     {
-        if (! $record->hasGrantType($grantType)) {
+        if (is_array($record->grant_types) && ! in_array($grantType, $record->grant_types)) {
             return false;
         }
 


driesvints:
Thank you but this doesn't really explains the issue. Could you offer a concrete example? Also if this is a bug fix it can go to 12.x

hafezdivandari:
@AsmaShayea Are you overriding the default `Client` model using `Passport::useClientModel()`? If yes, your custom `Client` model class should extend the Passport `Client` model [according to docs](https://laravel.com/docs/11.x/passport#overriding-default-models):

```php
use Laravel\Passport\Client as PassportClient;
 
class Client extends PassportClient
{
    // ...
}
```

taylorotwell:

Thanks!

PR #1751, opened 2024-05-30
[13.x] Fix determining revoked records

hafezdivandari:
This PR fixes how we determine if the specified record is revoked and adds integration tests for this.
### Auth Code
If the given auth code does not exist (removed), `Bridge/AuthCodeRepository::isAuthCodeRevoked()` returns `false` mistakenly.

### Access Token
`TokenRepository::isAccessTokenRevoked()` is using `find` to determine if the given access token is revoked, this causes performance issue as `TokenGaurd` calls `find` twice on every request.

### Refresh Token
`RefreshTokenRepository::isRefreshTokenRevoked()` also uses `find` to determine if the given refresh token is revoked.

### DIFF
--- a/src/Bridge/AuthCodeRepository.php
+++ b/src/Bridge/AuthCodeRepository.php
@@ -48,6 +48,6 @@ class AuthCodeRepository implements AuthCodeRepositoryInterface
      */
     public function isAuthCodeRevoked(string $codeId): bool
     {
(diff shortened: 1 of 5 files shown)


driesvints:
I guess this solves the long outstanding https://github.com/laravel/passport/issues/382 issue? If that's the case then we should definitely give this a go.

PR #1752, opened 2024-05-30
[13.x] Remove redundant auth code columns

hafezdivandari:
We are unnecessarily persisting `scopes`, `client_id`, and `user_id` to `oauth_auth_codes` table. These values are encrypted on the auth code itself and we have absolutely no usage for them.

### Why?
* Updating these values doesn't have any affect on the auth code, e.g. if one changes the `scopes` value, the auth consent form doesn't change, because the real scopes are decrypted from the code itself not from DB.
* The code is useless if the associated client or user have been deleted, so there is no need to manually delete them by `user_id` or `client_id`. Auth codes are short-lived ([always 10 mins](https://github.com/laravel/passport/blob/8ea1dd41745c89769fd8c4b207c4739eea707e95/src/PassportServiceProvider.php#L208C13-L208C38)) and will be expired and purged from DB after that.
* There is no usage for them, completely redundant. We just store them on DB and never retrieve these values.


taylorotwell:
Let's just leave it. Otherwise it's a pointless breaking change.

PR #1755, opened 2024-06-17
[13.x] Use the `oauth_scopes` property of the bearer token on `TokenGuard`

### LINKED ISSUE #382: TokenGuard query token twice

hafezdivandari:
Fixes #382 

As discussed [here](https://github.com/laravel/passport/pull/1751#issuecomment-2141482284), this PR fixes a long-standing issue #382.

The `TokenGuard` authenticates the user via bearer token and assigns this token to the authenticated user. This will be useful when the developer wants to check if the token has a given scope using the `tokenCan` method on the authenticated `App\Models\User` instance:

```php
$request->user()->tokenCan('place-orders')
```

The problem is that the `TokenGuard` guard (and also `CheckClientCredentials` middleware) unnecessarily query the DB to find the token's record and retrieve its scopes, but we have already decoded the token and we can simply use the `oauth_scopes` property instead.

This PR fixes the issue by separating the logic into new `Passport\AccessToken` class from the `Passport\Token` model, so we don't need to query the DB to find the token. This can be considered as performance optimization and also a security fix!

### Upgrade Guide
There shouldn't be any breaking change here, the `user()->tokenCan()` is the only documented functionality and works like before. The `user()->token()` returns `TransientToken` if the user is authenticated via the cookie just like before, but now returns `Passport\AccessToken` instance instead of `Passport\Token` instance when the user is authenticated via bearer token, but to avoid any breaking change, this class dynamically forwards method calls to `Passport\Token`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -26,6 +26,12 @@ Passport now always hashes client secrets using Laravel's `Hash` facade. If you 
 
 In light of this change, the `Passport::$hashesClientSecrets` property and `Passport::hashClientSecrets()` method has been removed.
 
+### The User's Token Instance
+
+PR: https://github.com/laravel/passport/pull/1755
+
+When authenticating users via bearer tokens, the `User` model's `token` method now returns an instance of `Laravel\Passport\AccessToken` class instead of `Laravel\Passport\Token`.
+
 ### Personal Access Client Table and Model Removal
 
 PR: https://github.com/laravel/passport/pull/1749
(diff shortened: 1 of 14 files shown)


taylorotwell:
Is it not possible for the scopes to be manipulated by the user if we are using the scopes present on the request attributes?

hafezdivandari:
@taylorotwell the `oauth_scopes` are not on the request attributes, but signed inside the bearer token. We should only trust the values on the signed / enceypted token itself, the user has authorized this client with these scopes. On the token guard, we decode the token and use `oauth_access_token_id` to determine if the token exists and not revoked,  `oauth_user_id` to find the user, `oauth_client_id` to find the client and we should use `oauth_scopes` to determine the authorized scopes too.

Issue #1491, opened 2021-10-15
[10.x] ClientFactory breaks when using UUIDs and quiet creation


axlon:
- Passport Version: 10.1.3
- Laravel Version: 8.63.0
- PHP Version: 8.0.10
- Database Driver & Version: n/a

### Description:

Laravel 8.59 introduced the ability to create models quietly (without events) using model factories. Because Laravel Passport currently uses a model event listener to set a UUID on the Client model, Passport's model factory will fail to create client models when using both quiet creation and the UUIDs option.

Example error:
```
Integrity constraint violation: 19 NOT NULL constraint failed: oauth_clients.id (SQL: insert into "oauth_clients" ("user_id", "name", "secret", "redirect", "personal_access_client", "password_client", "revoked", "updated_at", "created_at") values (?, Broekenmagazijn, $2y$10$SJO8Kazjw5qFI6gGkST81O1AJT7hsqBMS71CqdA/INkFDdHXkO1Je, http://www.dewit.com/perspiciatis-accusantium-voluptas-esse-aut-debitis-officia-velit, 0, 0, 0, 2021-10-15 07:41:31, 2021-10-15 07:41:31)
```

### Steps To Reproduce:

- Clone [this demo repository](https://github.com/axlon/passport-bug-demo)
- Run Composer install and create an .env file
- Run PHPUnit (its configured to use in-memory database), you should see the error immediately

### Possible fixes:

One way to address this would be to move the UUID check out of the model event listener and into the `ClientFactory` and `ClientCommand`. This would make Passport's client model behave like any other model that has a non-incremental primary key, in that you would always have to explicitly define it before inserting the model into the database. I propose making this change on master as people might be relying on current behaviour in their applications.

Another solution could be to build a fallback into the model factory, by leaving the model event listener as it is and also adding the check to the model factory. This would have the added benefit of being a non-breaking change (as far as I can tell), but the downside of having the same check in two places.

I'm able to make a PR for either change (or another fix) if any are deemed necessary/acceptable.

PR #1757, opened 2024-06-23
[13.x] Use unique IDs on client model

### LINKED ISSUE #1706: Hope to support ULID
Hope to support ULID

hafezdivandari:
Uses `\Illuminate\Database\Eloquent\Concerns\HasUniqueIds` trait's methods on `Client` model instead of the old way of generating / assigning unique ID to primary key. It's now much easier to override `newUniqueId` and use ULID for example (fixes #1706) also avoid the issues like #1491.


src/Client.php
```diff
@@ -59,19 +59,16 @@ class Client extends Model
...
-        parent::boot();
+        parent::__construct($attributes);
 
-        static::creating(function ($model) {
-            if (Passport::clientUuids()) {
-                $model->{$model->getKeyName()} = $model->{$model->getKeyName()} ?: (string) Str::orderedUuid();
-            }
-        });
+        $this->usesUniqueIds = Passport::clientUuids();
     }
```
driesvints, line 72: We don't recommend overwriting a model's constructor. Can we do this somehow else? maybe with a `bootingX` method or something?

hafezdivandari: boot methods are static but `$this->usesUniqueIds` is not, we can use `initializeX` method, are you ok with defining a new trait for this? tbh, I personally prefer to get rid of `Passport::clientUuids()` totally.

driesvints: We can't really get rid of that method unless we require older apps to fill out that database column which is quite a bit of a hassle for them. I guess we can give this one a shot.

hafezdivandari: @driesvints actually there is no need to fill out that DB column, `string(36)` can be used if they were using incremental ids, in this case they can keep old ids, and get uuid for new clients. what you think?

hafezdivandari: > unless we require older apps to fill out that database column

Not necessarily, they can keep the old IDs as is, and get uuid for new clients by using `string(36)` column type.

PR #1759, opened 2024-06-24
[13.x] Refactor migrations

hafezdivandari:
There is no breaking change on this PR, just better migrations for new apps. No need to change/modify columns for existing apps.

* Use `passport.connection` on migrations (Related to #1738)
* Use `foreignId` instead of `unsignedBigInteger`, makes the purpose of the column more clear.
* IDs are always 80 char long, more optimized storage.

### DIFF
--- a/database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
+++ b/database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
@@ -12,9 +12,9 @@ return new class extends Migration
     public function up(): void
     {
         Schema::create('oauth_auth_codes', function (Blueprint $table) {
-            $table->string('id', 100)->primary();
-            $table->unsignedBigInteger('user_id')->index();
-            $table->unsignedBigInteger('client_id');
+            $table->char('id', 80)->primary();
+            $table->foreignId('user_id');
+            $table->foreignId('client_id');
             $table->text('scopes')->nullable();
             $table->boolean('revoked');
             $table->dateTime('expires_at')->nullable();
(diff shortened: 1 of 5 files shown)


database/migrations/2016_06_01_000001_create_oauth_auth_codes_table.php
```diff
@@ -12,9 +12,9 @@
     public function up(): void
     {
         Schema::create('oauth_auth_codes', function (Blueprint $table) {
-            $table->string('id', 100)->primary();
-            $table->unsignedBigInteger('user_id')->index();
-            $table->unsignedBigInteger('client_id');
+            $table->char('id', 80)->primary();
+            $table->foreignId('user_id');
```
taylorotwell: Why no index here anymore?

hafezdivandari: Because the column itself is redundant #1752

siarheipashkevich:
> * Use `foreignId` instead of `unsignedBigInteger`, makes the purpose of the column more clear.

@hafezdivandari hey, could you please explain why it is more clear?

 We don't use any constraints for the foreign columns and it is just `unsignedBigInteger` as it was early, why did you change it?

hafezdivandari:
@siarheipashkevich `foreignId` is an alias for `unsignedBigInteger`. It makes the purpose of the column more clear because **that column supposed to be a foreign key** that references the client ID not just an unsigined big integer!

> We don't use any constraints for the foreign columns and it is just unsignedBigInteger as it was early, why did you change it?

You can manually do it if you want, it is just easier now. Why not by default? Maybe because not all DB driver (e.g MongoDB) has that.

siarheipashkevich:
@hafezdivandari I tried it in MySQL and there is not any foreign keys just bigint unsigned column.

hafezdivandari:
@sindicatoesp that is the expected behavior, as I said they are just identical, you can add `->constrained()` to have the foreign key constraint: https://laravel.com/docs/12.x/migrations#foreign-key-constraints

siarheipashkevich:
@hafezdivandari got it, thanks

PR #1761, opened 2024-06-30
[13.x] Force confidential PAT client

hafezdivandari:
This PR forces personal access token grant to be confidential with the [same approach as client credentials grant](https://github.com/thephpleague/oauth2-server/blob/2ed9e5f65045bebf9e99c33ef1558dcd6d0206b7/src/Grant/ClientCredentialsGrant.php#L41-L45).

### DIFF
--- a/src/Bridge/PersonalAccessGrant.php
+++ b/src/Bridge/PersonalAccessGrant.php
@@ -3,7 +3,9 @@
 namespace Laravel\Passport\Bridge;
 
 use DateInterval;
+use League\OAuth2\Server\Exception\OAuthServerException;
 use League\OAuth2\Server\Grant\AbstractGrant;
+use League\OAuth2\Server\RequestEvent;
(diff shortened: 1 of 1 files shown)


driesvints:
@hafezdivandari could perhaps help to explain why we'd want this?

hafezdivandari:
sure @driesvints, the "Personal access token" grant and "Client Credential" grant must be confidential, other grants confidentiality are optional, as I mentioned above the client credentials grant already [has this check](https://github.com/thephpleague/oauth2-server/blob/2ed9e5f65045bebf9e99c33ef1558dcd6d0206b7/src/Grant/ClientCredentialsGrant.php#L41-L45), but the extended PAT grant does not, we do this check on the [repository](https://github.com/laravel/passport/blob/8ea1dd41745c89769fd8c4b207c4739eea707e95/src/Bridge/ClientRepository.php#L83) but it's a good practice to also check this on grant to be consistent with the oauth server.

PR #1768, opened 2024-07-04
[13.x] Configure the user provider for PAT

hafezdivandari:
We have documented how to add a guard with 'passport' as driver and set its 'provider'. So it's common practice to have several guards and providers. This PR makes PAT factory respect the guard provider when issuing personal access tokens.

### Changes
* Add feature tests.
* Require `user_id` on `PersonalAccessGrant`.
* Add `getProvider()` method to `HasApiTokens` to get the user provider name.
* Make `PersonalAccessTokenFactory` to check for `config('passport.personal_access_client.{provider}')` before getting default PAT credentials from `config('passport.personal_access_client')`. For example, having `customers` auth provider:

    ```php
    // config/passport.php

    'personal_access_client' => [
        'id' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_ID'),
        'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET'),

        'customers' => [
            'id' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_CUSTOMER_ID'),
            'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_CUSTOMER_SECRET'),
        ]
    ],
    ```

### DIFF
--- a/src/PersonalAccessTokenFactory.php
+++ b/src/PersonalAccessTokenFactory.php
@@ -19,6 +19,13 @@ class PersonalAccessTokenFactory
     protected $server;
 
     /**
+     * The client repository instance.
+     *
+     * @var \Laravel\Passport\ClientRepository
+     */
+    protected $clients;
+
+    /**
      * The token repository instance.
      *
      * @var \Laravel\Passport\TokenRepository
@@ -36,15 +43,20 @@ class PersonalAccessTokenFactory
      * Create a new personal access token factory instance.
      *
      * @param  \League\OAuth2\Server\AuthorizationServer  $server
+     * @param  \Laravel\Passport\ClientRepository  $clients
      * @param  \Laravel\Passport\TokenRepository  $tokens
      * @param  \Lcobucci\JWT\Parser  $jwt
      * @return void
      */
-    public function __construct(AuthorizationServer $server, TokenRepository $tokens, JwtParser $jwt)
+    public function __construct(AuthorizationServer $server,
+                                ClientRepository $clients,
+                                TokenRepository $tokens,
+                                JwtParser $jwt)
     {
         $this->jwt = $jwt;
         $this->tokens = $tokens;
         $this->server = $server;
+        $this->clients = $clients;
     }
 
     /**
@@ -88,7 +100,9 @@ class PersonalAccessTokenFactory
             ? config("passport.personal_access_client.$provider", config('passport.personal_access_client'))
             : config('passport.personal_access_client');
 
-        if (! $config) {
+        $client = isset($config['id']) ? $this->clients->findActive($config['id']) : null;
+
+        if (! $client || ($client->provider && $client->provider !== $provider)) {
             throw new RuntimeException(
                 'Personal access client not found. Please create one and set the `PASSPORT_PERSONAL_ACCESS_CLIENT_ID` and `PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET` environment variables.'
             );
--- a/src/PersonalAccessTokenFactory.php
+++ b/src/PersonalAccessTokenFactory.php
@@ -53,12 +53,13 @@ class PersonalAccessTokenFactory
      * @param  mixed  $userId
      * @param  string  $name
      * @param  string[]  $scopes
+     * @param  string|null  $provider
      * @return \Laravel\Passport\PersonalAccessTokenResult
      */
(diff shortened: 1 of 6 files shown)


driesvints:
@hafezdivandari forgive me but I don't really understand why all of this is necessary. Passport should only issue tokens for one client? Why offer support for different guards? Could you give some practical use cases?

hafezdivandari:
@driesvints thank you for asking. We already support different guards and user providers not only on Passport but on the [entire Laravel auth system](https://laravel.com/docs/11.x/authentication#adding-custom-user-providers:~:text=Laravel%27s%20authentication%20facilities%20are%20made%20up%20of%20%22guards%22%20and%20%22providers%22).

On Passport, the `Client` model has `provider` property and the [token guard uses this `provider`](https://github.com/laravel/passport/blob/8ea1dd41745c89769fd8c4b207c4739eea707e95/src/Guards/TokenGuard.php#L186-L197) to retrieve the user.
You can add `HasApiTokens` trait to any authenticatable model, define a customized auth guard / provider and it works!

Sanctum has the same ability, that you can use `HasApiTokens` trait to any authenticatable model, but handles the auth differently - it has [`tokenable`](https://github.com/laravel/sanctum/blob/fb04acee0fb043750893211234a3103f9dce6517/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php#L16) property on its PAT model, that means you can issue tokens for different models / providers / guards.

What this PR does is that it fixes a bug on Passport PAT when using customized user provider. But the support for different guards is not about Passport or something that this PR offer.

driesvints:
Thanks @hafezdivandari. I think I might not fully understand yet. For me, there should only be the current `id` and `secret` keys on `personal_access_client` and nothing else. I don't yet understand how custom user providers come into play. How can there ever be a use case when we need to issue access tokens for a different ID and secret than that of the base one? All PAT should originate from the same one imo since it represents the main app.

hafezdivandari:
@driesvints let me explain that, assume that you have different guards / user providers like [this one](https://github.com/laravel/framework/discussions/49977). Then you issue token for each one:

```php
$user->createToken('user token');                  // uses default client to issue token
$businessUser->createToken('business user token'); // also uses the default client to issue token!
```

both these tokens have the same `oauth_client_id` that has `users` as `provider`. So when a business user tries to login, the `TokenGuards` retrieves a normal user instead and logins a wrong entity!

driesvints:
@hafezdivandari thanks for that. I fully understand now. I think the PR is okay in this case but just got one remark:

So for each user provider you need to run `php artisan passport:client --personal` and create a separate one?

Shouldn't the config then look like:

```php
'personal_access_client' => [
    'users' => [
        'id' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_ID'),
        'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET'),
    ],

    'customers' => [
        'id' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_CUSTOMER_ID'),
        'secret' => env('PASSPORT_PERSONAL_ACCESS_CLIENT_CUSTOMER_SECRET'),
    ],
],
```

hafezdivandari:
@driesvints for backward compatibility, it first look for the specified **provider** credentials, if not found it fallbacks to `personal_access_client.id` and `personal_access_client.secret` like before, for those who just use default guard / provider. So we don't need any upgrade guide entry for this change.

src/PersonalAccessTokenFactory.php
```diff
@@ -78,13 +91,18 @@ public function make($userId, string $name, array $scopes = [])
...
     {
-        $config = config('passport.personal_access_client');
+        $config = $provider
+            ? config("passport.personal_access_client.$provider", config('passport.personal_access_client'))
+            : config('passport.personal_access_client');
+
+        $client = isset($config['id']) ? $this->clients->findActive($config['id']) : null;
 
-        if (! $config) {
+        if (! $client || ($client->provider && $client->provider !== $provider)) {
```
hafezdivandari, line 105: To be the same as the check on `TokenGuard`:

https://github.com/laravel/passport/blob/8ea1dd41745c89769fd8c4b207c4739eea707e95/src/Guards/TokenGuard.php#L186-L188

driesvints:
Thanks for your work @hafezdivandari

driesvints (approved):

PR #1750, opened 2024-05-30
[13.x] Device Authorization Grant RFC8628

hafezdivandari:
This PR adds support for Device Authorization Grant RFC8628.

- [x] thephpleague/oauth2-server#1412 
- [x] thephpleague/oauth2-server#1410

### Grant type
- `urn:ietf:params:oauth:grant-type:device_code`
- [RFC8628](https://datatracker.ietf.org/doc/html/rfc8628)
- Fixes #903

### ORM
- `oauth_device_codes` table.
- `Laravel\Passport\DeviceCode` model.
- `Passport::useDeviceCodeModel()`

### Routes
- `POST /device/code`: Request device code / user code.
- `GET /device`: Display a form for entering the user code.
- `GET /device/authorize`: Display authorization consent.
- `POST /device/authorize`: Approve authorization.
- `DELETE /device/authorize`: Deny authorization.

### Views
- `Passport::deviceAuthorizationView()` ([Jestream / Inertia sample](https://github.com/laravel/jetstream/pull/1521/files#diff-f1c4b7fafa7d58b84af73bb6549ec11488a9f4639faa8f68677a8d48c8f93789)) ([Jestream / Livewire sample](https://github.com/laravel/jetstream/pull/1521/files#diff-c372a963621383fb69fb13c56ebdb1ca63a0fade8477801d01ff352f19471bce))
- `Passport::deviceUserCodeView()` ([Jestream / Inertia sample](https://github.com/laravel/jetstream/pull/1521/files#diff-2f842159c63cb5251fcab8ec8246b7c96625fb60be30a0f55e6c898de035d64b)) ([Jestream / Livewire sample](https://github.com/laravel/jetstream/pull/1521/files#diff-c5404af4abfe1e7d0f5fb79c61abfa850e66e0691ac1595224eef4ec562b1992))

### Commands
- New `--device` option on `passport:client` command.
- Ask user about enabling device flow on `passport:client` command.
- Purge expired / revoked device codes via `passport:purge` command.

## Device Authorization Grant

The OAuth2 device authorization grant allows browserless or limited input devices, such as TVs and game consoles, to obtain an access token by exchanging a "device code". When using device flow, the device client will instruct the user to use a secondary device, such as a computer or a smartphone and connect to your server where they will enter the provided "user code" and either approve or deny the access request.

[...]

PR #1771, opened 2024-07-09
[13.x] Make Passport headless (Support Laravel Jetstream and Breeze)

hafezdivandari:
- [x] Requires #1744 

Passport has a view for authorization / consent, [this view doesn't load correctly](https://github.com/laravel/passport/assets/56585913/de320101-d923-4472-8d8c-34d2fc386656) for many years (maybe since Mix / Bootstrap CSS are not installed on Laravel Skeleton by default). Also more views are needed for Device flow (#1750), makes user experience even worse!

This PR makes Passport headless just like Fortify because Passport doesn't have any pre-installed UI scaffolding. Therefore, we add support for Jetstream and Breeze that already have UI for different stacks:

* Laravel Jetstream: laravel/jetstream#1521
* Laravel Breeze: laravel/breeze#398
* Laravel UI: I have the PR [ready for Laravel UI](https://github.com/laravel/passport/assets/56585913/242a7291-f618-4608-a3d3-4c902db694e9) but OAuth option seems to be confusing for Laravel UI? as it doesn't call `install:api` during installation unlike Jetstream and Breeze. Let me know if you want a PR for this.

## Changes
* Makes Passport headless.
  * New `Passport::viewPrefix()` method (inspired by Fortify).
  * New `Passport::viewNamespace()` (inspired by Fortify).
* Add new methods to be compatible with Jetstream and interchangable with Sanctum:
  * New `Passport::defaultScopes()` method. 
  * New `Passport::validScopes()` method.
  * New `User::currentAccessToken()` as an alias for `User::token()` to be interchange with Sanctum.
* Fix view response and rename it to `SimpleViewResponse` (inspired by Fortify), to be generic for the future views.
* Fix `Client::firstParty()` method. We consider a client first-party when it doesn't belong to any user.


hafezdivandari:
Hi @driesvints sorry to bother you, I just wanted to ask your opinion before continuing my work on this.

I'm gonna add a `passport:ui` artisan command that installs stubs based on user UI preferences. Also I'm gonna add support for Jetstream "API Token Manager". What do you think?

driesvints:
Hey @hafezdivandari. That seems fine for us! Thank you so much for all your work on this 👏

inmanturbo:
@hafezdivandari You've done a great deal of solid work on this throughout the ecosystem and it's much appreciated.

I maintain some custom in house sso servers for a handleful of privately owned businesses, and they all use Passort for oauth. They are all running on forge provisioned servers and so we try keep them up to date to stay in line with current forge offerings.

In the past the way to use passport with something like breeze or jetstream was to publish the view and modify to our needs, or otherwise place our own stub in the `resources/views/vendor/passport/authorize.blade.php`.

How would this change affect these existing applications?

hafezdivandari:
Thank you @inmanturbo 

When the view is on the `resources/views/vendor/passport/authorize.blade.php` path, `Passport::viewNamespace('passport')` should do the work.

As stated on the PR description, more views are comming for the new device flow. Each view will have its own method to render the view (just like Fortify) e.g `Passport::authorizationView()`, useful for Inertia stack. Also `Passport::viewPrefix()` and `Passport::viewNamespace()` helper methods are available to set the path for all views, useful for Livewire stack.

You may also check the [upgrade guide entry](https://github.com/laravel/passport/pull/1771/files#diff-434625111321abb225365806085a3753cb854d098ed9b4b6dc2001235c7f7bc4) for more info.

inmanturbo (commented):
~~Why not just leave these here?~~

~~Seems like Jetstream, Breeze et al could all just use the actions from Passport since they would need to depend on passport in order to support it anyway. That way the classes can be used elsewhere as well without having to depend on jetstream, especially if the json api were to be deprecated, this would give some support for those implementing their own json api.~~

~~Similarly Fortify provides action classes for some of its features as well as interfaces for others to guide implementation.~~

hafezdivandari:
Hi @taylorotwell this PR might have missed your attention for review?

hafezdivandari:
Hi @taylorotwell this PR might have missed your attention for review?

PR #1772, opened 2024-07-09
[12.x] Fix purge command

### LINKED ISSUE #1770: Purge command using wrong query for expired tokens
### Passport Version

12.2.0

### Laravel Version

11.10

### PHP Version

8.3.7

### Database Driver & Version

MySQL 8.3.0

### Description

When you use the command `passport:purge` to purge all the revoked and expired tokens, if you use the argument `--hours`; the tokens that expire the same day as today; are not going to expire until the next day.

The problem is in the PurgeCommand class, that the query for controlling the expired tokens, is using whereDate and orWhereDate; instead of where and orWhere. With this condition, the hours argument is not taken into account, so, tokens that expired one hour ago, are going to work all day, because, there is an another error, that the expired tokens, still work and are not revoked.

Apart from this version of the library, there is a Laravel 9.52.5, with Passport 11.8.4, PHP 8.1.16 and MySQL 5.7.11 with the same error.

### Steps To Reproduce

1. Create a token with a expires_at date of today.
2. Change the date manually or wait until the token is expired.
3. Use the `passport:purge --hours=1` command to purge all revoked and expired tokens for more than 1 hour.
4. Check that the revoked tokens are purged, but the expired token not.

Creating a scheduled task to execute this instead of the purge command, works:

``` php
$expired = Carbon::now();

Passport::token()->where('revoked', 1)->orWhere('expires_at', '<', $expired)->delete();
Passport::authCode()->where('revoked', 1)->orWhere('expires_at', '<', $expired)->delete();
Passport::refreshToken()->where('revoked', 1)->orWhere('expires_at', '<', $expired)->delete();
```

The queries are the same as the PurgeCommand class, lines 41-43. The change has to be done also in lines 55-57.

hafezdivandari:
Fixes #1770

The `passport:purge` command was not working properly with `--hours` option. This PR fixes this and adds some tests for the command.

### DIFF
--- a/src/Console/PurgeCommand.php
+++ b/src/Console/PurgeCommand.php
@@ -38,10 +38,9 @@ class PurgeCommand extends Command
             ? Carbon::now()->subHours($this->option('hours'))
             : false;
 
-        $constraint = function (Builder $query) use ($revoked, $expired) {
-            $query->when($revoked, fn () => $query->orWhere('revoked', true))
-                ->when($expired, fn () => $query->orWhere('expires_at', '<', $expired));
-        };
+        $constraint = fn (Builder $query) => $query
+            ->when($revoked, fn () => $query->orWhere('revoked', true))
+            ->when($expired, fn () => $query->orWhere('expires_at', '<', $expired));
 
         Passport::token()->where($constraint)->delete();
         Passport::authCode()->where($constraint)->delete();
--- a/src/Console/PurgeCommand.php
+++ b/src/Console/PurgeCommand.php
@@ -39,33 +39,18 @@ class PurgeCommand extends Command
             : false;
 
         Passport::token()->where(function (Builder $query) use ($revoked, $expired) {
-            $query->when($revoked, function ($query) {
-                $query->orWhere('revoked', true);
-            });
-
-            $query->when($expired, function ($query, $expired) {
-                $query->orWhere('expires_at', '<', $expired);
-            });
+            $query->when($revoked, fn (Builder $query) => $query->orWhere('revoked', true));
+            $query->when($expired, fn (Builder $query) => $query->orWhere('expires_at', '<', $expired));
         })->delete();
 
         Passport::authCode()->where(function (Builder $query) use ($revoked, $expired) {
-            $query->when($revoked, function ($query) {
-                $query->orWhere('revoked', true);
-            });
-
-            $query->when($expired, function ($query, $expired) {
-                $query->orWhere('expires_at', '<', $expired);
-            });
+            $query->when($revoked, fn (Builder $query) => $query->orWhere('revoked', true));
+            $query->when($expired, fn (Builder $query) => $query->orWhere('expires_at', '<', $expired));
         })->delete();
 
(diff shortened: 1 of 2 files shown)


src/Console/PurgeCommand.php
```diff
@@ -32,33 +33,44 @@ class PurgeCommand extends Command
...
-            : Carbon::now()->subDays(7);
+            : false;
 
-        if (($this->option('revoked') && $this->option('expired')) ||
-            (! $this->option('revoked') && ! $this->option('expired'))) {
-            Passport::token()->where('revoked', 1)->orWhereDate('expires_at', '<', $expired)->delete();
-            Passport::authCode()->where('revoked', 1)->orWhereDate('expires_at', '<', $expired)->delete();
-            Passport::refreshToken()->where('revoked', 1)->orWhereDate('expires_at', '<', $expired)->delete();
+        Passport::token()->where(function (Builder $query) use ($revoked, $expired) {
+            $query->when($revoked, function ($query) {
```
siarheipashkevich: What do you think about this style?

```
$query->when($revoked, fn (Builder $query) => $query->orWhere('revoked', true));
```

hafezdivandari: I prefer that, but the `when` closure doesn't return the `query` instance [according to docs](https://laravel.com/docs/11.x/queries#conditional-clauses), using arrow functions it does!

siarheipashkevich: Got it, but I think we should add type for $query ?

hafezdivandari: Changed to arrow function as its seems ok according to the doc type, thanks:

https://github.com/laravel/framework/blob/68e88bbcee1e9b0b02ca267916ee25f45ab3935c/src/Illuminate/Conditionable/Traits/Conditionable.php#L17

src/Console/PurgeCommand.php
```diff
@@ -32,33 +33,29 @@ class PurgeCommand extends Command
...
-            Passport::authCode()->whereDate('expires_at', '<', $expired)->delete();
-            Passport::refreshToken()->whereDate('expires_at', '<', $expired)->delete();
-
-            $this->option('hours')
-                ? $this->components->info('Purged items expired for more than '.$this->option('hours').' hours.')
-                : $this->components->info('Purged items expired for more than seven days.');
-        }
+            : false;
+
+        Passport::token()->where(function (Builder $query) use ($revoked, $expired) {
```
siarheipashkevich:
```
Passport::token()
    ->where(function (Builder $query) use ($revoked, $expired) {
        $query->when($revoked, fn () => $query->orWhere('revoked', true))
            ->when($expired, fn () => $query->orWhere('expires_at', '<', $expired));
    })
    ->delete();
```

do you like this?

siarheipashkevich: you refactored )

hafezdivandari:
@driesvints would you please merge 12.x into 13.x?

hafezdivandari:
@driesvints would you please merge 12.x into 13.x? Thanks!

driesvints:
@hafezdivandari done.

PR #1773, opened 2024-07-19
[13.x] Fix user-token relations

hafezdivandari:
1.  `$user->getProviders()` has been added recently on PR #1768, this method actually must not return `null` in a healthy environment / configuration. Also, it is not possible to use Passport with a `database` provider. This PR fixes these 2 situations on this method.
2. The relation between the authenticable model (`User` model by default) and `Token` model depends on the client's `provider`. This PR makes sure we are retrieving the tokens associated with the right authenticable model and adds integration tests for this.
   - Note: On Sanctum [`user->tokens()` relation is morphable](https://github.com/laravel/sanctum/blob/3f57a4cd069ad9442ab2082ffafa4f4186e41ce8/src/HasApiTokens.php#L22-L25), so we know which model is the owner of the key, on Passport we should check the client's `provider` to know which model is the owner of the key.

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -3,6 +3,7 @@
 namespace Laravel\Passport;
 
 use Illuminate\Container\Container;
+use LogicException;
 
 trait HasApiTokens
 {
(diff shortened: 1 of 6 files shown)


src/HasApiTokens.php
```diff
@@ -70,22 +82,18 @@ public function createToken($name, array $scopes = [])
...
-            if (in_array($provider, $providers)
-                && (($config['driver'] === 'eloquent' && is_a($this, $config['model']))
-                    || ($config['driver'] === 'database' && $config['table'] === $this->getTable()))) {
+            if (in_array($provider, $providers) && $config['driver'] === 'eloquent' && is_a($this, $config['model'])) {
                 return $provider;
             }
         }
 
-        return null;
+        throw new LogicException('Unable to determine authentication provider for this model from configuration.');
```
taylorotwell, line 96: So what should they do in this case?

hafezdivandari: This should never happen, the authenticatable model should have a provider registered on `auth` configuration as we instruct the developer when [installing Passport](https://laravel.com/docs/11.x/passport#installation:~:text=Finally%2C%20in%20your,incoming%20API%20requests%3A)

PR #1780, opened 2024-08-18
[13.x] Improve issuing PATs

hafezdivandari:
This PR improves issuing personal access tokens:
* Reduces number of queries to retrieve the client from the DB from 3 to 1.
* Removes the necessity of redeclaring client credentials on `passport.personal_access_client` config property and environment variables for each user provider.
* Reorganizes personal access grant feature tests.
* Fully backward compatible and no upgrading steps is needed.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
  <td>

1. Create a client using `passport:client --personal` for each user provider.
1. Set client credentials on `passport.personal_access_token` config property [for each user provider](https://github.com/laravel/passport/pull/1768#issuecomment-2210487903).
1. Issue access token using `$user->createToken()`.
   1. Get client credentials for the given user provider from config.
   1. Retrieve the client from DB and validate provider.
   1. Pass `client_id`, `client_secret` and `user_id` to the grant.
   1. Retrieve the client from DB to validate the secret.
   1. Retrieve the client from DB to issue the token.

  </td>
  <td>

1. Create a client using `passport:client --personal` for each user provider.
1. Issue access token using `$user->createToken()`.
   1. Pass `provider` and `user_id` to the grant.
   1. Retrieve the client from DB for the given provider to issue the token.
 
</td>
</tr>
</table>

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -53,13 +53,13 @@ When authenticating users via bearer tokens, the `User` model's `token` method n
 
 ### Personal Access Client Table and Model Removal
 
-PR: https://github.com/laravel/passport/pull/1749
+PR: https://github.com/laravel/passport/pull/1749, https://github.com/laravel/passport/pull/1780
 
 Passport's `oauth_personal_access_clients` table has been redundant and unnecessary for several release cycles. Therefore, this release of Passport no longer interacts with this table or its corresponding model. If you wish, you may create a migration that drops this table:
 
     Schema::drop('oauth_personal_access_clients');
 
-In addition, the `Laravel\Passport\PersonalAccessClient` model, `Passport::$personalAccessClientModel` property, `Passport::usePersonalAccessClientModel()`, `Passport::personalAccessClientModel()`, and `Passport::personalAccessClient()` methods have been removed.
+In addition, the `passport.personal_access_client` config property, `Laravel\Passport\PersonalAccessClient` model, `Passport::$personalAccessClientModel` property, `Passport::usePersonalAccessClientModel()`, `Passport::personalAccessClientModel()`, and `Passport::personalAccessClient()` methods have been removed.
 
 ## Upgrading To 12.0 From 11.x
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -59,7 +59,7 @@ Passport's `oauth_personal_access_clients` table has been redundant and unnecess
 
     Schema::drop('oauth_personal_access_clients');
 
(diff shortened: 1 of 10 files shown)


autaut03:
@hafezdivandari Hey. This PR essentially removed the ability of passing in a custom `client_id` to the `personal_access` grant, which is useful when an application has multiple personal access clients. In our case, there's one that manages user-created access tokens, and another that's used for the login flow.

Was this removal really necessary? It was even [suggested before](https://github.com/laravel/passport/pull/1517) that `createToken` accepts an optional `$clientId`, so there's definitely at least a couple of us that would like that sort of functionality, and this PR further distances us from that.

hafezdivandari:
@autaut03 the ability is not removed, it's actually much simpler to use now. If you have multiple personal access clients for different user providers, Passport now automatically choose the right client (based on user provider) when creating a PAT.

You may refer to the tests here for some sample usage: https://github.com/laravel/passport/blob/13.x/tests/Feature/PersonalAccessGrantTest.php

autaut03:
@hafezdivandari The provider is the same. Only the clients differ - to be able to separate system-generated and user-created tokens when displaying them to the user.

E.g. a user creates a token (`POST /personal-access-tokens`, `client_id = 1`) and sees it in the list of their tokens (`GET /personal-access-tokens`, `client_id = 1`). Then the system creates a token using the `personal_access` grant for a different purpose (in our case - temporary access token for 2fa) with a different `client_id = 2`.

By separating the clients, we can still display only the user-created tokens to the user, and exclude ones created internally by the system. The user provider in both cases is the same and completely irrelevant.

hafezdivandari:
@autaut03 It's really strange (non-standard IMHO) use case, because normally there shouldn't be any "system-generated PATs", but if you want to do this for whatever reason, you are able to customize the `Laravel\Passport\ClientRepository` class and override the `personalAccessClient` method. Something like the following code:

```php
namespace App\Repositories;

class ClientRepository extends \Laravel\Passport\ClientRepository
{
    #[\Override]
    public function personalAccessClient(string $provider): Client
    {
        // return your client...
    }
}

// Then in you app service provider....
$this->app->when(\Laravel\Passport\Bridge\PersonalAccessGrant::class)
            ->needs(\Laravel\Passport\ClientRepository::class)
            ->give(fn () => new \App\Repositories\ClientRepository);
```

siarheipashkevich:
@autaut03 hello, did you find the solution?

autaut03:
@siarheipashkevich I wasted a lot of time trying to use Passport for this but after fighting it for a while I decided to just skip grants altogether and created tokens directly through `AccessTokenRepository`:

```php
$accessToken = $this->accessTokenRepository->getNewToken($client, $scopes, $user->id);
// set it, private key, expiration and other attributes
$this->accessTokenRepository->persistNewAccessToken($accessToken);
```

Same thing for refresh tokens if you need those. This avoided all the `AuthorizationServer` and grant shenanigans.

PR #1786, opened 2024-09-24
retrieve primary key identifier in AuthCodeRepository & TokenRepository

JorrinKievit:
I am using custom table and column names for all Passport created tables. Everything works using the class overrides, except for these particular queries in AuthCodeRepository and TokenRepository. I use mutators to rewrite some model attributes, but that was not working for both. Now this query respects the set primary key on the authCode & token model instead. I am fairly new to Laravel, so there might be a better solution 😄.

This code also exists in 13.x, so I guess this should be merged into 13.x too?


taylorotwell:

PR #1790, opened 2024-10-01
[13.x] Make revoking refresh tokens optional

hafezdivandari:
- [x] thephpleague/oauth2-server#1449

This PR introduces a new `Passport::$revokeRefreshTokenAfterUse` property, in line with the new `GrantTypeInterface::revokeRefreshTokens()` function added in `oauth2-server` v9.0 useful for safer token rotation.

Disabling refresh token revocation can be useful if you want the client to manually revoke the old refresh token only after it successfully receives the new access token and refresh token.

Consider the following scenarios:

### Refresh Token Revocation Enabled (Default behavior)
1. The client requests to refresh an access token.
2. The old access token is revoked.
3. The old refresh token is revoked.
4. A new access token and refresh token are issued in response.
5. The client loses network connection and does not receive the new access and refresh tokens.
6. The client cannot use the old refresh token because it has already been revoked.
7. Dead end!

### Refresh Token Revocation Disabled
1. The client requests to refresh an access token.
2. The old access token is revoked.
3. A new access token and refresh token are issued in response.
4. The client loses network connection and does not receive the new access and refresh tokens.
5. The client can use the old refresh token again to receive new tokens.
6. The client manually requests the old refresh token's revocation after successfully receiving the new one.


ezequidias:
Hello @hafezdivandari , how are you? I'm following the new improvements that are being made to Laravel Passport! I'm very excited about the new version of Laravel Passport!

I'm happy that Laravel Passport is receiving significant improvements! Congratulations for all your support and dedication

I would like to know if there is a simpler way to add CSRF together with Laravel Passport!

Laravel Sanctum is session-based and with that it is easily possible to add CSRF! Would it be possible to add CSRF to Laravel Passport more easily?

hafezdivandari:
Hi @ezequidias, thank you for your comment. It's better to discuss about CSRF on [Laravel discussion](https://github.com/laravel/framework/discussions) board instead, as it's not related to this PR.

mhdcodes:
:+1: 

We're encountering this exact issue with Next.js where, due to a canceled request, the client ends up stuck with an old, invalid refresh token, even though the backend has already generated a new pair of access/refresh tokens.

mhdcodes (commented):
:fire:

hettiger:
> 6. The client manually requests the old refresh token's revocation after successfully receiving the new one.

Hi @hafezdivandari,

I think the »Refresh Token Revocation Disabled« scenario needs an extra API endpoint so that clients can revoke a refresh token. Any idea, how to lookup the refresh token if the client does not have the refresh token id? I.e., client will send the encrypted refresh token. Implementing such an API endpoint, we'd probably have to decrypt it similar to https://github.com/thephpleague/oauth2-server/blob/aa002150cc1e503f557df6dd995f0974ed1d0e47/src/Grant/RefreshTokenGrant.php#L110-L117 and then revoke the token via the id from the decoded JSON. Am I missing something? Is there a clean "Passport" way for the lookup?

hafezdivandari:
Hi @hettiger you may refer to support for Token revocation #1858

hettiger:
> Hi @hettiger you may refer to support for Token revocation #1858

Thank you @hafezdivandari thats amazing 🚀

PR #1796, opened 2024-10-30
Include the passport guard when throwing exception

willrowe:
This passes the guard used for passport when throwing an authentication exception.

When this exception is thrown, the default behavior is to redirect to the login page. This login page is usually associated with the default `web` guard. If a custom guard is used for Passport, you may want to redirect to a different URL. Including the guard itself in the exception allows the exception handler to detect it and change the redirect.

### DIFF
--- a/src/Exceptions/AuthenticationException.php
+++ b/src/Exceptions/AuthenticationException.php
@@ -3,13 +3,14 @@
 namespace Laravel\Passport\Exceptions;
 
 use Illuminate\Auth\AuthenticationException as Exception;
+use Illuminate\Support\Facades\Config;
 
 class AuthenticationException extends Exception
 {
     public function __construct($message = 'Unauthenticated.', array $guards = null, $redirectTo = null)
     {
         if (is_null($guards)) {
-            $guards = [config('passport.guard')];
+            $guards = [Config::get('passport.guard')];
         }
 
         parent::__construct($message, $guards, $redirectTo);
--- a/src/Exceptions/AuthenticationException.php
+++ b/src/Exceptions/AuthenticationException.php
@@ -6,4 +6,12 @@ use Illuminate\Auth\AuthenticationException as Exception;
 
 class AuthenticationException extends Exception
 {
+    public function __construct($message = 'Unauthenticated.', array $guards = null, $redirectTo = null)
+    {
+        if (is_null($guards)) {
+            $guards = [config('passport.guard')];
+        }
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Tests are failing on this one.

willrowe:
Sorry, I got caught up with something and haven't been able to get back to this yet. Thanks for marking it as draft.

hafezdivandari:
The guard on `AuthenticationException` not always equals `passport.guard` but depends on the context. For example this exception is thrown on `CheckScopes` middleware, but it has nothing to do with `passport.guard`. Actually [this only makes sense on `AuthorizationController`](https://github.com/laravel/passport/blob/6594120ff686782af9af82d30498b7be117d937a/src/PassportServiceProvider.php#L131-L133):

```diff
// src/Http/Controllers/AuthorizationController.php

protected function promptForLogin(Request $request): never
{
    $request->session()->put('promptedForLogin', true);

-   throw new AuthenticationException;
+   throw new AuthenticationException(guards: isset($this->guard->name) ? [$this->guard->name] : []);
}
```

Note: `SessionGuard` is the only implementation of the `StatefulGuard` and it [has `name` property](https://github.com/laravel/framework/blob/61e221c18a89780ed5b51b59f04c0b4c389a457d/src/Illuminate/Auth/SessionGuard.php#L41).

willrowe:
> The guard on `AuthenticationException` not always equals `passport.guard` but depends on the context. For example this exception is thrown on `CheckScopes` middleware, but it has nothing to do with `passport.guard`. Actually [this only makes sense on `AuthorizationController`](https://github.com/laravel/passport/blob/6594120ff686782af9af82d30498b7be117d937a/src/PassportServiceProvider.php#L131-L133):
> 
> ```diff
> // src/Http/Controllers/AuthorizationController.php
> 
> protected function promptForLogin(Request $request): never
> {
>     $request->session()->put('promptedForLogin', true);
> 
> -   throw new AuthenticationException;
> +   throw new AuthenticationException(guards: isset($this->guard->name) ? [$this->guard->name] : []);
> }
> ```
> 
> Note: `SessionGuard` is the only implementation of the `StatefulGuard` and it [has `name` property](https://github.com/laravel/framework/blob/61e221c18a89780ed5b51b59f04c0b4c389a457d/src/Illuminate/Auth/SessionGuard.php#L41).

Good points, I was trying to make the least amount of changes, but you're right, it would not be accurate. I'm going to rethink the approach.

PR #1806, opened 2025-02-27
Feature: Add a trait for easy testing applications that use the Passport package

alizeinodin:
In testing, when we use the Sanctum package, everything is easy. We can to use the Sanctum::actingAs() method for testing But in the Passport, We must config our client save it in the database, and then write our test. This trait (WithPersoanlClient) will ease our work and like the RefreshDatabase trait can be used in our test. A Personal client is created for the user and without any config user can write a test and use Passport::actingAs().

The developer must define two properties in the .env file:

```
PASSPORT_PERSONAL_ACCESS_CLIENT_ID=[client_id]

PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET=[client-secret]

```

These values will be created for the user when installing the Passport in personal client mode.

### DIFF
--- a/src/Testing/WithPersonalClient.php
+++ b/src/Testing/WithPersonalClient.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace App\Traits;
+
+use Laravel\Passport\Client;
+use Laravel\Passport\PersonalAccessClient;
+
+trait WithPersonalClient
+{
+    public function setUp(): void
+    {
+
+        $client = Client::factory()->create([
+            'id' => env("PASSPORT_PERSONAL_ACCESS_CLIENT_ID"),
+            'secret' => env("PASSPORT_PERSONAL_ACCESS_CLIENT_SECRET"),
+            'personal_access_client' => true
+        ]);
+
+        PersonalAccessClient::create([
+            'client_id' => $client->id,
+        ]);
+    }
+}


hafezdivandari:
But a client is not needed when using `Passport::actingAs()`, what am I missing here? You may refer to [docs](https://laravel.com/docs/12.x/passport#testing):

```php
use App\Models\User;
 
public function test_servers_can_be_created(): void
{
    Passport::actingAs(
        User::factory()->create(),
        ['create-servers']
    );
 
    $response = $this->post('/api/create-server');
 
    $response->assertStatus(201);
}
```

alizeinodin:
> But a client is not needed when using `Passport::actingAs()`, what am I missing here? You may refer to [docs](https://laravel.com/docs/12.x/passport#testing):
> 
> ```
> use App\Models\User;
> use Laravel\Passport\Passport;
>  
> public function test_servers_can_be_created(): void
> {
>     Passport::actingAs(
>         User::factory()->create(),
>         ['create-servers']
>     );
>  
>     $response = $this->post('/api/create-server');
>  
>     $response->assertStatus(201);
> }
> ```

When we want to test auth routes, this error returns:
Attempt to read property "secret" on null

This issue was solved by this trait.

hafezdivandari:
Would you please share your test and/or the exception's stack trace?

taylorotwell:

PR #1813, opened 2025-05-04
Use isset to check for initialisation

patrickomeara:
While upgrading to 13 I hit the below error:

```
Error: Typed static property Laravel\Passport\Passport::$keyPath must not be accessed before initialization in /Users/pat/Code/web/vendor/laravel/passport/src/Passport.php:407
```


hafezdivandari:
Duplicate of #1811

PR #1819, opened 2025-05-21
Changed getTokenFromRequest method to handle null value and prevent type error

eldair:
This code change solves `Laravel\Passport\Guards\TokenGuard::getTokenFromRequest(): Return value must be of type string, null returned` error which can happen when `X-CSRF-TOKEN` is missing from the request.

It also aligns `getTokenFromRequest` method with it's counterpart in `VerifyCsrfToken` class (which has no return value and handles null value in `tokensMatch` method)

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -257,15 +257,17 @@ class TokenGuard implements Guard
      */
     protected function validCsrf(array $token): bool
     {
-        return isset($token['csrf']) && hash_equals(
-            $token['csrf'], $this->getTokenFromRequest()
-        );
+        $requestToken = $this->getTokenFromRequest();
+
+        return isset($token['csrf']) &&
+               is_string($requestToken) &&
+               hash_equals($token['csrf'], $requestToken);
     }
 
     /**
      * Get the CSRF token from the request.
      */
-    protected function getTokenFromRequest(): string
+    protected function getTokenFromRequest(): ?string
     {
         $token = $this->request->header('X-CSRF-TOKEN');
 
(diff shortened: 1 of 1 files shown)


hafezdivandari:
I think it is safer to just change the return type to `?string` ?

eldair:
@hafezdivandari `hash_equals` function which uses the return value of this method requires all arguments to be `string`

hafezdivandari:
@eldair yes, that's why we should change the `validCsrf` method instead, check if the token is not `null` there.

eldair:
@hafezdivandari something like this? Partly copied from similar method (`tokensMatch`) in `Illuminate\Foundation\Http\Middleware\VerifyCsrfToken`

hafezdivandari:
Lgtm, thanks.

src/Guards/TokenGuard.php
```diff
@@ -256,20 +257,26 @@ protected function decodeJwtTokenCookie(): array
...
+    protected function getTokenFromRequest(): ?string
     {
         $token = $this->request->header('X-CSRF-TOKEN');
 
         if (! $token && $header = $this->request->header('X-XSRF-TOKEN')) {
-            $token = CookieValuePrefix::remove($this->encrypter->decrypt($header, static::serialized()));
+            try {
+                $token = CookieValuePrefix::remove($this->encrypter->decrypt($header, static::serialized()));
+            } catch (DecryptException) {
+                $token = '';
```
hafezdivandari:
```suggestion
                $token = null;
```

taylorotwell:
Tests fail.

hafezdivandari:
@taylorotwell failed tests are unrelated to changes on this PR and already have been fixed via laravel/framework#55800 (not released yet).

@eldair you may mark this PR as ready again, thanks.

PR #1821, opened 2025-05-27
fix: possible php error if redirect column is an empty string

Tofandel:
image


hafezdivandari:
This actually never happens, the `redirect_uris` column (if exists) is always a json array (may be emtpy `[]`) but never `null` or `''` unless you have manipulated the table manually, if so I suggest to have default value of `[]` for your previouse records instead of this PR.

Tofandel:
> This actually never happens
> unless you have manipulated the table manually

So it can happen :) especially after a migration

This is just a very basic sanity check

PR #1823, opened 2025-06-03
[13.x] Document changes to `oauth_clients` table in upgrade guide

gdebrauwer:


hafezdivandari:
As you know, these changes are backward compatible, so no upgarde guide entry is needed, as it may cause more confusion than clearifying things IMO. Also it doesn't explain the steps required to apply the changes. The upgarde guide doesn't meant to explain same things as the changelog?

gdebrauwer:
@hafezdivandari I initially did not change that table in my projects, but then my OAuth tests failed. These failures were caused by calling `Client::factory()->create()`. That factory assumes the table uses the new columns.

hafezdivandari:
@gdebrauwer You are right, so we can instruct the user to use the old `ClientFactory`, something like the following:

Passport's `Laravel\Passport\Database\Factories\ClientFactory` factory class has been changed based on the new schema, you may use [the old factory class](https://github.com/laravel/passport/blob/12.x/database/factories/ClientFactory.php)...

PR #1837, opened 2025-07-29
REFACTOR: Standardize all routes with the pattern `passport.<resource>{*}.<action>`

renoirtech:
The route `passport.tokens` was the only one not complying with the naming pattern adopted on the rest of the project's routes. Also it can mislead to the interpretation that if the name omitted the action, it might be a getter.

In order to let this organized in the same standard, I am suggesting this refactor. 

image

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -5,7 +5,7 @@ use Laravel\Passport\Passport;
 
 Route::post('/token', [
     'uses' => 'AccessTokenController@issueToken',
-    'as' => 'token',
+    'as' => 'token.issue',
     'middleware' => 'throttle',
 ]);
 


taylorotwell:
This would be a breaking change.

renoirtech:
@taylorotwell so, honest question:

should I recreate this PR pointing to the dev branch? Or not even bother since it would be closed anyway?

hafezdivandari:
@renoirtech merging this means we should add an entry to the upgrade guide for the next major version and ask the devs to find and replace every usage of this route name on their apps, and people are gonna ask why? If there is not any convincing answer to this question, then this PR will be probably closed.

renoirtech:
> @renoirtech merging this means we should add an entry to the upgrade guide for the next major version and ask the devs to find and replace every usage of this route name on their apps, and people are gonna ask why? If there is not any convincing answer to this question, then this PR will be probably closed.

That makes sense. Thank you.

PR #1844, opened 2025-08-07
 Fix viewPrefix Method to Match Documentation Example

trippo:
In the upgrade.md guide, there’s an example showing how to set the view prefix using:

```php
// Or using conventional names under the given prefix...
Passport::viewPrefix('auth.oauth');
 ```
 
However, in practice, this does not work as intended unless the prefix includes a trailing dot, like so:
```php
Passport::viewPrefix('auth.oauth.');
```
To address this inconsistency, this PR updates the viewPrefix method to automatically handle the trailing dot, making it consistent with the documentation. This avoids requiring users to remember to add the dot manually and ensures the example in the upgrade guide works out of the box.

Alternatively, we would have to update the documentation — but this fix ensures a better developer experience.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -598,9 +598,9 @@ class Passport
      */
     public static function viewPrefix(string $prefix): void
     {
-        static::authorizationView($prefix.'authorize');
-        static::deviceAuthorizationView($prefix.'device.authorize');
-        static::deviceUserCodeView($prefix.'device.user-code');
+        static::authorizationView($prefix.'.authorize');
+        static::deviceAuthorizationView($prefix.'.device.authorize');
+        static::deviceUserCodeView($prefix.'.device.user-code');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Passport.php
```diff
@@ -598,9 +598,9 @@ public static function viewNamespace(string $namespace): void
      */
     public static function viewPrefix(string $prefix): void
     {
-        static::authorizationView($prefix.'authorize');
-        static::deviceAuthorizationView($prefix.'device.authorize');
-        static::deviceUserCodeView($prefix.'device.user-code');
+        static::authorizationView($prefix.'.authorize');
```
hafezdivandari, line 603:
```suggestion
        $prefix = rtrim($prefix, '.');

        static::authorizationView($prefix.'.authorize');
```

hafezdivandari (commented):
Avoid a breaking change by trimming dot from end of the given prefix.

PR #1847, opened 2025-08-27
[13.x] Added option to set device code token expire date

jayan-blutui:
This PR adds an option to set a custom expiry date for device code tokens. This implementation is similar to that of `tokensExpireIn` method. The registration of the device code grant has also been updated to add the new option or fallback to the `tokensExpireIn` if not set.

The reason for creating this PR is so that I can set a custom expiry date for the device code grant that I plan to use for my first party CLI tool. Currently my `tokensExpireIn` in is set to 8 hours which would lead to a bad developer experience for anyone wanting to use my CLI tool on a daily basis.

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -75,6 +75,11 @@ class Passport
     public static ?DateInterval $personalAccessTokensExpireIn = null;
 
     /**
+     * The date when device code tokens expire.
+     */
+    public static ?DateInterval $deviceCodeTokensExpireIn = null;
+
+    /**
      * The name for API token cookies.
      */
     public static string $cookie = 'laravel_token';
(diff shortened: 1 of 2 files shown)


taylorotwell:
@hafezdivandari could you review this if you have a chance?

hafezdivandari:
Sorry, but I don’t think we should support custom token expiration time for the **Device Code** grant.

It only makes sense to support custom expiration times in these cases:

- **Personal Access Token** (already supported via `Passport::$personalAccessTokensExpireIn`)  
- **Client Credentials Grant** (not currently supported, but could be)  
- **Refresh Token Grant** (already supported via `Passport::$refreshTokensExpireIn`)  
- **All authorization-based grants** (Auth Code, Device Code, Password, and Implicit) should share the same expiration time, configured via `Passport::$tokensExpireIn`  

In your use case, I suggest relying on a long-lived refresh token to renew your short-lived access token instead.

jayan-blutui:
No worries. Totally understand the reasoning. Appreciate the review @hafezdivandari

PR #1850, opened 2025-09-10
Update Client Model

samsin33:
On generating personal access token **createToken()** throws error which is
"Personal access client not found for '$provider' user provider. Please create one."

It occurs due to **confidential()** in Client model returns false as the secret attribute is hidden, but getAttributes() does the job.

**Note: I tested passport along with mongodb package.**

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -209,7 +209,7 @@ class Client extends Model
      */
     public function confidential(): bool
     {
-        return ! empty($this->secret);
+        return ! empty($this->getAttributes()['secret'] ?? null);
     }
 
     /**


hafezdivandari:
This has been merged, but I see no relation between your issue (the thrown excpetion) and the changes on this PR.

> It occurs due to confidential() in Client model returns false as the secret attribute is hidden.

This statement is not true?

We have passing tests for issuing PATs here: https://github.com/laravel/passport/blob/13.x/tests/Feature/PersonalAccessGrantTest.php

samsin33:
@hafezdivandari My issue got fixed when I override the confidential() method in my custom Client model, hence the PR. May be the issue only occurs with mongo db.
**Before override:**

before


**After override:**

after

hafezdivandari:
I'm trying to find the source of the issue here, it may be MongoDB bug with accessing hidden attributes, or you have other overriden methods/customized client model  that is causing this issue. For example I see `personal_access` attribute in your client which seems wierd.

samsin33:
I overridden only confidendial() and nothing else, rest the model was compatible with mongodb as mentioned in the documentation here - https://www.mongodb.com/docs/drivers/php/laravel-mongodb/current/user-authentication/#std-label-laravel-user-auth-passport.

Once I run this artisan command - php artisan passport:client --personal
the below record is created.
image

The personal_access_client attribute is automatically added.

**You can ignore the "personal_access" attribute that I added manually for testing purpose.**

samsin33:
This issue only occurs with mongodb may be because of the casts, accessors and mutators, which may be not in sync with the mongodb base model. MySQL is working fine.

lolclark:
Would this not be the same needed for the validateClient function in src/Bridge/ClientRepository as it uses : 

`$this->hasher->check($clientSecret, $record->secret)`

But as hidden in mongodb model, $record->secret is empty. I tried this with mongodb and always get invalid client, so i extended the ClientRepository like this : 

`return $record && ! empty($clientSecret) && $this->hasher->check($clientSecret, $record->getAttributes()['secret']);`

And it works fine now ! Sorry this might not be the correct place to post this, but it seems to be the same problem.

Thanks

hafezdivandari:
@samsin33 @Torquin I’d suggest reporting this issue on the laravel-mongodb package instead.

cc @GromNaN could you please weigh in? It seems that hidden attributes on an Eloquent model are not accessible when using MongoDB.

GromNaN:
I'm tracking this issue for investigation: [PHPORM-415](https://jira.mongodb.org/browse/PHPORM-415)

PR #1831, opened 2025-06-16
added migration file check before publish passport-migration

tareqsifat:
# Laravel Passport: Prevent Duplicate Migration Files

## ✨ Summary

This pull request improves the behavior of the `passport:install` command by adding a check to avoid re-publishing Passport migration files if they already exist in the `database/migrations` directory.

---

## 🥉 Problem

When running `php artisan passport:install` on a fresh Laravel setup using Laragon's Cmder terminal, the following error occurred:

```
LogicException
Choice question must have at least 1 choice available.
```

After troubleshooting, Attempting to rerun the command after partial execution led to **duplicate migration files** and migration errors.

---

## ✅ Solution

This PR introduces a pre-check before publishing Passport migration files. It ensures the command:

```php
$this->call('vendor:publish', ['--tag' => 'passport-migrations']);
```

...is only executed **if** the Passport migration files don't already exist.

### Migration Check Logic

```php
$migrationFilesExist = collect(glob(database_path('migrations/*_create_oauth_*.php')))->isNotEmpty();

if (! $migrationFilesExist) {
    $this->call('vendor:publish', ['--tag' => 'passport-migrations']);
} else {
    $this->components->info('Passport migration files already exist. Skipping publishing...');
}
```
## 🧹 Code Clean-up

- Removed unused imports (`DB` and `Schema`) for cleaner code and to pass static analysis.
---

## 🎯 Benefits

- ✅ Prevents accidental duplication of migration files
- ✅ Ensures safer and more predictable CLI execution
- ✅ Avoids migration class name conflicts
- ✅ Improves developer experience (DX) by being idempotent
- ✅ Aligns with Laravel's philosophy of intuitive tooling

---

## 🌟 Notes

This enhancement is fully backward compatible. Existing users of `passport:install` will now benefit from an extra safeguard against duplicate migrations without needing to change their workflow.

---

Thanks for reviewing this PR! Looking forward to your feedback and suggestions.

Issue #1781, opened 2024-08-19
Repeated migrations created with 'php artisan passport:install ' command without existence check


Ozckar07:
### Passport Version

12.3

### Laravel Version

11.9

### PHP Version

8.2.0

### Description

When executing the php artisan passport:install command in a Laravel project, it generates new migration files without checking if they already exist. As a result, each time the command is run, duplicate migration files are created, leading to multiple copies of the same migrations. This causes errors during the migration process because the migration files already exist and cannot be executed again, leading to database conflicts.

### Steps To Reproduce

1. Set up a fresh Laravel project with Laravel Passport.
2. Run php artisan passport:install once to generate the initial migrations.
3. Run php artisan passport:install again and notice that new migration files are created, duplicating the previous ones.
4. Repeat the command, and each time, new copies of the same migrations are generated.

**Expected behavior:**
The php artisan passport:install command should check if the migrations already exist before generating new ones, preventing the creation of duplicate migration files.

**Actual behavior:**
Each time the command is executed, duplicate migration files are created, causing an error during the migration execution due to the presence of existing migration files.

driesvints:
It's expected to only run this command once.

Ozckar07:
"In previous versions of Laravel, it was possible to run the php artisan passport:install command multiple times without issues, as the command would simply proceed if there were no migrations pending, allowing the generation of new CLIENT_ID values. However, in Laravel 11, running this command more than once leads to the replication of the Passport migrations, as you can see in the output. This is because each time the command is executed, it attempts to publish the migrations again and prompts for running them. If the encryption keys already exist, the command throws an error unless the --force option is used. Therefore, this command should now only be run once, unless there's a specific reason to overwrite the existing keys or re-migrate, to avoid unnecessary duplication or errors."


`PS C:\Users\app> php artisan passport:install

   ERROR  Encryption keys already exist. Use the --force option to overwrite them.

   INFO  Publishing [passport-migrations] assets.

  Copying directory [C:\Users\app\database\migrations]  DONE

 Would you like to run all pending database migrations? (yes/no) [yes]:
 > y

   INFO  Nothing to migrate.

 Would you like to create the "personal access" and "password grant" clients? (yes/no) [yes]:
 > y

   INFO  Personal access client created successfully.

  Client ID ........................................................ 3
  Client secret ............. nvAmoD4GraSdSaSI1M5cBK49PiR0NgVEGLI86Uo5

   INFO  Password grant client created successfully.

[...]

Issue #1729, opened 2024-03-14
Passport install prompt


Tahiaji:
### Passport Version

12.0.0

### Laravel Version

11.0.5

### PHP Version

8.2

### Database Driver & Version

8.0

### Description

After updating the package, the command `passport:install` now has two prompts for user input.

![image](https://github.com/laravel/passport/assets/2605542/247c548a-3389-4f82-bcf3-3a2816d6c762)

This command is used in deployments and tests and there is no possibility of a manual response. This change is not mentioned in the upgrade guide.

How can I now execute this command without user interaction?

### Steps To Reproduce

```shell
php artisan passport:install
```

Tahiaji:
Example case of usage in tests https://stackoverflow.com/a/50243105/1202097

nunomaduro:
The `passport:install` command was not meant to be run multiple times. Meaning, it should not be part of a deployment process or run before tests.

Typically, that command should run only once per project locally (because it publishes migrations, etc). This change, of adding prompts to this install command is not mentioned in the upgrade guide because that command only serves 'fresh' new Laravel applications.

Tahiaji:
Even if such behavior was not intended - there are many places where it can be used without user interaction.

And even if this is expected behavior, the change in the operating logic is extremely significant and it would be nice to at least mention it in the change history

MarcEspiard:
Hey, just thought I'd mention that this also broke my tests during the upgrade from 10 to 11.
It would be nice to at least mention it in the upgrade guide or the changelog.

Fixed it by publishing the passport migrations and replacing the passport:install call in my test seeder by: 
```php
Artisan::call('passport:keys');
Artisan::call('passport:client', ['--personal' => true, '--name' => config('app.name') . ' Personal Access Client']);
$provider = in_array('users', array_keys(config('auth.providers'))) ? 'users' : null;
Artisan::call('passport:client', ['--password' => true, '--name' => config('app.name') . ' Password Grant Client', '--provider' => $provider]);
```

RaylsonCavalcante:
Show! Isso resolveu o problema.
Testei e funciona.

RaylsonCavalcante:
> Show! Isso resolveu o problema.
Testei e funciona.
> 
> ```
> Artisan::call('passport:keys');
> Artisan::call('passport:client', ['--personal' => true, '--name' => config('app.name') . ' Personal Access Client']);
> $provider = in_array('users', array_keys(config('auth.providers'))) ? 'users' : null;
[...]

PR #1855, opened 2025-09-24
[13.x] Prevent duplicate Passport migrations by skipping existing files in passport:install

devrabiul:
**Description:**
The `passport:install` command currently republishes all Passport migration stubs on every run. Each new copy gets a fresh timestamp, which leads to duplicate migration files and database errors when running `migrate`.

This is a **serious problem** for projects running Passport in production or multiple environments.

This commit updates `InstallCommand` to:

• Check for existing Passport migration files in the app’s `database/migrations` directory.
• Skip publishing a stub if its base name already exists.
• Publish only missing migrations with a new timestamp.
• Log clear messages for published and skipped migrations.

**Example output:**

```
Published Passport migration: 2025_09_24_120000_create_oauth_auth_codes_table.php
Skipped existing Passport migration: _create_oauth_access_tokens_table.php
```

This makes the command idempotent and safe to run multiple times, while retaining the default behaviour for first-time installs.

**Tested on:** Laravel 12.x

---

**Note:**
@laravel – this issue can cause duplicate tables and migration errors in real projects. Fixing this will prevent developers from encountering database conflicts when running `passport:install` multiple times.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -35,7 +35,36 @@ class InstallCommand extends Command
         ]);
 
         $this->call('vendor:publish', ['--tag' => 'passport-config']);
-        $this->call('vendor:publish', ['--tag' => 'passport-migrations']);
+                
+        // Selectively publish Passport migrations (plain PHP)
+        $migrationPath = database_path('migrations');
+        $passportMigrationPath = base_path('vendor/laravel/passport/database/migrations');
+
+        // Build a list of all migration stubs in the package
+        $stubFiles = glob($passportMigrationPath . '/*.php');
+
+        $this->components->info('Checking Passport migrations…');
+
+        foreach ($stubFiles as $stubPath) {
+            // Strip the timestamp from the stub filename
+            $stubFilename = basename($stubPath);
+            $baseName = preg_replace('/^\d+_\d+_\d+_\d+_/', '_', $stubFilename);
+
+            // Check if a migration with this base name already exists
+            $existing = glob($migrationPath . '/*' . $baseName);
+
+            if (empty($existing)) {
+                // No migration yet → copy it with a new timestamp
+                $newName = date('Y_m_d_His') . $baseName;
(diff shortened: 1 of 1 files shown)


hafezdivandari:
Duplicate of https://github.com/laravel/passport/issues/1729#issuecomment-1997832885, https://github.com/laravel/passport/issues/1781#issuecomment-2296822524, #1831

> The `passport:install` command was not meant to be run multiple times. Meaning, it should not be part of a deployment process or run before tests.

You probably just need `passport:keys`.

devrabiul:
Hi @hafezdivandari,

I understand that `passport:install` wasn’t originally intended to be run multiple times, and I respect that. However, in real-world workflows—CI/CD pipelines, local environment resets, or multi-environment setups—it *sometimes is run more than once*, which currently causes **duplicate migrations and errors**.

This PR doesn’t change the core behavior of the command. It simply **makes it safe to re-run** by skipping already published migrations. First-time installs still work exactly the same, so nothing is broken.

Regarding `passport:keys`, that only regenerates encryption keys and doesn’t address the duplicate migration issue, which is the core problem here.

As a compromise, the PR could also log a warning if `passport:install` is re-run, keeping users aware that it’s not meant for multiple executions, while still preventing errors.

The goal is **developer safety and idempotence** without changing intended behavior. This small change could prevent a lot of frustration and accidental production errors.

Thanks for considering!

hafezdivandari:
Runing `passport:install` is equivalent to executing the following commands:

1. `passport:keys`
2. `vendor:publish --tag=passport-config`
3. `vensor:publish --tag=passport-migrations`
4. `migrate`
5. `passport:client --personal`

As mentioned above, this command shouldn't be called multiple times. You can't make it safe for re-run, as this may change in the future and other prompts/commands be added.

Just pick the commands you need, most of the time only `passport:keys` is enough. However, you may try to make `vendor:publish` command safe for re-run instead maybe?

taylorotwell:

Issue #24, opened 2016-08-19
Separating resource & authorisation server roles


AndyBursh:
Hi Laravel Passport contributors,

I'm glad to see Laravel now has an official, integrated oauth 2 implementation. However, I notice that right now there is no way to separate the resource server and authorisation roles (as defined in the oauth 2 spec). That is to say, Passport cannot be setup purely to be a resource server which accepts and validates JWTs against a known authorisation sever's public key/certificate*.

Is this something that is envisioned in any roadmap of the package? (I don't see one documented anywhere, so perhaps those plans are in somebody's head?) I would be interested in contributing to the splitting of those roles, because I see this as a shortcoming of many other like packages.

Regards,
Andy

*Unless I'm very much mistaken. If somebody could point me in the right direction, that would be appreciated

themsaid:
I'm interested to know why is it useful to make this type of separation? Just Curious :)

lukepolo:
Its useful when you have a large resources with a lot of endpoints that need securing so you can have a standalone authorize server. This would allow us to setup multiple micro services, and still use the same authorization server.

Resource Server A :
Resource Server B:

Service A - Photo Service
Service B - Video Service
Service C - Processing Service

Scenario 1 - Security : 

Service A / B / C , ask for a token authorized by Resource A. And therefore do not require us to send our password to 3 different services just one. Also if resource A is compromised , we can then reject all tokens from that resource server and use Resource Server B. 

Scenario 2 - Temporary Access to resources :

Service A wants access to Service C for a user. We now can just temporarily give Service A access to the users resources without the users name and password.

AndyBursh:
@lukepolo , could you clarify what you mean by "Auth Resource Server", please? Do you mean a server acting as both the authorisation and resource server roles? I think I understand your scenarios, but better safe than sorry :)

A third scenario could be when communicating across business boundaries. Perhaps the business has bought in some fancy new SSO kit that implements OAuth 2 (and maybe OpenId connect), and now they want to be able to log in to their purchase order system with it. In this case, the PO system isn't creating the token, but simply accepting and verifying it.

In some cases, it's even feasible that the service receiving the tokens doesn't need or want any more information than what's in the token. Continuing with the PO system example, here are a couple of scenarios where information in the token would be sufficient

* **Access control**. The scopes contained in the token can be used to drive access control. For example, if a token only contains the `department_a` scope, then the user can't view department B's data, or any of the pages specific to department B

[...]

PR #1858, opened 2025-09-29
[13.x] Support Token Revocation RFC7009 and Token Introspection RFC7662

hafezdivandari:
* thephpleague/oauth2-server#1473

## Token Revocation

[Token Revocation RFC7009](https://datatracker.ietf.org/doc/html/rfc7009)

First, enable the token revocation by calling the `Passport::enableTokenRevocation` method in the boot method of your application's `App\Providers\AppServiceProvider` class:

```php
Passport::enableTokenRevocation();
```

Then, the consuming application should make a `POST` request to your application's `/oauth/revoke` route:

```php
Http::asForm()->post('https://laravel.test/oauth/revoke', [
    'client_id' => 'your-client-id',
    'client_secret' => 'your-client-secret', // Required for confidential clients only...
    'token' => 'your-token-here',            // Either an access token or a refresh token to revoke...
    'token_type_hint' => 'access_token',     // Optional, to help the server optimize the token lookup. 'access_token' or 'refresh_token'...
]);
```

This will return an empty `200` response.

## Token Introspection

[Token Introspection RFC7662](https://datatracker.ietf.org/doc/html/rfc7662)

First, enable the token introspection by calling the `Passport::enableTokenIntrospection` method in the boot method of your application's `App\Providers\AppServiceProvider` class:

```php
Passport::enableTokenIntrospection();
```

Then, the consuming application should make a `POST` request to your application's `/oauth/introspect` route:

```php
Http::asForm()->post('https://laravel.test/oauth/introspect', [
    'client_id' => 'your-client-id',
    'client_secret' => 'your-client-secret', // Required for confidential clients only...
    'token' => 'your-token-here',            // Either an access token or a refresh token to introspect...
    'token_type_hint' => 'access_token',     // Optional, to help the server optimize the token lookup. 'access_token' or 'refresh_token'...
]);
```

This will return a JSON response containing `active` attribute that indicates whether or not the presented token is currently active.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,13 +27,19 @@
         "illuminate/encryption": "^11.35|^12.0",
         "illuminate/http": "^11.35|^12.0",
         "illuminate/support": "^11.35|^12.0",
-        "league/oauth2-server": "^9.2",
+        "league/oauth2-server": "dev-master-token-revocation-introspection",
         "php-http/discovery": "^1.20",
         "phpseclib/phpseclib": "^3.0",
         "psr/http-factory-implementation": "*",
         "symfony/console": "^7.1",
         "symfony/psr-http-message-bridge": "^7.1"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/hafezdivandari/oauth2-server"
+        }
+    ],
     "require-dev": {
         "mockery/mockery": "^1.6",
         "orchestra/testbench": "^9.9|^10.0",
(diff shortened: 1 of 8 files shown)


ulcuber:
Great work. I think we will also need two **middlewares** settings for each of new routes like `passport.introspection.middlewares` and `passport.revokation.middlewares` so we can check for specific **Client** as **Token Introspection** is primarily needed only by trusted **Resource Server** and **Token Revokation** could be even more sensitive API where I'd rely solely on the **Authorization Server**. And that said, we will need **Token Introspection Endpoint** Consumer (**Resource Server**) as discussed in:

https://github.com/laravel/passport/issues/24

PR #1859, opened 2025-10-08
Fix »OAuth Client Table Changes« migration snippet in the upgrade guide

hettiger:
The migration code snippet of the v13 upgrade guide immediately crashed on me.
I had to apply several fixes.

Please see commit descriptions for more details.

Using `\App\Models\User::class` even though old apps might still use `\App\User::class` does not seem to be that big of an issue. The IDE should immediately highlight the missing class. (eefdb76730de8bc8ed5b031ecfb2435f28a81a72)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -151,7 +151,7 @@ foreach (Passport::client()->cursor() as $client) {
     Model::withoutTimestamps(fn () => $client->forceFill([
         'owner_id' => $client->user_id,
         'owner_type' => $client->user_id
-            ? config('auth.providers.'.$client->provider.'.model', \App\Models\User::class)
+            ? config('auth.providers.'.($client->provider ?: config('auth.guards.api.provider')).'.model')
             : null,
         'redirect_uris' => $client->redirect_uris,
         'grant_types' => $client->grant_types,
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -159,8 +159,6 @@ foreach (Passport::client()->cursor() as $client) {
 }
 
 Schema::table('oauth_clients', function (Blueprint $table) {
-    $table->text('grant_types')->nullable(false)->change();
-
     $table->dropColumn(['user_id', 'redirect', 'personal_access_client', 'password_client']);
 });
 ```
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -143,7 +143,7 @@ Schema::table('oauth_clients', function (Blueprint $table) {
 
     $table->after('provider', function (Blueprint $table) {
         $table->text('redirect_uris');
-        $table->text('grant_types');
+        $table->text('grant_types')->nullable();
     });
 });
 
@@ -157,6 +157,8 @@ foreach (Passport::client()->cursor() as $client) {
 }
 
 Schema::table('oauth_clients', function (Blueprint $table) {
+    $table->text('grant_types')->nullable(false)->change();
+
     $table->dropColumn(['user_id', 'redirect', 'personal_access_client', 'password_client']);
 });
 ```
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -139,9 +139,7 @@ If you prefer to use the new structure, you may create a migration to apply the 
 
 ```php
 Schema::table('oauth_clients', function (Blueprint $table) {
-    $table->after('user_id', function (Blueprint $table) {
-        $table->nullableMorphs('owner');
-    });
+    $table->nullableMorphs('owner', after: 'user_id');
 
     $table->after('provider', function (Blueprint $table) {
         $table->text('redirect_uris');
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -150,7 +150,9 @@ Schema::table('oauth_clients', function (Blueprint $table) {
 foreach (Passport::client()->cursor() as $client) {
     Model::withoutTimestamps(fn () => $client->forceFill([
         'owner_id' => $client->user_id,
-        'owner_type' => $client->user_id ? config('auth.providers.'.$client->provider.'.model') : null,
+        'owner_type' => $client->user_id
+            ? config('auth.providers.'.$client->provider.'.model', \App\Models\User::class)
+            : null,
         'redirect_uris' => $client->redirect_uris,
         'grant_types' => $client->grant_types,
     ])->save());


UPGRADE.md
```diff
@@ -139,26 +139,28 @@ If you prefer to use the new structure, you may create a migration to apply the
 
 ```php
 Schema::table('oauth_clients', function (Blueprint $table) {
-    $table->after('user_id', function (Blueprint $table) {
-        $table->nullableMorphs('owner');
-    });
+    $table->nullableMorphs('owner', after: 'user_id');
```
hafezdivandari, line 142: Passport supports Laravel 11+ but the suggested syntax is available since 12+ AFAIK. Anything wrong with current syntax?

UPGRADE.md
```diff
@@ -139,26 +139,28 @@ If you prefer to use the new structure, you may create a migration to apply the
...
 
     $table->after('provider', function (Blueprint $table) {
         $table->text('redirect_uris');
-        $table->text('grant_types');
+        $table->text('grant_types')->nullable();
```
hafezdivandari, line 146: Declare `default` instead, you may apply the same change to `redirect_uris` column too:

```suggestion
        $table->text('grant_types')->default('[]');
```

UPGRADE.md
```diff
@@ -139,26 +139,28 @@ If you prefer to use the new structure, you may create a migration to apply the
...
     });
 });
 
 foreach (Passport::client()->cursor() as $client) {
     Model::withoutTimestamps(fn () => $client->forceFill([
         'owner_id' => $client->user_id,
-        'owner_type' => $client->user_id ? config('auth.providers.'.$client->provider.'.model') : null,
+        'owner_type' => $client->user_id
+            ? config('auth.providers.'.$client->provider.'.model', \App\Models\User::class)
```
hafezdivandari:
```suggestion
            ? config('auth.providers.'.($client->provider ?: config('auth.guards.api.provider')).'.model')
```

UPGRADE.md
```diff
@@ -139,26 +139,28 @@ If you prefer to use the new structure, you may create a migration to apply the
...
+            : null,
         'redirect_uris' => $client->redirect_uris,
         'grant_types' => $client->grant_types,
     ])->save());
 }
 
 Schema::table('oauth_clients', function (Blueprint $table) {
+    $table->text('grant_types')->nullable(false)->change();
```
hafezdivandari: No need anymore.

UPGRADE.md
hettiger: The `$table->after('user_id', function (Blueprint $table) { … });` wrap did not work for me in combination with `$table->nullableMorphs('owner');`.

> Passport supports Laravel 11+ but the suggested syntax is available since 12+ AFAIK.

I didn't check that. If it needs to work on L11 too and the `owner_type` + `owner_id` columns should end up right after the `user_id` column, we'd probably have to manually create the individual columns within the `$table->after('user_id', function (Blueprint $table) { … });` wrap.

UPGRADE.md
hettiger: > Declare default instead

```php
// \Laravel\Passport\Client::grantTypes

Attribute::make(
    get: fn (?string $value): array => isset($value) ? $this->fromJson($value) : array_keys(array_filter([
        'authorization_code' => ! empty($this->redirect_uris),
        'client_credentials' => $this->confidential() && $this->firstParty(),
        'implicit' => ! empty($this->redirect_uris),
        'password' => $this->password_client,
        'personal_access' => $this->personal_access_client && $this->confidential(),
        'refresh_token' => true,
        'urn:ietf:params:oauth:grant-type:device_code' => true,
    ])),
);
```

If you default to `'[]'`, `isset($value)` will evaluate to `true` for all rows during the migration. Therefore `$this->fromJson($value)` is being called instead of the `array_keys(array_filter([ … ]))` code, that actually preserves the data (`$this->password_client` and `$this->personal_access_client`) during the migration.

**⇒ Your suggested change would drop data instead of preserving it.**

> you may apply the same change to redirect_uris column too:

`redirect_uris` does not use an isset check so it works just fine. No need to change anything about `redirect_uris`.

UPGRADE.md
hettiger: See: https://github.com/laravel/passport/pull/1859#discussion_r2414883223

hettiger:
Thank you for the super quick review @hafezdivandari

UPGRADE.md
hafezdivandari: You're right, so I guess `nullable` is fine for both columns?

UPGRADE.md
hafezdivandari: > The $table->after('user_id', function (Blueprint $table) { … }); wrap did not work for me in combination with $table->nullableMorphs('owner');

Weired, I'll run a test on the framework to make sure!

UPGRADE.md
hafezdivandari: Still redundant? it's fine to keep it nullable for migrated apps imo.

UPGRADE.md
hettiger:
```php
// \Laravel\Passport\Client::redirectUris

Attribute::make(
    get: fn (?string $value, array $attributes): array => match (true) {
        ! empty($value) => $this->fromJson($value),
        ! empty($attributes['redirect']) => explode(',', $attributes['redirect']),
        default => [],
    },
);
```

I tested it with just one redirect URL as well as multiple URLs joined by commas. It always worked just fine. The `redirect` column was not nullable, so we can expect a value. I don't think we need to change anything about it. However, adding the nullable and removing it later, just as we do with `grant_types` does not hurt either.

UPGRADE.md
hettiger: It probably wont make a difference because the Attribute internally falls back to the `array_keys(array_filter([ ... ]))` logic. I'll remove it.

UPGRADE.md
hafezdivandari: We should make them both `nullable` because adding a not null column without default causes error on some DB drivers!

UPGRADE.md
hafezdivandari: It was working fine, but a bug /breaking change has been introduced by laravel/framework#56613

Please keep it as is, I'll send a PR to the framework to fix this.

hettiger:
@taylorotwell @hafezdivandari 

I think this was not 100% ready. I didn't have the time to answer sooner and now coming back here I see this has already been merged. It's not a big deal — no need to revert. However, maybe @hafezdivandari wants to apply his suggested changes in a follow-up PR. I consider this as done on my end. 

Thank you! :-)

PR #1860, opened 2025-10-10
Fix ClientController::store() breaking change introduced via #1745

hettiger:
Fixes `ClientController::store()` breaking change introduced via #1745

```php
// \Laravel\Passport\Http\Controllers\ClientController::store

- if (Passport::$hashesClientSecrets) {
-     return ['plainSecret' => $client->plainSecret] + $client->toArray();
- }
+ $client->secret = $client->plainSecret;

return $client->makeVisible('secret');
```

This change obviously breaks usages that previously relied on the return type array with the additional `plainSecret` data. E.g., the old Vue components used `plainSecret` to present that to the user so that he could save it, etc. Since hashing is now mandatory, I restored the previous behavior without the now obsolete `Passport::$hashesClientSecrets` check:

```php
return ['plainSecret' => $client->plainSecret] + $client->toArray();
```

I also updated the tests. I know it looks a bit fishy but I had not much choice since it's a unit test … (didn't want to make too big of a change out of this … it's deprecated anyways …)

### DIFF
--- a/src/Http/Controllers/ClientController.php
+++ b/src/Http/Controllers/ClientController.php
@@ -38,7 +38,7 @@ class ClientController
     /**
      * Store a new client.
      */
-    public function store(Request $request): Client
+    public function store(Request $request): array
     {
         $this->validation->make($request->all(), [
             'name' => ['required', 'string', 'max:255'],
@@ -53,9 +53,7 @@ class ClientController
             $request->user(),
         );
 
-        $client->secret = $client->plainSecret;
-
-        return $client->makeVisible('secret');
+        return ['plainSecret' => $client->plainSecret] + $client->toArray();
     }
 
     /**
--- a/tests/Unit/ClientControllerTest.php
+++ b/tests/Unit/ClientControllerTest.php
@@ -51,7 +51,7 @@ class ClientControllerTest extends TestCase
         $clients->shouldReceive('createAuthorizationCodeGrantClient')
             ->once()
             ->with('client name', ['http://localhost'], true, $user)
-            ->andReturn($client = new Client);
+            ->andReturn($client = new Client(['name' => 'client']));
 
         $redirectRule = m::mock(RedirectRule::class);
 
@@ -70,7 +70,10 @@ class ClientControllerTest extends TestCase
             $clients, $validator, $redirectRule
         );
 
-        $this->assertEquals($client, $controller->store($request));
+        $this->assertEquals([
+            'name' => $client->name,
+            'plainSecret' => $client->plainSecret,
+        ], $controller->store($request));
     }
 
     public function test_public_clients_can_be_stored()
@@ -89,7 +92,7 @@ class ClientControllerTest extends TestCase
         $clients->shouldReceive('createAuthorizationCodeGrantClient')
             ->once()
             ->with('client name', ['http://localhost'], false, $user)
-            ->andReturn($client = new Client);
+            ->andReturn($client = new Client(['name' => 'client']));
 
         $redirectRule = m::mock(RedirectRule::class);
 
@@ -109,7 +112,10 @@ class ClientControllerTest extends TestCase
             $clients, $validator, $redirectRule
         );
 
-        $this->assertEquals($client, $controller->store($request));
+        $this->assertEquals([
+            'name' => $client->name,
+            'plainSecret' => $client->plainSecret,
+        ], $controller->store($request));
     }
 
     public function test_clients_can_be_updated()


src/Http/Controllers/ClientController.php
```diff
@@ -38,7 +38,7 @@ public function forUser(Request $request): Collection
     /**
      * Store a new client.
      */
-    public function store(Request $request): Client
+    public function store(Request $request): array
```
hafezdivandari, line 41: BC change!

src/Http/Controllers/ClientController.php
```diff
@@ -53,9 +53,7 @@ public function store(Request $request): Client
             $request->user(),
         );
 
-        $client->secret = $client->plainSecret;
-
-        return $client->makeVisible('secret');
+        return ['plainSecret' => $client->plainSecret] + $client->toArray();
```
hafezdivandari, line 56:
```suggestion
        $client->secret = $client->plainSecret;

        return $client->makeVisible('secret')->mergeAppends(['plain_secret']);
```

src/Http/Controllers/ClientController.php
hettiger: @hafezdivandari it has been array|Client before: https://github.com/laravel/passport/blob/89d2594c0a7185bafeb82ed7e1de1a2e152162dd/src/Http/Controllers/ClientController.php#L76-L78

You updated it to just Client via https://github.com/laravel/passport/commit/ef122adb47edee48c6d83493102b7dd43a281541#diff-6656ea7db3c71aee64d3cec3512c61d176630ed9f86a2b7359565fd3bfd57aa3 ignoring that people may already rely on the array etc.

In my book thats a breaking change introduced by you?

hafezdivandari: That PR was on a major release 12 to 13 where breaking changes are expected to happen, but this PR is on a patch release!

src/Http/Controllers/ClientController.php
hettiger: I thought about this as well. However, it's again breaking existing code bases like mine with no good reason. This is deprecated code. Why change the behavior now just because you want to have the Client return type over array?

By the way: I think this will throw `BadMethodCallException  Call to undefined method Laravel\Passport\Client::getPlainSecretAttribute()`; `\Laravel\Passport\Client::$plainSecret` is a property. It's not a getter.

src/Http/Controllers/ClientController.php
hettiger: Yes of course, but was this breaking change intended? Looked like a mistake to me and upgrade.md does not mention such a BC, right?

hettiger: Want me to add a note in the https://github.com/laravel/passport/blob/13.x/UPGRADE.md#json-api-deprecation section + update the `Client` model to move towards your suggested solution in https://github.com/laravel/passport/pull/1860#discussion_r2421552365 instead? It would still break old upgraded apps but at least the `plain_secret` is available so it would be an easy fix.

hafezdivandari: TBH I see no good reason to change anything here, because `secret` will now contain the plain secret value that’s only visible right after creating a new client. The change on `array|Client` to stricter return type `Client` was intended, and still makes sense (We have already explained the return type changes on the upgrade guide). However if you insist to include `plain_secret` too, my suggestion is to append it. That's just my suggestion and the final decision is for maintainers to make.

hettiger: Thanks for the feedback @hafezdivandari 

> because secret will now contain the plain secret value that’s only visible right after creating a new client. 

Unfortunately that's not true. My first attempt was to update the old Vue components to use `secret` directly, expecting it to return the plain secret in that case. However, it didn't. 

```php
// \Laravel\Passport\Client::secret

    protected function secret(): Attribute
    {
        return Attribute::make(
            set: function (?string $value): ?string {
                $this->plainSecret = $value;

                return $this->castAttributeAsHashedString('secret', $value);
            },
        );
    }
```

As you can see, it is getting hashed immediately in the setter. I don't think, we can leave it as is tbh.

hafezdivandari: Good catch, let me write a test for it and think for a solution, in the meantime, please mark this PR as draft.

hettiger: Marked as draft. Thank you for jumping in and taking care. 🚀

hafezdivandari: @hettiger I just sent #1861, please let me know if it resolves the issue, thanks!

hettiger: Thank you @hafezdivandari it's working as before now. Only had to update `Client.vue` as expected:

```vue
                        if (response.data.plain_secret) {
                            this.showClientSecret(response.data.plain_secret);
                        }
```

> It was `plainSecret` before …

PR #1862, opened 2025-10-20
ActingAsClient should be that same as how client credientals actual work

JerryVerhoef:
League Oauth server is confirming to the RFC standard there should always be a sub field in JWT token. And sub field is mapped to client identifier as fallback if the user is not present. This happens with all client_crediential grants.

related to:
https://github.com/thephpleague/oauth2-server/issues/1456


taylorotwell:
Is this fixing an actual problem? If so, can you describe it?

PR #1866, opened 2025-10-24
feat: Allow `findForPassport` to optionally receive the OAuth client

maximepvrt:
Currently, Laravel Passport uses the `findForPassport(string $username)` method to retrieve the user during password grant authentication. This method only receives the username, which limits cases where we might want to filter users based on the OAuth client.

This PR introduces the possibility to define `findForPassport` with a **second optional parameter**, a `ClientEntityInterface $client`, allowing:

* Filtering users based on the OAuth client if needed.
* Maintaining compatibility with existing methods that only accept one parameter.

Example usage:

```php
use Laravel\Passport\Bridge\Client;

public function findForPassport(string $username, Client $client)
{
    return $this->where('email', $username)
        ->whereHas('clients', function ($q) use ($client) {
            $q->where('id', $client->getIdentifier());
        })->first();
}
```

This is **not a breaking change**: existing methods with a single parameter will continue to work.

📖 I also prepared a PR for the documentation to explain this new possibility: https://github.com/laravel/docs/pull/10880

### DIFF
--- a/src/Bridge/UserRepository.php
+++ b/src/Bridge/UserRepository.php
@@ -6,8 +6,8 @@ use Illuminate\Contracts\Hashing\Hasher;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 use League\OAuth2\Server\Entities\UserEntityInterface;
 use League\OAuth2\Server\Repositories\UserRepositoryInterface;
-use RuntimeException;
 use ReflectionMethod;
+use RuntimeException;
 
 class UserRepository implements UserRepositoryInterface
 {
--- a/src/Bridge/UserRepository.php
+++ b/src/Bridge/UserRepository.php
@@ -44,7 +44,7 @@ class UserRepository implements UserRepositoryInterface
             $method = new ReflectionMethod($instance, 'findForPassport');
             $args = [$username];
             if ($method->getNumberOfParameters() >= 2) {
-                $args[] = $client;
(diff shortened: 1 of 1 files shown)


src/Bridge/UserRepository.php
```diff
@@ -39,9 +40,17 @@ public function getUserEntityByUserCredentials(
...
+        if (method_exists($instance, 'findForPassport')) {
+            $method = new ReflectionMethod($instance, 'findForPassport');
+            $args = [$username];
+            if ($method->getNumberOfParameters() >= 2) {
+                $args[] = $clientEntity;
+            }
+            $user = $method->invokeArgs($instance, $args);
+        } else {
+            $user = $instance->where('email', $username)->first();
+        }
```
hafezdivandari:
```suggestion
        $user = method_exists($model, 'findForPassport')
            ? (new $model)->findForPassport($username, $clientEntity)
            : (new $model)->where('email', $username)->first();
```

hafezdivandari (commented):
No need to use reflection, just pass the `$clientEntity` as argument. If you want to make this change, please apply the same to line 38:

```diff
- $user = (new $model)->findAndValidateForPassport($username, $password);
+ $user = (new $model)->findAndValidateForPassport($username, $password, $clientEntity);
```

maximepvrt:
@hafezdivandari updated ;-)

PR #1871, opened 2025-11-21
[13.x] Fix possible 500 error when retrieving user on token guard

### LINKED ISSUE #1868: TokenGuard throws 500 when Client Credentials token ID type (UUID) mismatches user model ID type (int)
### Passport Version

13.0

### Laravel Version

12.37.0

### PHP Version

8.4.1

### Database Driver & Version

pgsql 18

### Description

When a Client Credentials token is sent to a route protected by the user guard (auth:api), Laravel Passport attempts to retrieve a user using the token’s oauth_user_id. 
Because:

User model primary key: integer ($incrementing = true)

Client Credentials token oauth_user_id: UUID string

the database query fails (WHERE id = 'uuid-string'), resulting in a 500 Internal Server Error.

Expected behavior: Passport should detect that the token belongs to a client, not a user, and gracefully return 401 Unauthorized, instead of throwing a DB exception.

### Steps To Reproduce

1. Issue a Client Credentials token.
2. Define a route protected by the user guard:
```
Route::middleware('auth:api')->get('/user-endpoint', function () {
    return response()->json(['success' => true]);
});
```

3. send a request to the route with the Client Credentials token in the Authorization: Bearer header.

4. Observe: 500 Internal Server Error occurs because the guard attempts to query the user with a UUID string ID.

Expected Behavior

  Passport should detect that the token is not associated with a user.
  Return 401 Unauthorized.
  Avoid executing a query with mismatched ID types (int vs UUID).

Actual Behavior
  
  Database exception occurs due to type mismatch: WHERE id = 'uuid-string'.
  Server responds with 500 Internal Server Error instead of 401.

hafezdivandari:
Fixes #1868

This PR fixes an issue where using a valid client-credentials token on an `auth:api` guarded endpoint accidentally may triggers a database type mismatch, causing a 500 error instead of the expected 401.

This PR also fixes a few PHPDoc types.

### DIFF
--- a/src/Console/ClientCommand.php
+++ b/src/Console/ClientCommand.php
@@ -131,7 +131,7 @@ class ClientCommand extends Command
     {
         $confidential = $this->hasOption('public')
             ? ! $this->option('public')
-            : $this->confirm('Would you like to make this client confidential?', true);
+            : $this->components->confirm('Would you like to make this client confidential?', true);
 
         return $clients->createDeviceAuthorizationGrantClient($this->option('name'), $confidential);
     }
@@ -151,7 +151,7 @@ class ClientCommand extends Command
             : $this->components->confirm('Would you like to make this client confidential?', true);
 
         $enableDeviceFlow = Passport::$deviceCodeGrantEnabled &&
-            $this->confirm('Would you like to enable the device authorization flow for this client?');
+            $this->components->confirm('Would you like to enable the device authorization flow for this client?');
 
         return $clients->createAuthorizationCodeGrantClient(
             $this->option('name'), explode(',', $redirect), $confidential, null, $enableDeviceFlow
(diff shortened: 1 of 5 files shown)


Bloemendaal:
@hafezdivandari Hi! May I ask why the docblock for the `$next` param was changed from Symfony Response to Illuminate? In the Laravel documentation on middleware they use Symfony: https://laravel.com/docs/13.x/middleware#defining-middleware

Theoretically, I can change my type to the Illuminate variant to stop phpstan from complaining, but that'd be a bit risky I guess for middleware return types? 🤔

hafezdivandari:
@Bloemendaal Hi, it is because `withCookie` has been defined on `Illuminate\Http\Response` only and doesn't exist on `Symfony\Component\HttpFoundation\Response` class, and it means that the `$response` has been resolved as `Illuminate\Http\Response` at this point.

https://github.com/laravel/passport/blob/87608729a291dfcae6027887de5effec16ebf979/src/Http/Middleware/CreateFreshApiToken.php#L50

Bloemendaal:
Aaah got it! Thanks for your fast response! 🙌

Issue #1870, opened 2025-11-21
redirectUris and grantTypes accessors receive null $value, causing empty array return


valkyriweb:
### Passport Version

13.4.0

### Laravel Version

12.x

### PHP Version

8.4

### Database Driver & Version

PostgreSQL 16

### Description

The `redirectUris` and `grantTypes` Attribute accessors in `Laravel\Passport\Client` receive `$value` as `null` from Laravel's Attribute system, even when the database contains valid JSON data. This causes the accessors to return empty arrays.

This breaks OAuth 2.1 dynamic client registration when using `laravel/mcp`, as newly created clients return `redirect_uris: []` in the response despite the data being correctly stored in the database.

### Steps To Reproduce

```php
$clients = app(\Laravel\Passport\ClientRepository::class);

$client = $clients->createAuthorizationCodeGrantClient(
    name: 'Test Client',
    redirectUris: ['https://example.com/callback'],
    confidential: false,
);

// Database has correct data:
dump($client->getAttributes()['redirect_uris']); 
// Output: '["https:\/\/example.com\/callback"]' (correct JSON string)

// But accessor returns empty:
dump($client->redirectUris);
// Output: [] (empty array - BUG)
```

### Root Cause

The accessor in `Client.php`:

```php
protected function redirectUris(): Attribute
{
    return Attribute::make(
        get: fn (?string $value, array $attributes): array => match (true) {
            ! empty($value) => $this->fromJson($value),
            // ...
        },
    );
}
```

Laravel passes `$value` as `null`, but the data exists in `$attributes['redirect_uris']`. The accessor should read from `$attributes` directly.

### Suggested Fix

```php
protected function redirectUris(): Attribute
{
    return Attribute::make(
        get: function (mixed $value, array $attributes): array {
            $redirectUris = $attributes['redirect_uris'] ?? $value;

            if (is_array($redirectUris)) {
                return $redirectUris;
            }

            if (! empty($redirectUris) && is_string($redirectUris)) {
                return $this->fromJson($redirectUris);
            }

            if (! empty($attributes['redirect'])) {
                return explode(',', $attributes['redirect']);
            }

            return [];
        },
    );
}
```

The same fix should be applied to `grantTypes()`.

### Workaround

Create a custom Client model that extends `Laravel\Passport\Client`, override the accessors, and register it with `Passport::useClientModel()`.

hafezdivandari:
Model attributes are snake case. Try:

```php
dump($client->redirect_uris);
[...]

PR #1872, opened 2025-11-25
Fix unhandled possible redirect_uris field usage in Client Model

mathiasmoser:
This Bugfix adds consideration of the newly added redirect_uris field in Oauth Clients Table. This new field was introduced with Passport 13. If you decided to switch to new oauth_clients table fields you will stumble over the described problem.

### Behaviour without fix:

If you are using the Oauth Register Route in Laravel MCP to register a new Client you send the request:

**The Register Request:**

POST oauth/register

Body:
{"client_name":"Test","redirect_uris":["https://test1.com", "https://test2.com"]}

**Response:**
{
    "client_id": "51",
    "grant_types": [
        "authorization_code",
        "implicit",
        "refresh_token",
        "urn:ietf:params:oauth:grant-type:device_code"
    ],
    "response_types": [
        "code"
    ],
    "redirect_uris": [],
    "scope": "mcp:use",
    "token_endpoint_auth_method": "none"
}

**Problem:** redirect_uris are empty because of the missing consideration of a potential redirect_uris field (instead of the old redirect field).

### Behaviour with fix:

**The Register Request:**

POST oauth/register

Body:
{"client_name":"Test","redirect_uris":["https://test1.com", "https://test2.com"]}

**Response:**
{
    "client_id": "51",
    "grant_types": [
        "authorization_code",
        "implicit",
        "refresh_token",
        "urn:ietf:params:oauth:grant-type:device_code"
    ],
    "response_types": [
        "code"
    ],
    "redirect_uris": [
        "https://test1.com",
        "https://test2.com"
    ],
    "scope": "mcp:use",
    "token_endpoint_auth_method": "none"
}

The Redirect URIs are now correctly returned!

### Side Effects of the Bug

This Bug currently does lead to non working MCP Oauth Workflow with Anthropic Claude. OpenAI instead seems to have no problems with the missing redirect_uris in the response.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -153,6 +153,7 @@ class Client extends Model
             get: fn (?string $value, array $attributes): array => match (true) {
                 ! empty($value) => $this->fromJson($value),
                 ! empty($attributes['redirect']) => explode(',', $attributes['redirect']),
+                ! empty($attributes['redirect_uris']) => $this->fromJson($attributes['redirect_uris']),
                 default => [],
             },
         );


hafezdivandari:
This is a bug in the mcp package not Passport:

* https://github.com/laravel/mcp/issues/112
* https://github.com/laravel/passport/issues/1870#issuecomment-3563098117

mathiasmoser:
@hafezdivandari thanks for your input.

I still think that it is a bug in Passport.

Why?

The MCP Package is a Consumer of the Package and uses the offered Accessor Method "redirectUris". This method relies on a field name "redirect" which was the old way (before Passport 13) to store the redirectUris in the Client Table. 

With Passport 13 the Structure of Client Table was refactored and the new field name is redirect_uris. But the switch to the new Client Table structure is not mandatory - so both possible field names (redirect and redirect_uris) should be supported.

This is exactly what this PR will provide.

After that it's safe to access the available Accessor Method "redirectUris" on the Client Model and for example also the MCP Package Problem is solved.

hafezdivandari:
@mathiasmoser I understand what you mean, but this accessor returns `redirect_uris` attribute if present and fallbacks to `redirects` as expected. As I said in the issue's comment above Model attributes are **snake case**. Please close this one and send a PR to MCP package if you want:

https://github.com/laravel/mcp/blob/01d44b89b20480409990a5051d522686756ddf2d/src/Server/Http/Controllers/OAuthRegisterController.php#L47-L54

```diff
return response()->json([
    'client_id' => (string) $client->id,
-   'grant_types' => $client->grantTypes,
+   'grant_types' => $client->grant_types,
    'response_types' => ['code'],
-   'redirect_uris' => $client->redirectUris,
+   'redirect_uris' => $client->redirect_uris,
    'scope' => 'mcp:use',
    'token_endpoint_auth_method' => 'none',
]);
```

cc @pushpak1300

mathiasmoser:
So this means $client->redirectUris should never be called because it would behave wrongly in combination with redirect_uris field. in my eyes my fix would have solved this. but fine, i will close the pr.

PR #1880, opened 2026-01-28
Add dedicated token lifetime for client credentials grant

sajanp:
## Summary

This PR adds a `Passport::clientCredentialsTokensExpireIn()` method for configuring token lifetimes specifically for the client credentials grant, mirroring the existing `personalAccessTokensExpireIn()` pattern.

This is a non-breaking change. When not explicitly configured, client credentials tokens continue to use `Passport::tokensExpireIn()` as before.

## Motivation

The client credentials grant is designed for autonomous machine-to-machine interactions where tokens are regenerated automatically. Currently there is no way to configure token lifetime specifically for this grant type without affecting all other grants.

Many developers using Laravel Passport may not be familiar with the security implications of long-lived tokens or the nuances between OAuth 2.0 grant types. Providing a dedicated configuration method helps guide them toward better security practices.

### Industry Guidance

- **[RFC 9700 - OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/rfc9700/)** (January 2025): "Access tokens should be restricted to the minimum required... helps reduce the impact of access token leakage."
- **[OWASP OAuth2 Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/OAuth2_Cheat_Sheet.html)**: "The validity of an access token should be between 5 and 15 minutes, depending on the sensitivity of the resources."
- **[Google API Security Guidelines](https://cloud.google.com/apigee/docs/api-platform/antipatterns/oauth-long-expiration)**: Explicitly calls out long token expiration as an antipattern, recommending 30 minutes as a starting point.

## Usage

```php
use DateInterval;

public function boot(): void
{
    Passport::clientCredentialsTokensExpireIn(new DateInterval('PT1H'));
}
```

## Changes

- Added `Passport::$clientCredentialsTokensExpireIn` property and `Passport::clientCredentialsTokensExpireIn()` method
- Service provider uses `clientCredentialsTokensExpireIn() ?? tokensExpireIn()` for backwards compatibility
- Added test for custom client credentials token expiration

### DIFF
--- a/src/Passport.php
+++ b/src/Passport.php
@@ -75,6 +75,11 @@ class Passport
     public static ?DateInterval $personalAccessTokensExpireIn = null;
 
     /**
+     * The interval when client credentials tokens expire.
+     */
+    public static ?DateInterval $clientCredentialsTokensExpireIn = null;
+
+    /**
      * The name for API token cookies.
      */
     public static string $cookie = 'laravel_token';
@@ -320,6 +325,20 @@ class Passport
         }
 
         return static::$personalAccessTokensExpireIn = $date instanceof DateTimeInterface
+            ? Date::now()->diff($date)
+            : $date;
+    }
+
+    /**
+     * Get or set when client credentials grant tokens expire.
+     */
+    public static function clientCredentialsTokensExpireIn(DateTimeInterface|DateInterval|null $date = null): ?DateInterval
+    {
+        if (is_null($date)) {
+            return static::$clientCredentialsTokensExpireIn;
+        }
+
+        return static::$clientCredentialsTokensExpireIn = $date instanceof DateTimeInterface
             ? Date::now()->diff($date)
             : $date;
     }
--- a/src/PassportServiceProvider.php
+++ b/src/PassportServiceProvider.php
@@ -158,7 +158,7 @@ class PassportServiceProvider extends ServiceProvider
                 }
 
                 $server->enableGrantType(
-                    new ClientCredentialsGrant, Passport::tokensExpireIn()
+                    new ClientCredentialsGrant, Passport::clientCredentialsTokensExpireIn() ?? Passport::tokensExpireIn()
                 );
 
                 if (Passport::$implicitGrantEnabled) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## [Unreleased](https://github.com/laravel/passport/compare/v13.4.4...13.x)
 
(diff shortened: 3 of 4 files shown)


hafezdivandari:
Related to https://github.com/laravel/passport/pull/1847#issuecomment-3237826904

Being able to set a custom expiration time for client credential tokens is useful, but my suggestion is to set the default to 1 year as is, and send it against 13.x without a breaking change. Then you can send another PR/issue to request for 1 hour default in the next major version.

src/Passport.php
```diff
@@ -324,6 +329,20 @@ public static function personalAccessTokensExpireIn(DateTimeInterface|DateInterv
             : $date;
     }
 
+    /**
+     * Get or set when client credentials grant tokens expire.
+     */
+    public static function clientCredentialsTokensExpireIn(DateTimeInterface|DateInterval|null $date = null): DateInterval
+    {
+        if (is_null($date)) {
+            return static::$clientCredentialsTokensExpireIn ??= new DateInterval('PT1H');
```
hafezdivandari:
```suggestion
            return static::$clientCredentialsTokensExpireIn ??= new DateInterval('P1Y');
```

tests/Feature/PassportTestCase.php
```diff
@@ -19,6 +19,8 @@ abstract class PassportTestCase extends TestCase
 
     protected function setUp(): void
     {
+        Passport::$clientCredentialsTokensExpireIn = null;
```
hafezdivandari: Move this to the `setup` method of `‎tests/Feature/ClientCredentialsGrantTest.php‎` test itself, it is easier to follow there.

sajanp:
> Related to [#1847 (comment)](https://github.com/laravel/passport/pull/1847#issuecomment-3237826904)
> 
> Being able to set a custom expiration time for client credential tokens is useful, but my suggestion is to set the default to 1 year as is, and send it against 13.x without a breaking change. Then you can send another PR/issue to request for 1 hour default in the next major version.

I think that is fair.  I will do just that later today.  Thanks.

src/PassportServiceProvider.php
```diff
@@ -158,7 +158,7 @@ function (AuthorizationServer $server): void {
                 }
 
                 $server->enableGrantType(
-                    new ClientCredentialsGrant, Passport::tokensExpireIn()
+                    new ClientCredentialsGrant, Passport::clientCredentialsTokensExpireIn()
```
hafezdivandari:
```suggestion
                    new ClientCredentialsGrant, Passport::clientCredentialsTokensExpireIn() ?? Passport::tokensExpireIn()
```

sajanp:
Took me a while to come back to this - apologies.  The PR has been fixed/updated.

PR #1887, opened 2026-02-18
Add missing `OAuthenticatable` interface step to upgrade guide

FeBe95:
The upgrade guide was missing a step instructing users to add the `Laravel\Passport\Contracts\OAuthenticatable` interface to their `App\Models\User` model. This interface was introduced in Passport 13. Prior versions only required the `HasApiTokens` trait. Without this, upgrading users will have no indication they need to add it.

## Changes

Added a new entry to `UPGRADE.md` under the "Upgrading To 13.0 From 12.x" section documenting the required `User` model change, consistent with what the installation docs already state.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -70,6 +70,23 @@ PR: https://github.com/laravel/passport/pull/1755
 
 When authenticating users via bearer tokens, the `User` model's `token` method now returns an instance of `Laravel\Passport\AccessToken` class instead of `Laravel\Passport\Token`.
 
+### User Model Interface
+
+PR: https://github.com/laravel/passport/pull/1797
+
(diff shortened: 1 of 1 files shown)


hafezdivandari:
This is optional, that's why it is not mentioned on the upgrade guide but in the docs for new installations. Upgrade guide usually contains breaking changes only.

FeBe95:
@taylorotwell Thanks for merging!

@hafezdivandari We were already using the `HasApiTokens` trait before. The trait now includes the `@phpstan-require-implements` annotation to make sure that the trait is only used with `OAuthenticatable` classes. PHPStan started reporting errors because of this change, but I couldn’t find any mention of it in the upgrade guide. That’s why I thought it would be helpful to include it there.

PR #1892, opened 2026-02-22
Allow custom middleware to be registered

timacdonald:
Like other packages, such as Fortify, Nova, and Horizon, this adds the ability to add custom middleware, via config, to the routes passport registers.

The motivation behind this feature is to allow us to intercept any passport routes and enforce certain rules. The main one being that the `client_id` must be a UUID.

Right now, when someone hits our application with `client_id=123`, a query exception is thrown as the database is expecting a UUID. For most databases, it probably just 404s, but for certain databases, they error out when the wrong type is used for the query.

This would allow us to create a similar pattern we use for our routes, where we 404 before we even hit the database, e.g., 

```php
Route::get('environments/{environment}')
    ->whereUuid('environment');
```

The middleware will allow us to handle query parameters (the current problem) as well as route parameters.


config/passport.php
```diff
@@ -15,6 +15,8 @@
 
     'guard' => 'web',
 
+    'middleware' => [],
+
```
timacdonald, line 19: I'm sure this could use some Taylor comment love.

PR #1895, opened 2026-03-18
Fix session serialization for auth request and device code

### LINKED ISSUE #1894: Session Serialization default in L13 causes exception
### Passport Version

^13.0

### Laravel Version

^13.0

### PHP Version

8.4

### Database Driver & Version

MySQL 9.4

### Description

I'm using Passport to authenticate an MCP server. The new default session serialization in Laravel 13 is JSON, which causes errors during authorization. 

```php
Laravel\Passport\Http\Controllers\ApproveAuthorizationController::getAuthRequestFromSession(): Return value must be of type League\OAuth2\Server\RequestTypes\AuthorizationRequestInterface, array returned
```

### Steps To Reproduce

Setup an app using Laravel MCP (Server, Tools), Laravel Passport.
Add routes to `ai.php` - `Mcp::oauthRoutes()`
Add `auth:api` middleware to MCP server route
Attempt to authenticate a user, using OAuth

JoshSalway:
## Summary
- Laravel 13 defaults to JSON session serialization, which cannot properly serialize/deserialize League OAuth2 objects (`AuthorizationRequestInterface`, `DeviceCodeEntityInterface`)
- When retrieved from a JSON-serialized session, these objects are returned as arrays instead of their original class instances, causing `ApproveAuthorizationController::getAuthRequestFromSession()` to fail with "Return value must be of type AuthorizationRequestInterface, array returned"
- This fix explicitly `serialize()`s these objects to strings before storing them in the session, and `unserialize()`s them when retrieved
- The `is_string()` check provides backward compatibility with existing sessions that may still contain objects from the native PHP session serializer

Fixes #1894

## Test plan
- [x] Updated unit tests to use real `AuthorizationRequest` objects to verify serialization works correctly
- [x] All 101 unit tests pass
- [x] Backward compatible: `is_string()` check handles both serialized strings and legacy object values

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -82,7 +82,7 @@ class AuthorizationController
         }
 
         $request->session()->put('authToken', $authToken = Str::random());
-        $request->session()->put('authRequest', $authRequest);
+        $request->session()->put('authRequest', serialize($authRequest));
 
         return $viewResponse->withParameters([
             'client' => $client,
--- a/src/Http/Controllers/DeviceAuthorizationController.php
+++ b/src/Http/Controllers/DeviceAuthorizationController.php
@@ -55,7 +55,7 @@ class DeviceAuthorizationController
         $client = $this->clients->find($deviceCode->getClient()->getIdentifier());
 
         $request->session()->put('authToken', $authToken = Str::random());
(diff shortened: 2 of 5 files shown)


taylorotwell:
Should probably use allowed_classes.

JoshSalway:
Moving to fork.

PR #1898, opened 2026-03-26
[13.x] Parse prompt parameter as a space-delimited list per OIDC spec

jhm-ciberman:
The OIDC Core 1.0 spec ([Section 3.1.2.1](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)) defines `prompt` as:

> OPTIONAL. Space-delimited, case-sensitive list of ASCII string values that specifies whether the Authorization Server prompts the End-User for reauthentication and consent.

Currently, Passport checks the `prompt` parameter using strict string equality, so a request like `prompt=login consent` silently matches none of the checks.

This PR parses `prompt` into an array via `explode()` and replaces the four equality checks with `in_array()`. Single-value requests like `prompt=login` continue to work identically.

### DIFF
--- a/src/Http/Controllers/AuthorizationController.php
+++ b/src/Http/Controllers/AuthorizationController.php
@@ -49,13 +49,16 @@ class AuthorizationController
             ($psrRequest->getQueryParams()['response_type'] ?? null) === 'token'
         );
 
+        $prompt = $request->input('prompt');
+        $promptValues = $prompt ? explode(' ', $prompt) : [];
+
         if ($this->guard->guest()) {
-            $request->input('prompt') === 'none'
+            in_array('none', $promptValues)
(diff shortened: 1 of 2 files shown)


jhm-ciberman:
The static analysis fail was introduced in https://github.com/laravel/passport/pull/1897. It's unrelated to this change.

jhm-ciberman:
@taylorotwell Sorry, any reason why this PR was rejected? In our company we are implementing OpenID connect via passport and having this will meant that we are fully compliant. 

Just to be clear, this is a bugfix. Not a new feature. And it's fully retro compatible. 

Edit: I added an Issue explaining the bug in detail: https://github.com/laravel/passport/issues/1899

PR #1901, opened 2026-04-01
Prevent user impersonation via client credentials token

### LINKED ISSUE #1900: TokenGuard::authenticateViaBearerToken() does not validate oauth_user_id type, causing user impersonation with client_credentials tokens
### Passport Version

13.x (latest)

### Laravel Version

11.x

### PHP Version

8.2+

### Database Driver & Version

MySQL 8.x

### Description

TokenGuard::authenticateViaBearerToken() passes oauth_user_id directly to provider->retrieveById() without validating that it's actually a user identifier:
// src/Guards/TokenGuard.php
protected function authenticateViaBearerToken(): ?Authenticatable
{
    if (! $psr = $this->getPsrRequestViaBearerToken()) {
        return null;
    }

    // ...

    $user = $this->provider->retrieveById(
        $psr->getAttribute('oauth_user_id') ?: null  // ← No validation
    );

    return $user?->withAccessToken(AccessToken::fromPsrRequest($psr));
}

The Problem:

For client_credentials tokens, the underlying league/oauth2-server library sets the JWT sub claim to the client's UUID (because getSubjectIdentifier() falls back to getClient()->getIdentifier() when there's no user). The BearerTokenValidator then puts this UUID into the oauth_user_id PSR attribute.

When retrieveById() is called with a UUID string on a users table with integer primary keys, MySQL performs implicit type casting:

UUID: '019c9d23-9763-7303-9bdb-3a0a6bf0xf90'
MySQL cast: → integer 19
Result: User::find(19) → returns a real user!

This means any client_credentials (guest) token silently authenticates as a real user, exposing their data.

### Steps To Reproduce

Configure Passport with client_credentials grant
Ensure the OAuth client uses UUID identifiers (Passport v13 default)
Ensure the users table uses integer auto-increment primary keys
Have a user whose ID matches the leading digits of the client UUID
Request a client_credentials token: POST /oauth/token with grant_type=client_credentials
Use the token with Auth::guard('api')->user() or $request->user()
The guard returns the unrelated user instead of null
Expected Behavior:

Auth::guard('api')->user() should return null for client_credentials tokens, since no user is associated with the token.

Actual Behavior:

Auth::guard('api')->user() returns a real user (e.g., user ID 19) due to MySQL implicit type casting of the client UUID.

pushpak1300:
For `client_credentials` tokens, League's `AccessTokenTrait::getSubjectIdentifier()` falls back to the client UUID when no user exists. The `BearerTokenValidator` maps this into `oauth_user_id`, causing `TokenGuard` to call `retrieveById()` with the client UUID. On MySQL with integer primary keys, implicit type casting can match an unrelated user row.

Fixes #1900

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -141,13 +141,17 @@ class TokenGuard implements Guard
 
         $this->setClient($client);
 
+        $oauthUserId = $psr->getAttribute('oauth_user_id');
+
+        if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
+            return null;
+        }
+
         // If the access token is valid we will retrieve the user according to the user ID
         // associated with the token. We will use the provider implementation which may
         // be used to retrieve users from Eloquent. Next, we'll be ready to continue.
         try {
-            $user = $this->provider->retrieveById(
-                $psr->getAttribute('oauth_user_id') ?: null
-            );
+            $user = $this->provider->retrieveById($oauthUserId);
         } catch (Exception) {
             return null;
         }
--- a/tests/Unit/TokenGuardTest.php
+++ b/tests/Unit/TokenGuardTest.php
@@ -47,17 +47,17 @@ class TokenGuardTest extends TestCase
 
         $resourceServer->shouldReceive('validateAuthenticatedRequest')->andReturn($psr = m::mock(ServerRequestInterface::class));
         $psr->shouldReceive('getAttribute')->with('oauth_user_id')->andReturn(1);
-        $psr->shouldReceive('getAttribute')->with('oauth_client_id')->andReturn(1);
+        $psr->shouldReceive('getAttribute')->with('oauth_client_id')->andReturn(2);
         $psr->shouldReceive('getAttribute')->with('oauth_access_token_id')->andReturn('token');
         $psr->shouldReceive('getAttributes')->andReturn([
             'oauth_user_id' => 1,
-            'oauth_client_id' => 1,
+            'oauth_client_id' => 2,
             'oauth_access_token_id' => 'token',
             'oauth_scopes' => [],
         ]);
         $userProvider->shouldReceive('retrieveById')->with(1)->andReturn(new TokenGuardTestUser);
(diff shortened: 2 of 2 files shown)


src/Guards/TokenGuard.php
```diff
@@ -141,13 +141,17 @@ protected function authenticateViaBearerToken(): ?Authenticatable
 
         $this->setClient($client);
 
+        $oauthUserId = $psr->getAttribute('oauth_user_id');
+
+        if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
```
jonnott, line 146: @pushpak1300 If the project is using `Passport::$clientUuids = false`, and the user Id and client Id both happened to be e.g. 1 (perhaps in a test), would this not return null rather than operate as expected?

hafezdivandari: Fixed via #1902

michaelw85:
I'm debugging an issue where a single user keeps getting logged out in my app and I tracked it down to this change.
```php
        // TokenGuard.php:147
        if (empty($oauthUserId) || ($oauthUserId === $psr->getAttribute('oauth_client_id') && $client->hasGrantType('client_credentials'))) {
            return null;
        }
```
So in my case, the user has ID 2, and the oauth client also has ID 2 with the grant type "client_credentials". I'm really confused about this code. Does this mean I'm not allowed to have a user ID that matches an oauth client ID?

hafezdivandari:
@michaelw85 please refer to #1913

michaelw85:
@hafezdivandari Thank you for pointing me in the right direction!

michaelw85:
@hafezdivandari FYI, I checked my DB, and I did the migrations in the past. What I noticed at a first glance is that my owner_type and id columns are empty. Going to dig a bit deeper.

Edit:
For anyone stumbling upon this issue I have resolved my issue by removing `"client_credentials"` from the `oauth_clients.grant_types` column for my client. My authorization [flow](https://www.laravel.wiki/en/passport#converting-authorization-codes-to-access-tokens) was using the `grant_type: 'authorization_code'`

PR #1902, opened 2026-04-01
[13.X] fix: resolve user as null when user ID matches client ID on integer key setups

### LINKED ISSUE #1903: Valid bearer tokens rejected when user ID matches client ID (integer PKs)
PR #1901 (v13.7.1) added this check to `TokenGuard::authenticateViaBearerToken()`:

```php
if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
    return null;
}
```

This works fine with UUID client IDs since a user ID would never collide with a client UUID. But with integer client IDs, it's very common for user ID 1 to authenticate through client ID 1 (password grant). The guard rejects the token thinking it's a `client_credentials` token.

**To reproduce:** Create a password grant token for user `id=1` via client `id=1`, then try to authenticate. Returns `null` instead of the user.

The `client_credentials` grant type info is available on the token itself. Checking the grant type instead of comparing IDs would avoid false positives.

- PHP 8.4 / Laravel 13.3 / Passport 13.7.1

Can-Kar:
## Problem

When using integer (non-UUID) primary keys, if a user's ID happens to equal the ID of the password grant client, `TokenGuard::authenticateViaBearerToken()` incorrectly returns `null` and the user cannot be authenticated.

The faulty check in `TokenGuard.php`:

```php
if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
    return null;
}
```

This condition was intended to detect `client_credentials `grant tokens (which have no associated user). However, when integer IDs are used, a legitimate password grant token for user ID `2` will be rejected if the client ID is also `2`, because the strict equality check `$oauthUserId === $psr->getAttribute('oauth_client_id')` evaluates to `true`.

## Root Cause

The `UPGRADE.md` for v13 documents UUID migration as optional. Applications that continue using integer IDs are vulnerable to this ID collision.

The `=== oauth_client_id check` alone is too broad — it rejects any token where the user ID matches the client ID, regardless of grant type. The fix scopes this check to only apply when the client actually uses the `client_credentials` grant type.

Reproduction Steps

1.     Use integer (non-UUID) primary keys for both `users` and `oauth_clients` tables.
2.     Seed a password grant client — ensure its auto-incremented ID ends up as `N`.
3.     Create a user with ID `N`. 
4.     Obtain an `access token` for that user via the `password grant`.
5.     Make an authenticated request — the guard returns `null` and the request is rejected with a 401 Unauthorized response.

### DIFF
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -143,7 +143,7 @@ class TokenGuard implements Guard
 
         $oauthUserId = $psr->getAttribute('oauth_user_id');
 
-        if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
+        if (empty($oauthUserId)) {
             return null;
         }
 
--- a/src/Guards/TokenGuard.php
+++ b/src/Guards/TokenGuard.php
@@ -143,7 +143,7 @@ class TokenGuard implements Guard
 
         $oauthUserId = $psr->getAttribute('oauth_user_id');
 
-        if (empty($oauthUserId)) {
+        if (empty($oauthUserId) || ($oauthUserId === $psr->getAttribute('oauth_client_id') && $client->hasGrantType('client_credentials'))) {
             return null;
         }
 
--- a/tests/Unit/TokenGuardTest.php
+++ b/tests/Unit/TokenGuardTest.php
@@ -176,9 +176,44 @@ class TokenGuardTest extends TestCase
         $psr->shouldReceive('getAttribute')->with('oauth_user_id')->andReturn($clientId);
         $psr->shouldReceive('getAttribute')->with('oauth_client_id')->andReturn($clientId);
         $userProvider->shouldReceive('getProviderName')->andReturn(null);
-        $userProvider->shouldReceive('retrieveById')->never();
+        $userProvider->shouldReceive('retrieveById')->with($clientId)->andReturn(null);
 
         $this->assertNull($guard->user());
+    }
+
+    public function test_user_is_resolved_when_user_id_matches_client_id()
+    {
+        $resourceServer = m::mock(ResourceServer::class);
+        $userProvider = m::mock(PassportUserProvider::class);
+        $clients = m::mock(ClientRepository::class);
+        $encrypter = m::mock(Encrypter::class);
+
+        $clients->shouldReceive('findActive')
+            ->with(1)
+            ->andReturn(new TokenGuardTestClient);
+
+        $request = Request::create('/');
+        $request->headers->set('Authorization', 'Bearer token');
+
+        $guard = new TokenGuard($resourceServer, $userProvider, $clients, $encrypter, $request);
+
+        $resourceServer->shouldReceive('validateAuthenticatedRequest')->andReturn($psr = m::mock(ServerRequestInterface::class));
+        $psr->shouldReceive('getAttribute')->with('oauth_user_id')->andReturn(1);
+        $psr->shouldReceive('getAttribute')->with('oauth_client_id')->andReturn(1);
+        $psr->shouldReceive('getAttribute')->with('oauth_access_token_id')->andReturn('token');
+        $psr->shouldReceive('getAttributes')->andReturn([
+            'oauth_user_id' => 1,
+            'oauth_client_id' => 1,
+            'oauth_access_token_id' => 'token',
+            'oauth_scopes' => [],
+        ]);
+        $userProvider->shouldReceive('getProviderName')->andReturn(null);
+        $userProvider->shouldReceive('retrieveById')->with(1)->andReturn(new TokenGuardTestUser);
+
+        $user = $guard->user();
+
+        $this->assertInstanceOf(TokenGuardTestUser::class, $user);
+        $this->assertEquals(AccessToken::fromPsrRequest($psr), $user->currentAccessToken());
     }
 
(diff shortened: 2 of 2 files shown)


siarheipashkevich:
@Can-Kar yes, I've the same issue after upgrading. The token issued but is invalid.

src/Guards/TokenGuard.php
```diff
@@ -143,7 +143,7 @@ protected function authenticateViaBearerToken(): ?Authenticatable
 
         $oauthUserId = $psr->getAttribute('oauth_user_id');
 
-        if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
+        if (empty($oauthUserId)) {
```
hafezdivandari:
```suggestion
        if (empty($oauthUserId) || ($oauthUserId === $psr->getAttribute('oauth_client_id') && $client->hasGrantType('client_credentials'))) {
```

hafezdivandari: Please also consider reverting the changes that was made to the extisting tests ('1' to '2' changes) on #1901

Cc @pushpak1300

src/Guards/TokenGuard.php
```diff
@@ -143,7 +143,7 @@ protected function authenticateViaBearerToken(): ?Authenticatable
 
         $oauthUserId = $psr->getAttribute('oauth_user_id');
 
-        if (empty($oauthUserId) || $oauthUserId === $psr->getAttribute('oauth_client_id')) {
+        if (empty($oauthUserId) || ($oauthUserId === $psr->getAttribute('oauth_client_id') && $client->hasGrantType('client_credentials'))) {
```
pushpak1300, line 146: The grant type that issued the token isn't stored anywhere, correct ? so we can't check it at the guard level.
Can we use `Passport::$clientUuids` instead ? when clients use UUIDs, the collision is impossible; when they use integers, we skip the check.

```php
 if (is_null($oauthUserId) || (Passport::$clientUuids && $oauthUserId === $psr->getAttribute('oauth_client_id'))) {
      return null;
  }
```

CC: @hafezdivandari

hafezdivandari: The issue is that user impersonation is more likely when a client UUID is not used. If you have a user with ID 2 and a client with ID 2, it becomes impossible to determine whether a token is a client credentials token or not.

The best we can do is narrow the condition (as my requested changes): if the client supports the client credentials grant, and the `oauth_client_id` and `oauth_user_id` attributes are the same, this can be considered a potential impersonation scenario, although not with complete certainty. This also introduces issues in the opposite case, particularly when attempting to authenticate a client via the `EnsureClientIsResourceOwner` middleware.

As a rule of thumb, it is not safe to use the client credentials grant with incremental primary keys (without a client UUID). User IDs and client IDs must not overlap and should remain globally unique. (We may need to add this to the docs beside `EnsureClientIsResourceOwner` middleware section?)

pushpak1300: okay fair enough. I think make sense documenting this issue.

hafezdivandari (commented):
lgtm

PS: it seems that narrowing the condition as we do on this PR is the best we can do, full explanation: https://github.com/laravel/passport/pull/1902#discussion_r3026389975 and https://github.com/thephpleague/oauth2-server/issues/1456#issuecomment-2734989996

src/Guards/TokenGuard.php
Levivb: I kinda feel like adding `$oauthUserId === $psr->getAttribute('oauth_client_id')` is a breaking change. 

Older systems using `laravel/passport` for years are most likely not migrating to uuids when upgrading `laravel/passport` -> especially since it is an optional migration and changing to uuids invalidates all existing tokens. So a big no-no for systems with a lot of active users.

And for a lot of systems which implement `laravel/passport` for the first time start out with user id 1 and client id 1, cause of the simple fact that autoincrement int primary keys in dbs behave like that. So practically all (older) systems using `laravel/passport` have a significantly high change or bumping into this bug :|

The changelog also doesn't mention this as a breaking change that needs careful planning: https://github.com/laravel/passport/blob/13.x/CHANGELOG.md#v1371---2026-04-01
When testing this locally and in the pipeline, nothing seems to be a problem until it hits production in which the database contents triggers the `$oauthUserId === $psr->getAttribute('oauth_client_id')` fallback to `null`, immediately dropping all api calls for random users, but not all users

hafezdivandari: ~Simple~ solution: https://github.com/laravel/passport/issues/1907#issuecomment-4208575885

Levivb: thanks for the quick reply. however, stating there's a 'simple solution' here ignores the facts that bigger and more complex applications are using this package and 'simple' isn't always simple. Also, if that fix would work, it would be nice to have that knowledge somewhere more visible instead of buried in a closed issue (but maybe documentation has been updated by now?)

And, this only fixes it for the collision in which it is fixable. But when automating the creation of oauth clients, a collision can't be fully known beforehand (hence: not simple)

hafezdivandari: Right, please feel free to send a PR for documenting this.

Levivb: https://github.com/laravel/passport/pull/1910 :)

PR #1911, opened 2026-04-16
Clarify that running `passport:hash` is not optional

bram-pkg:
Just a small clarification, looking at the source code of `validateClient` in `ClientRepository`:

```php
public function validateClient(string $clientIdentifier, ?string $clientSecret, ?string $grantType): bool
{
    $record = $this->clients->findActive($clientIdentifier);

    return $record && ! empty($clientSecret) && $this->hasher->check($clientSecret, $record->secret);
}
```

Hashing your secrets is no longer optional, client validation will fail if the `secret` column is not hashed. The wording in the current UPGRADE.md implies this is optional by using "MAY", which means optional as per: https://www.rfc-editor.org/rfc/rfc2119

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -85,7 +85,7 @@ As a consequence of this change, the `'passport.client_uuids'` configuration pro
 
 PR: https://github.com/laravel/passport/pull/1745
 
-Passport now always hashes client secrets using Laravel's `Hash` facade. If you are currently storing your client secrets in plain text, you may invoke the `passport:hash` Artisan command to hash all of your existing client secrets:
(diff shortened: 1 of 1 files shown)


hafezdivandari:
You don’t need to run it again if you’ve already done so and are using hashed secrets. You only need to run it if you haven’t yet. That is why it says: You **may**...

bram-pkg:
The sentence starts with "If you are currently storing your client secrets in plain text, you *may*", which isn't true though. It implies that it is optional to hash your secrets if you are currently storing them in plain text. That's not true.

If you are currently storing your secrets in plain text, you *must* hash them.

Issue #1912, opened 2026-04-23
CVE-2026-39976 fix incomplete for integer-ID setups without grant_types column


jaapredeker:
### Passport Version

13.7.1

### Laravel Version

12.x

### PHP Version

8.3+

### Description

The fix introduced in #1901 and refined in #1902 relies on `$client->hasGrantType('client_credentials')` to scope the collision check in `TokenGuard`:

```php
if (empty($oauthUserId) || ($oauthUserId === $psr->getAttribute('oauth_client_id') && $client->hasGrantType('client_credentials'))) {
    return null;
}
```

However, when the grant_types column is absent or null, Client::grantTypes() falls back to computing grant types based on client properties:

```php
'client_credentials' => $this->confidential() && $this->firstParty(),
  ```

When the grant_types column is missing or null, hasGrantType('client_credentials') returns true for any confidential first-party client (confidential() && firstParty()), which describes a typical internal API client. The `grant_types` column was introduced in Passport 13 but its migration is explicitly documented as **optional** in `UPGRADE.md`:

> "These changes are fully backward compatible, and no action is required on your part."

For applications that have `Passport::$clientUuids = false` (integer IDs) and have **not** run the optional `grant_types` schema migration, any client where the user ID happens to match the client ID will have its `refresh_token` request incorrectly rejected with a 401, because `TokenGuard` treats it as a `client_credentials` token and returns null. The security advisory and `UPGRADE.md` make no mention of this migration being required to avoid this regression.

### Steps To Reproduce

1. Install Passport 13.7.1
2. Set `Passport::$clientUuids = false` in `AppServiceProvider`
3. Skip the optional `grant_types` migration from `UPGRADE.md` — leave the column absent or null
4. Create a client — it receives integer ID e.g. `3`
5. Create a user with ID `3`
6. Obtain an access token for that user
7. Use the `refresh_token` to request a new access token via the `refresh_token` grant
8. The request is rejected with a 401 — the guard incorrectly treats the refresh token as a `client_credentials` token and returns null

hafezdivandari:
I suggest opening a PR to remove “(optional)” from the upgrade guide entry title and instead add a note within the entry explaining that it’s optional but highly recommended, especially when using integer-based client IDs.

P.S. It’s worth noting that one detail in your issue description is incorrect:

 > However, hasGrantType() has the following fallback behavior when the grant_types column is absent or null:

   ```
[...]

PR #1913, opened 2026-04-24
[13.x] Clarify grant_types migration is required for integer-based cl…

jaapredeker:
The UPGRADE.md entry for the oauth_clients table changes was marked as (Optional) with the note that "no action is required." While technically true for most setups, this is misleading for applications using integer-based client IDs (Passport::$clientUuids = false).

  The security fix introduced in #1901 and #1902 relies on hasGrantType('client_credentials') in TokenGuard. Without the grant_types migration, the fallback computes grant types based on client properties (confidential() && firstParty()), which causes any user whose integer ID matches a client ID to have their token
  requests incorrectly rejected with a 401.

  This PR removes (Optional) from the section title and adds a note recommending the migration for integer-based client ID setups. Related to #1912.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -155,12 +155,13 @@ Alternatively, you may disable this validation entirely (not recommended) by set
 Passport::$validateKeyPermissions = false;
 ```
 
-### OAuth Client Table Changes (Optional)
+### OAuth Client Table Changes
 
 PR: https://github.com/laravel/passport/pull/1744, https://github.com/laravel/passport/pull/1797
 
+Passport 13 introduces a new schema for the `oauth_clients` table. However, these changes are **fully backward compatible**, and **no action is required** on your part.
 
-Passport 13 introduces a new schema for the `oauth_clients` table. However, These changes are **fully backward compatible**, and **no action is required** on your part.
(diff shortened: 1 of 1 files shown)


nolanpro:
@hafezdivandari I'm still having trouble with the migration. After the upgrade but before the suggested migration is run, I have this personal access client:
```
mysql> select * from oauth_clients where id = 4\G
*************************** 1. row ***************************
                    id: 4
               user_id: NULL
                  name: ...
                secret: $2y$...
              provider: users
              redirect:
personal_access_client: 1
       password_client: 0
               revoked: 0
```
But `$client->grant_types`, used in the migration, returns this:
```php
\Laravel\Passport\Passport::client()->find(4)->grant_types

= [
    "client_credentials",
    "personal_access",
    "refresh_token",
    "urn:ietf:params:oauth:grant-type:device_code",
  ]
  ```
So it has client_credentials and user id 4 can't authenticate. It looks like it has client_credentials because the user_id is null and it has a secret.

As a workaround, I added to the migration:

```php
if ($client->personal_access_client === true && $client->password_client === false) {
  $grantTypes = ['personal_access'];
} else {
  $grantTypes = $client->grant_types;
}
```
But I'm not sure if that's the right way to fix it. Do you have any suggestions? Thanks!

hafezdivandari:
@nolanpro it is completely fine to update the `grant_types` attribute of your clients as you wish. You may check the following code to have better understanding of the default values for a personal access client:

https://github.com/laravel/passport/blob/6df408bc8b33866b1b18a51bc1a08c0f2b1c9c71/src/ClientRepository.php#L122-L128

Issue #1932, opened 2026-06-16
Lack of enum support on the Connection PHP attribute


rickgoemans:
### Passport Version

13.7.5

### Laravel Version

13.12.0

### PHP Version

8.4.22

### Database Driver & Version

MariaDB 10.6.24

### Description

When using custom models with the [Connection PHP attribute](https://api.laravel.com/docs/13.x/Illuminate/Database/Eloquent/Attributes/Connection.html) in combination with an UnitEnum, you'll get an error that `getConnectionName()` must return null or a string.

Since the attribute allows for UnitEnum, I guess this 1st party package should be adjusted for it as well.

Since we use a multi-db tenant setup in our project, we have to ensure the Passport models always connect to the "main" database and therefore extended the models to annotate them with the PHP attribute to enforce it.

### Steps To Reproduce

Our setup that should cover on how to reproduce:

AppServiceProvider:
```php
<?php

namespace App\Providers;

use Domain\Auth\Models\AuthCode;
use Domain\Auth\Models\Client;
use Domain\Auth\Models\RefreshToken;
use Domain\Auth\Models\Token;
use Illuminate\Support\ServiceProvider;
use Laravel\Passport\Passport;

class AppServiceProvider extends ServiceProvider
{
    public function boot(): void
    {
        $this->configurePassport();
        // ...
    }

    private function configurePassport(): void
    {
        Passport::useAuthCodeModel(AuthCode::class);
        Passport::useClientModel(Client::class);
        Passport::useRefreshTokenModel(RefreshToken::class);
        Passport::useTokenModel(Token::class);
    }

    // ...
}

```

Enum:
```php
<?php

namespace Domain\Auth\Enums;

enum ConnectionName: string
{
    case Main   = 'mysql';
    case Tenant = 'tenant';
}
```

Client model (but the same structure applies to `AuthCode`, `RefreshToken` and `Token` models):
```php
<?php

namespace Domain\Auth\Models;

use Domain\Auth\Enums\ConnectionName;
use Illuminate\Database\Eloquent\Attributes\Connection;
use Laravel\Passport\Client as PassportClient;

#[Connection(ConnectionName::Main)]
class Client extends PassportClient { 
   // ...
}
```

Error stacktrace:
```
[...]

PR #1936, opened 2026-06-24
Fix enum support for database connection name

rickgoemans:
Laravel by default has support for enum in the database connection, it does it by using the function `enum_value($this->connection)` in `getConnectionName()`, but this package it's `getConnectionName()` before my change was:
```php
/**
 * Get the current connection name for the model.
*/
public function getConnectionName(): ?string
{
    return $this->connection ?? config('passport.connection');
}
```

So in case of an enum, the return type is invalid.

This PR fixes it by using the `parent::getConnectionName()` as base and then fallback to the config file.

```php
/**
 * Get the current connection name for the model.
*/
public function getConnectionName(): ?string
{
    return parent::getConnectionName() ?? config('passport.connection');
}
```

This seems like a more scalable solution for the future if the base eloquent model ever changes.

Should be a simple fix.

### DIFF
--- a/src/AuthCode.php
+++ b/src/AuthCode.php
@@ -69,6 +69,6 @@ class AuthCode extends Model
      */
     public function getConnectionName(): ?string
     {
-        return $this->connection ?? config('passport.connection');
+        return parent::getConnectionName() ?? config('passport.connection');
     }
 }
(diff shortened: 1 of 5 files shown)


rickgoemans:
This PR addresses this issue:
https://github.com/laravel/passport/issues/1932

